PR #3, opened 2025-04-10
Remove transformers dependency

stegmuel:
## What has changed and why?

The dependency on the Transformers library is removed.

## How has it been tested?

The tests for the distillation method are updated to account for the changes

- [x ] Yes


JonasWurst:
This looks already much better. However, I would add another layer on top, since I can anticipate some refactoring incoming as soon as we want to provide more teacher models.

Hence, I'd propose the following structure:
_moduls
|_ teacher
    |_ dinov2
        - get_teacher_model
    |_ (other teacher models, aimv2, sam etc)
        - get_teacher_model
    - get_teacher_model
  
 In the distillation we would like to only interact with the teacher get_teacher_model.
  
 WDYT?

pyproject.toml
```diff
@@ -108,16 +107,30 @@ module = [
...
+]
+ignore_errors = true
+
+
+[[tool.mypy.overrides]]
+module = [
+    "lightning_fabric.utilities.rank_zero"
+]
+ignore_missing_imports = true
+disable_error_code = ["attr-defined"]
```
JonasWurst: I'm not sure I'm getting this handling fully. Is this something occurred only now?

Isn't `lightning_fabric.utilities.rank_zero` still to be part of the deps?

src/lightly_train/_methods/distillation/distillation.py
```diff
@@ -183,10 +178,7 @@ def _update_queue(self, x_teacher: Tensor) -> None:
     @torch.no_grad()
     def _forward_teacher(self, x: Tensor) -> Tensor:
         # Forward the images through the teacher model.
-        x = self.teacher_embedding_model(x)["last_hidden_state"]
-
-        # Select the [CLS] token from the (B, 1 + N, D) teacher features.
-        x = x[:, 0, :]
+        x = self.teacher_embedding_model(x)
```
JonasWurst, line 184: For the local version, we would need a new interface, is that correct? can you think of a way to prepare for that already?

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
```diff
@@ -0,0 +1,78 @@
...
+    cache_dir.mkdir(parents=True, exist_ok=True)
+    checkpoint_path = cache_dir / Path(url).name
+
+    if not checkpoint_path.exists():
+        logger.info(f"Downloading teacher weights from: {url}")
+        urlretrieve(url, checkpoint_path)
+    else:
+        logger.info(f"Using cached teacher weights from: {checkpoint_path}")
+
+    ckpt = torch.load(checkpoint_path, map_location="cpu")
```
JonasWurst: This could be refactored and moved into a separate function, considering that the get_config_path was worth a function.

Also, if other teacher models follow a similar logic, we can reuse it there

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
```diff
@@ -0,0 +1,78 @@
...
+    model.load_state_dict(ckpt, strict=True)
+    logger.info(f"Loaded teacher weights from {checkpoint_path}")
+
+    return model, embed_dim
```
JonasWurst: I would prefer If we can make the embed_dim a property of the model. Wdyt?

JonasWurst (commented):
In general it looks good. I have a few open points:
- There must be an issue with the CI, since we should have blank lines at the end of files. This is something we need to address in a different issue
- The license issue needs to be solved.
- I'm not sure how strong we should have other teacher models in our mind, since this would imply some refactoring coming.

src/lightly_train/_methods/distillation/distillation.py
stegmuel: Yes I can return the spatial tokens as well and take of discarding them in the method.
But that's a good point.

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
stegmuel: Makes sense. Other than that, we probably should add options regarding the location of the `cache_dir`.

src/lightly_train/_modules/teachers/__init__.py
```diff
@@ -0,0 +1,10 @@
+#
+# Copyright (c) Lightly AG and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+#
+from lightly_train._modules.teachers.build_teacher import get_teacher
+
+__all__ = ["get_teacher"]
```
guarin:
```suggestion
```

For internal code we usually don't add anything to the `__init__` files. You can still import the module like this:
```
from lightly_train._modules.teachers import build_teacher

...
build_teacher.get_teacher(...)
```

src/lightly_train/_modules/teachers/build_teacher.py
```diff
@@ -0,0 +1,24 @@
...
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+#
+from typing import Tuple
+
+from torch.nn import Module
+
+
+def get_teacher(teacher_name: str) -> Tuple[Module, int]:
```
guarin:
```suggestion
from __future__ import annotations

from torch.nn import Module


def get_teacher(teacher_name: str) -> tuple[Module, int]:
```

src/lightly_train/_modules/teachers/dinov2/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+#
+
+__version__ = "0.0.1"
+
+from lightly_train._modules.teachers.dinov2.build_teacher import get_dinov2_teacher
+
+__all__ = ["get_dinov2_teacher"]
```
guarin:
```suggestion
```

Same as comment above. I wouldn't add `__version__` as it is not used anywhere.

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
```diff
@@ -0,0 +1,78 @@
...
+    # Load config
+    config_path = get_config_path(config_name)
+    cfg = load_and_merge_config(config_path)
+
+    # Build model
+    model, _, embed_dim = build_model_from_cfg(cfg)
+    model.eval()
+
+    # Cache checkpoint
+    cache_dir = Path.home() / ".cache" / "lightlytrain" / "checkpoints"
```
guarin:
```suggestion
    cache_dir = Path.home() / ".cache" / "lightly-train" / "checkpoints"
```
Would use the same name as the package name here.

Can you add a TODO to move the cache dir definition to a more prominent location. Maybe `lightly_train/_data/cache.py` could be a good location. I would expect that we want to access this location from different parts of the code.

src/lightly_train/_modules/teachers/build_teacher.py
```diff
@@ -0,0 +1,24 @@
...
+
+    Returns the model in eval mode along with its embedding dimension.
+    Raises a ValueError if the teacher name is unknown.
+    """
+    if teacher_name.startswith("dinov2_"):
+        from lightly_train._modules.teachers.dinov2 import get_dinov2_teacher
+
+        return get_dinov2_teacher(teacher_name)
+    else:
+        raise ValueError(f"Unknown teacher: {teacher_name}")
```
guarin:
```suggestion
        raise ValueError(f"Unknown teacher: '{teacher_name}'")
```

This is a minor thing but is helpful if the string you print contains whitespaces. If you don't put quotes the error messages can become quite confusing. Same for the other messages in this PR.

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
guarin: Agree

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
```diff
@@ -0,0 +1,78 @@
...
+        logger.info(f"Using cached teacher weights from: {checkpoint_path}")
+
+    ckpt = torch.load(checkpoint_path, map_location="cpu")
+    model.load_state_dict(ckpt, strict=True)
+    logger.info(f"Loaded teacher weights from {checkpoint_path}")
+
+    return model, embed_dim
+
+
+def get_config_path(config_name: str) -> str:
```
guarin:
```suggestion
def get_config_path(config_name: str) -> Path:
```

Minor: Better to return the path if you already have it. You can always convert it to a string where needed.

src/lightly_train/_modules/teachers/dinov2/configs/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+
+from omegaconf import OmegaConf
+
+
+def load_config(config_name: str):
+    config_filename = config_name + ".yaml"
+    return OmegaConf.load(pathlib.Path(__file__).parent.resolve() / config_filename)
+
+
+dinov2_default_config = load_config("ssl_default_config")
```
guarin: Let's load this only inside the function and not when the file is imported.

src/lightly_train/_modules/teachers/dinov2/layers/block.py
```diff
@@ -0,0 +1,294 @@
...
+# LICENSE file in the root directory of this source tree.
+#
+
+# References:
+#   https://github.com/facebookresearch/dino/blob/master/vision_transformer.py
+#   https://github.com/rwightman/pytorch-image-models/tree/master/timm/layers/patch_embed.py
+
+import logging
+import os
+from typing import Any, Callable, Dict, List, Tuple
```
guarin, line 13:
```suggestion
from typing import Any, Callable
```

You can use `dict, list, tuple` when importing `from __future__ import annotations`

tests/_methods/distillation/test_distillation.py
```diff
@@ -175,18 +174,19 @@ def test_queue_update(self, mocker: MockerFixture) -> None:
...
+        mock_get_teacher_model.return_value = (None, teacher_embed_dim)
+
+        # Instantiate the distillation method.
         distill = Distillation(
             method_args=DistillationArgs(queue_size=queue_size),
             optimizer_args=DistillationLARSArgs(),
             embedding_model=helpers.get_embedding_model(),
             global_batch_size=batch_size,
         )
-        mock_automodel.from_pretrained.assert_called_once()
```
guarin, line 189: Good practice is to call `mock_get_teacher_model.assert_called_once()` here. This helps a lot to make sure the mocking works correctly.

tests/_methods/distillation/test_distillation.py
```diff
@@ -220,18 +220,19 @@ def test_teacher_queue_never_exceeds_capacity(self, mocker: MockerFixture) -> No
...
```
guarin, line 234: Same as above.

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
guarin: We should also allow users to configure the cache dir via a `LIGHTLY_TRAIN_CACHE_DIR` environment variable. I would already add it in this PR in case someone needs it.

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
stegmuel: Done.

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
```diff
@@ -0,0 +1,78 @@
...
+    model, _, embed_dim = build_model_from_cfg(cfg)
+    model.eval()
+
+    # Cache checkpoint
+    cache_dir = Path.home() / ".cache" / "lightlytrain" / "checkpoints"
+    cache_dir.mkdir(parents=True, exist_ok=True)
+    checkpoint_path = cache_dir / Path(url).name
+
+    if not checkpoint_path.exists():
+        logger.info(f"Downloading teacher weights from: {url}")
```
guarin:
```suggestion
        logger.info(f"Downloading teacher weights from: '{url}' and saving them to: '{checkpoint_path}'. The cache directory location can be configured with the LIGHTLY_TRAIN_CACHE_DIR environment variable.")
```

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
```diff
@@ -0,0 +1,78 @@
...
+        urlretrieve(url, checkpoint_path)
```
guarin: Add TODO to:
- add timeout in case the download gets stuck or is extremely slow
- add progress bar
- move download function to commonly accessible location like `lightly_train/_data/download.py`

src/lightly_train/_methods/distillation/distillation.py
guarin: Long term we should have the same interface as for the student models. It is already designed to support intermediate layers. For full DINOv2 training support we anyways need a DINOv2 ViT with this interface.

stegmuel: If we want the spatial features in the near future we can do so via `get_intermediate_layers()` without having to change anything in the `teachers` folder.

src/lightly_train/_modules/teachers/__init__.py
stegmuel: Changed that.

src/lightly_train/_modules/teachers/dinov2/__init__.py
stegmuel: Done.

src/lightly_train/_modules/teachers/build_teacher.py
stegmuel: Done.

tests/_methods/distillation/test_distillation.py
stegmuel: Done.

tests/_methods/distillation/test_distillation.py
stegmuel: Done.

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
JonasWurst: I would recommend to wrap a try block in case the download does not succeed

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
stegmuel: Done.

src/lightly_train/_modules/teachers/dinov2/configs/__init__.py
stegmuel: Done.

src/lightly_train/_modules/teachers/dinov2/layers/block.py
stegmuel: These are part of the files that we decided to copy-paste as is. Do we want to go there ?

guarin: Ah no then leave it as is 👍🏼

src/lightly_train/_modules/teachers/build_teacher.py
stegmuel: Done.

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
stegmuel: Done.

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
stegmuel: Done.

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
stegmuel: Done.

README.md
```diff
@@ -203,6 +203,17 @@ Lightly**Train** offers flexible licensing options to suit your specific needs:
...
+
+This repository includes code from the following open-source projects:
+
+- **DINOv2**
+  - Source: [facebookresearch/dinov2](https://github.com/facebookresearch/dinov2)
+  - License: Apache 2.0
+  - License file: [`licenses/DINOv2_LICENSE`](licences/DINOv2_LICENSE)
+
+All third-party code is used in accordance with their respective licenses.
+
```
guarin: Let's add this to a NOTICE file instead of the README. Otherwise the README will get overloaded if we add more dependencies.

Example NOTICE file: https://github.com/enonic/xp/blob/master/NOTICE.txt
You can keep the licenses separate in the `licenses/` directory as is.

src/lightly_train/_methods/distillation/distillation.py
```diff
@@ -81,7 +77,7 @@ def resolve_auto(
             )
         elif self.queue_size >= scaling_info.dataset_size:
             raise ValueError(
-                f"The specified queue size ({self.queue_size}) cannot be larger than the dataset size ({scaling_info.dataset_size})."
+                f"The specified queue size ('{self.queue_size}') cannot be larger than the dataset size ('{scaling_info.dataset_size}')."
```
guarin: Nit: No need to put quotes `''` for numbers.

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # Cache the teacher checkpoint.
+    checkpoint_path = checkpoint_dir / Path(url).name
+
+    if not checkpoint_path.exists():
+        logger.info(
+            f"Downloading teacher weights from: '{url}' and saving them to: '{checkpoint_path}'. "
+            "The cache directory location can be configured with the LIGHTLY_TRAIN_CACHE_DIR environment variable."
+        )
+        download_from_url(url, checkpoint_path, timeout=10.0)
```
guarin:
```suggestion
        download_from_url(url, checkpoint_path, timeout=180.0)
```

Lets give some slack for slow internet connections

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
```diff
@@ -0,0 +1,80 @@
...
+    else:
+        logger.info(f"Using cached teacher weights from: '{checkpoint_path}'")
+
+    ckpt = torch.load(checkpoint_path, map_location="cpu")
+    model.load_state_dict(ckpt, strict=True)
+    logger.info(f"Loaded teacher weights from '{checkpoint_path}'")
```
guarin: I would either drop this message or change it to debug level. Otherwise we send nearly the same information multiple times to the terminal.

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
```diff
@@ -0,0 +1,80 @@
...
+from lightly_train._modules.teachers.dinov2.models import build_model_from_cfg
+
+logger = logging.getLogger(__name__)
+
+
+TEACHER_MODELS = {
+    "dinov2_vitb14": {
+        "url": "https://dl.fbaipublicfiles.com/dinov2/dinov2_vitb14/dinov2_vitb14_pretrain.pth",
+        "config": "eval/vitb14_pretrain",
+    },
```
guarin, line 31: Will we add the other pretrained models as well?

src/lightly_train/_data/download.py
```diff
@@ -0,0 +1,37 @@
...
+                with open(destination, "wb") as out_file:
+                    while True:
+                        chunk = response.read(8192)
+                        if not chunk:
+                            break
+                        out_file.write(chunk)
+                        t.update(len(chunk))
+    except (URLError, HTTPError, TimeoutError, Exception) as e:
+        logger.error(f"Failed to download weights from {url}: {e}")
+        raise RuntimeError(f"Download failed: {e}")
```
guarin:
```suggestion
        raise RuntimeError(f"Download from '{url}' failed: {e}")
```

Exceptions are automatically logged as error

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
```diff
@@ -0,0 +1,80 @@
...
+    if not checkpoint_path.exists():
+        logger.info(
+            f"Downloading teacher weights from: '{url}' and saving them to: '{checkpoint_path}'. "
+            "The cache directory location can be configured with the LIGHTLY_TRAIN_CACHE_DIR environment variable."
+        )
+        download_from_url(url, checkpoint_path, timeout=10.0)
+    else:
+        logger.info(f"Using cached teacher weights from: '{checkpoint_path}'")
+
+    ckpt = torch.load(checkpoint_path, map_location="cpu")
```
guarin:
```suggestion
    ckpt = torch.load(checkpoint_path, map_location="cpu", weights_only=True)
```

Otherwise the following warning is shown in the terminal:
```
/Users/guarin/lightly-train/src/lightly_train/_modules/teachers/dinov2/build_teacher.py:67: FutureWarning: You are using `torch.load` with `weights_only=False` (the current default value), which uses the default pickle module implicitly. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling (See https://github.com/pytorch/pytorch/blob/main/SECURITY.md#untrusted-models for more details). In a future release, the default value for `weights_only` will be flipped to `True`. This limits the functions that could be executed during unpickling. Arbitrary objects will no longer be allowed to be loaded via this mode unless they are explicitly allowlisted by the user via `torch.serialization.add_safe_globals`. We recommend you start setting `weights_only=True` for any use case where you don't have full control of the loaded file. Please open an issue on GitHub for any issues related to this experimental feature.
  ckpt = torch.load(checkpoint_path, map_location="cpu")
```

src/lightly_train/_modules/teachers/dinov2/layers/attention.py
```diff
@@ -0,0 +1,90 @@
...
+
+XFORMERS_ENABLED = os.environ.get("XFORMERS_DISABLED") is None
+try:
+    if XFORMERS_ENABLED:
+        from xformers.ops import memory_efficient_attention, unbind
+
+        XFORMERS_AVAILABLE = True
+        logger.warning("xFormers is available (Attention)")
+    else:
+        logger.warning("xFormers is disabled (Attention)")
```
guarin: I think we should change the logging behavior here. We shouldn't send a warning to users if xFormers is available. If it is available we can log it with `logger.debug`.

If it is not available we should log it as `xFormers is not available. This can slow down XYZ etc... For faster training install it with 'pip install XYZ'`

Same for the other places where the xFormer checks are done.

Lets create a follow-up issue to handle xFormers properly.
* Should we install it by default?
* Should we add it as optional dependency to pyproject.toml?
* What are valid versions?
* Log version here: https://github.com/lightly-ai/lightly-train/blob/697aadc92e2199207ae25c61352aa4a448b1dcba/src/lightly_train/_system.py#L72-L77

guarin (commented):
Left some minor comments but otherwise the code LGMT! Tested it and everything worked smoothly :)

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
```diff
@@ -69,10 +71,10 @@ def get_dinov2_teacher(teacher_name: str) -> Module:
     return model
 
 
-def get_config_path(config_name: str) -> str:
+def get_config_path(config_name: str) -> Path:
```
JonasWurst, line 86:
```suggestion
def get_config_path(config_name: Path) -> Path:
```

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
```diff
@@ -43,22 +41,26 @@ def get_dinov2_teacher(teacher_name: str) -> Module:
     url = teacher_info["url"]
     config_name = teacher_info["config"]
 
-    # Load config
+    # Load config.
     config_path = get_config_path(config_name)
-    cfg = load_and_merge_config(config_path)
+    cfg = load_and_merge_config(str(config_path))
```
JonasWurst, line 58:
```suggestion
    cfg = load_and_merge_config(config_path)
```

JonasWurst (commented):
Small changes

README.md
stegmuel: Done.

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
stegmuel: If my understanding is correct timeout time is per package and not for the full file so that is already long. But I'll increase the time.

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
stegmuel: Done.

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
stegmuel: `config_name` is defined as `str` in `TEACHER_MODELS` dict, do we want to change that?

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
stegmuel: `load_and_merge_config` is part of the code taken as is from DINOv2 so I think it can't/shouldn't be changed.

src/lightly_train/_modules/teachers/dinov2/layers/attention.py
stegmuel: Created the issue.

src/lightly_train/_data/download.py
stegmuel: Done.

src/lightly_train/_methods/distillation/distillation.py
stegmuel: Removed.

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
stegmuel: Added.

src/lightly_train/_modules/teachers/dinov2/build_teacher.py
stegmuel: Moved to debug.

CHANGELOG.md
```diff
@@ -13,10 +13,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Distillation uses the official DINOv2 implementation for the teacher model.
+
 ### Deprecated
 
 ### Removed
 
+Dependency on the transformers library.
```
```suggestion
- Dependency on the transformers library.
```

README.md
```diff
@@ -203,16 +203,9 @@ Lightly**Train** offers flexible licensing options to suit your specific needs:
...
 
-This repository includes code from the following open-source projects:
-
-- **DINOv2**
-  - Source: [facebookresearch/dinov2](https://github.com/facebookresearch/dinov2)
-  - License: Apache 2.0
-  - License file: [`licenses/DINOv2_LICENSE`](licences/DINOv2_LICENSE)
-
-All third-party code is used in accordance with their respective licenses.
+See [`NOTICE`](./NOTICE) for a list of third-party libraries and their respective licenses.
```
guarin: We can drop this completely from the README. The NOTICE file is sufficient.

src/lightly_train/_modules/teachers/dinov2/layers/attention.py
guarin: Ah sorry, my message wasn't clear. We should change the logging behavior now and make a follow-up to address the points listed at the end of my comment.

guarin (approved):
LGTM! Feel free to resolve and merge

CHANGELOG.md
stegmuel: Done.

README.md
stegmuel: Removed.

PR #7, opened 2025-04-11
Issue templates

JonasWurst:
## What has changed and why?

This pr brings issue templates that can be used for creating issues. This way, we can try to enforce some basic information to be provided.

- [ ] Yes

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 [![Google Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/lightly-ai/lightly-train/blob/main/examples/notebooks/quick_start.ipynb)
 [![Python](https://img.shields.io/badge/Python-3.8%7C3.9%7C3.10%7C3.11%7C3.12-blue.svg)](https://docs.lightly.ai/train/stable/installation.html)
-[![OS](https://img.shields.io/badge/OS-Linux%7CMacOS%7CWindows%20%28soon%21%29-blue.svg)](https://docs.lightly.ai/train/stable/installation.html)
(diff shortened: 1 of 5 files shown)


.github/ISSUE_TEMPLATE/docs_improvement.md
```diff
@@ -0,0 +1,20 @@
...
+- File/Section:
+- Link (if applicable):
+
+### ✏️ Suggested Improvement
+
+What could be improved, corrected, or added?
+
+### 💬 Additional Notes
+
+Any context or references to clarify the documentation issue.
```
liopeer, line 20: Can we get an additional section here that asks whether they would be able/willing to contribute? Such that they can just set a checkmark.

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,23 @@
...
+
+What should the feature do? How would it work?
+
+### 🛠 Alternatives you've considered
+
+Have you considered any alternative solutions?
+
+### 📝 Additional context
+
+Add any other context, designs, or screenshots here.
```
liopeer, line 23: Same here: Checkmark if willing/able to make contribution.

pyproject.toml
```diff
@@ -43,6 +43,7 @@ dependencies = [
 dev = [
     "licenseheaders>=0.8.8",
     "mdformat>=0.7.21; python_version>='3.9'",  # Older versions change formatting behavior
+    "mdformat-frontmatter>=2.0.0",
```
liopeer, line 46: Does `mdformat` fail without this? If `mdformat` works, but just does not format correctly we should not add this, even if it's easy to remove again (since it's a dev dependency).

JonasWurst: Yes, the test are failing without it.

PR #21, opened 2025-04-15
Thomas trn 271 remove hf warning

stegmuel:
## What has changed and why?

Added a filter on the warnings to prevent HuggingFace warnings to show up in Colab.

## How has it been tested?

It has been tested that the warnings are not depicted anymore in Colab.

- [ ] Yes


src/lightly_train/_commands/_warnings.py
```diff
@@ -89,3 +89,11 @@ def filter_warnings() -> None:
     warnings.filterwarnings(
         "ignore", message="Grad strides do not match bucket view strides"
     )
+
+
+def filter_huggingface_warnings() -> None:
```
JonasWurst, line 94: can't it be part of the filter_train_warnings --> no separate call required?

stegmuel: The call has to be done before the import of AutoModel I think otherwise it is too late.

src/lightly_train/_methods/distillation/distillation.py
```diff
@@ -37,6 +35,12 @@
 )
 from lightly_train.types import Batch
 
+filter_huggingface_warnings()
```
JonasWurst, line 39: Please add the todo note also here, taht we have to change it back

stegmuel: Done.

PR #28, opened 2025-04-17
Add version compatibility table to docs

guarin:
## What has changed and why?

* Add version compatibility table to docs

I also moved the platform compatibility down and changed it to a table. This way both sections have the same structure. I moved them below the installation instructions as the installation instruction is the most important part and the two tables would take too much space above it.

## How has it been tested?

[Installation - LightlyTrain documentation version table.pdf](https://github.com/user-attachments/files/19796349/Installation.-.LightlyTrain.documentation.version.table.pdf)

- [x] Yes


yutong-xiang-97 (approved):
LGTM

PR #38, opened 2025-04-25
Update memory mapped datasets

guarin:
## What has changed and why?

* Create single memory mapped dataset files with SLURM and Windows
* Check that `out` path is identical on all ranks

We now create only a single mmap file on all platforms and in all distributed settings (SLURM, torch.distributed, lightning, ...). This speeds up dataset initialization a lot as we only iterate once over the dataset to initialize it. Before this was done on every rank individually.

I could make a follow-up PR to handle cases where we can recover from different out paths on different processes. We can do this if the processes are initialized after the training script is called. In that case we can share the out path from rank 0 to all other ranks. This could avoid errors in many cases.

## How has it been tested?

Tested on Mac and Linux (with and without SLURM):
* `out = Path("out/test") / datetime.now().strftime("%Y-%m-%d_%H-%M-%S")` --> Raises error as expected
* `out = Path("out/test")` -> No error and only a single mmap file is created. Verified that files are cleaned up after run.

- [x] Yes

### DIFF
--- a/src/lightly_train/_commands/common_helpers.py
+++ b/src/lightly_train/_commands/common_helpers.py
@@ -8,14 +8,15 @@
 from __future__ import annotations
 
 import contextlib
+import hashlib
 import json
 import logging
 import os
-import platform
+import tempfile
+import time
 import warnings
 from enum import Enum
 from pathlib import Path
-from tempfile import NamedTemporaryFile
 from typing import Any, Generator, Iterable, Literal
 
 import torch
@@ -23,9 +24,6 @@ from pytorch_lightning.accelerators.accelerator import Accelerator
 from pytorch_lightning.accelerators.cpu import CPUAccelerator
 from pytorch_lightning.accelerators.cuda import CUDAAccelerator
 from pytorch_lightning.accelerators.mps import MPSAccelerator
-from pytorch_lightning.plugins.environments import (  # type: ignore[attr-defined]
-    SLURMEnvironment,
-)
 from pytorch_lightning.strategies.strategy import Strategy
 from torch.nn import Module
 from torch.utils.data import Dataset
@@ -89,7 +87,7 @@ def get_accelerator(
         return CPUAccelerator()
 
 
-def _get_rank() -> int | None:
+def get_rank() -> int | None:
     """Get the rank of the current process.
 
     Copied from https://github.com/Lightning-AI/pytorch-lightning/blob/06a8d5bf33faf0a4f9a24207ae77b439354350af/src/lightning/fabric/utilities/rank_zero.py#L39-L49
@@ -107,7 +105,7 @@ def _get_rank() -> int | None:
 
 def is_rank_zero() -> bool:
     """Check if the current process is running on the first device."""
-    local_rank = _get_rank()
+    local_rank = get_rank()
     return local_rank == 0 or local_rank is None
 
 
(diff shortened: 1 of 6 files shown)


src/lightly_train/_commands/common_helpers.py
```diff
@@ -129,6 +126,78 @@ def get_out_dir(out: PathLike, resume: bool, overwrite: bool) -> Path:
...
+    in the path that is generated inside the training script. For example with:
+    >>> out = f"out/{datetime.now().strftime('%Y-%m-%d_%H-%M-%S')}"
+    This will result in different paths for each rank as the timestamp is generated
+    on each rank separately.
+    """
+    out_dir = Path(out).resolve()
+    out_tmp = get_verify_out_tmp_dir() / get_sha256(out_dir)
+    logger.debug(f"Creating temporary file '{out_tmp}' to verify out path.")
+
+    LIGHTLY_TRAIN_VERIFY_OUT_DIR_TIMEOUT_SEC = (
```
liopeer: I would move this to the top of the file, together with the 30s default timeout.

src/lightly_train/_commands/common_helpers.py
```diff
@@ -275,30 +335,46 @@ def get_dataset_mmap_filenames(
...
-        return memory_mapped_sequence.memory_mapped_sequence_from_filenames(
-            filenames=filenames,
-            mmap_filepath=mmap_filepath,
-        )
-    else:
-        if common_helpers.is_rank_zero():
-            # Save filenames to memory mapped file and return them.
-            return memory_mapped_sequence.memory_mapped_sequence_from_filenames(
+    tmp_path = mmap_filepath.with_suffix(".temp")
+    LIGHTLY_TRAIN_MMAP_TIMEOUT_SEC = "LIGHTLY_TRAIN_MMAP_TIMEOUT_SEC"
```
liopeer: Would also move to top of file.

src/lightly_train/_commands/common_helpers.py
```diff
@@ -275,30 +335,46 @@ def get_dataset_mmap_filenames(
...
+        if is_rank_zero():
+            # Save filenames to temporary file. Create the final file only once rank zero has
+            # finished writing all the filenames.
+            memory_mapped_sequence.write_filenames_to_file(
                 filenames=filenames,
-                mmap_filepath=mmap_filepath,
+                mmap_filepath=tmp_path,
             )
+            # Rename the temporary file to mmap_filepath.
+            tmp_path.replace(mmap_filepath.resolve())
```
liopeer, line 389: Nice solution!

tests/_commands/test_common_helpers.py
```diff
@@ -119,6 +119,59 @@ def test_get_out_dir__nonempty(
...
+    assert not any(common_helpers.get_verify_out_tmp_dir().iterdir())
+
+
+def test_verify_out_dir_equal_on_all_ranks__no_rank0(
+    tmp_path: Path, mocker: MockerFixture
+) -> None:
+    out_dir = tmp_path / "rank1"
+
+    mocker.patch.dict(
+        os.environ, {"RANK": "1", "LIGHTLY_TRAIN_VERIFY_OUT_DIR_TIMEOUT_SEC": "1"}
```
liopeer: One second is quite long for a unit test. Can we make this shorter or are we limited to integers?

liopeer (commented):
Looking really solid overall, but I would like to verify 2 things before merging:
 - **Windows compatibility** (since you change some code that had a Windows specific dispatch). My suspicion is that it might not work, as mentioned in the (deleted) comment: "On Windows, each process creates its own temporary file due to file handling restrictions."
 - **Multi-node training.** I don't think it would work for multi-node training as it is right now: How we currently check the rank is via https://github.com/lightly-ai/lightly-train/blob/ed3f1d208590c6822073b466e167cc680ad69b9b/src/lightly_train/_commands/common_helpers.py#L92 and this uses the env variables `"RANK", "LOCAL_RANK", "SLURM_PROCID", "JSM_NAMESPACE_RANK"`. AFAIK `"RANK"` is the global rank, so if `"RANK"` exists only the first process on the first node will create the file (instead of every first rank of each node). Also `"SLURM_PROCID"` seems to be the global rank and not the local rank.

src/lightly_train/_commands/common_helpers.py
guarin: I moved the env variables to `_constants.py`. Then everything is in the same place.

tests/_commands/test_common_helpers.py
guarin: Reduced to 0.1 sec

src/lightly_train/_commands/common_helpers.py
```diff
@@ -8,29 +8,31 @@
...
-from pytorch_lightning.plugins.environments import (  # type: ignore[attr-defined]
-    SLURMEnvironment,
-)
 from pytorch_lightning.strategies.strategy import Strategy
 from torch.nn import Module
 from torch.utils.data import Dataset
 
-from lightly_train._commands import common_helpers
+from lightly_train._constants import (
+    LIGHTLY_TRAIN_MASK_DIR,
```
liopeer, line 32: much cleaner, thanks!

liopeer (approved):
LGTM! Thanks for adding the Windows support again with the `_unlink_and_ignore` and now it's also really clean with the rank (local/global) checks, which can now also be used in potentional other places if necessary.

guarin:
Added an extra check to verify that `is_global_rank` only returns True if `node_rank` is zero or not set. This is an issue when you start the processes like this and `RANK` is not yet available when the function is called:

Node 0:
```
MASTER_PORT=3456 MASTER_ADDR=127.0.0.1 WORLD_SIZE=2 NODE_RANK=0 lightly-train train \
	out=./out/test_$(date +%Y%m%d_%H%M) \
	data=my_data_dir \
	model=torchvision/resnet18 \
	method=simclr \
	num_workers=2 \
	num_nodes=2 \
	devices=1 \
	accelerator=cpu
```

Node 1:
```
MASTER_PORT=3456 MASTER_ADDR=127.0.0.1 WORLD_SIZE=2 NODE_RANK=1 lightly-train train \
	out=./out/test_$(date +%Y%m%d_%H%M) \
	data=my_data_dir \
	model=torchvision/resnet18 \
	method=simclr \
	num_workers=2 \
	num_nodes=2 \
	devices=1 \
	accelerator=cpu
```


I also verified that each node creates its own mmap and verify files:
```
/var/folders/_9/smcq487j71v_2y8gxdt31ykr0000gn/T/lightly-train
├── data
│   ├── 22a7c3c6c6a7dd21ea6de77f4cb204bc1cf15e1cdb1757b4699e97300709db76.mmap
│   └── f7e59ab0b0625d621ec46aabca0f64486575433e2407e220d51346571a1397bb.mmap
└── verify-out
    ├── 22a7c3c6c6a7dd21ea6de77f4cb204bc1cf15e1cdb1757b4699e97300709db76
    └── f7e59ab0b0625d621ec46aabca0f64486575433e2407e220d51346571a1397bb
```

PR #45, opened 2025-04-30
Trn 330 export model each epoch

JonasWurst:
## What has changed and why?
In the past a model was only exported after a training successfully finished. With this pr, the model is exported after each epoch ensuring easy accessibility for users by not requiring them to convert from the checkpoints.

since the export methods of the packages did prompt the usage example, this had to be modified to not flood the logs.

## How has it been tested?
the exported ckpts were successfully loaded.

- [x] Yes


src/lightly_train/_callbacks/callback_helpers.py
```diff
@@ -48,6 +49,7 @@ def get_callbacks(
 ) -> list[Callback]:
     callbacks: list[Callback] = []
     callbacks.append(DataWaitTQDMProgressBar())
+    callbacks.append(ModelExport(model, out_dir=out))
```
liopeer:
```suggestion
    callbacks.append(ModelExport(model=model, out_dir=out))
```

src/lightly_train/_callbacks/export.py
```diff
@@ -0,0 +1,39 @@
...
+#
+from __future__ import annotations
+
+from pathlib import Path
+
+from pytorch_lightning import LightningModule, Trainer
+from pytorch_lightning.callbacks import Callback
+from torch.nn import Module
+
+from lightly_train._commands.common_helpers import ModelFormat, export_model
```
liopeer:
```suggestion
from lightly_train._commands import common_helpers
from lightly_train._commands.common_helpers import ModelFormat
```

src/lightly_train/_callbacks/export.py
```diff
@@ -0,0 +1,39 @@
...
+        self._model = model
+        self._out_dir = out_dir
+
+    def on_train_epoch_end(self, trainer: Trainer, pl_module: LightningModule) -> None:
+        # Delete the previous export if it exists
+        export_path = self._out_dir / "exported_models" / "exported_last.pt"
+        if export_path.exists():
+            export_path.unlink()
+
+        export_model(
```
liopeer:
```suggestion
        common_helpers.export_model(
```

src/lightly_train/_callbacks/callback_helpers.py
JonasWurst: adressed

src/lightly_train/_callbacks/export.py
JonasWurst: adressed

src/lightly_train/_callbacks/export.py
JonasWurst: addressed

liopeer (approved):
LGTM! Thanks for creating the `ModelExportArgs`. Makes it robust for future additions of configurations. 👍

src/lightly_train/_callbacks/export.py
```diff
@@ -0,0 +1,50 @@
...
+from pytorch_lightning.callbacks import Callback
+from torch.nn import Module
+
+from lightly_train._commands import common_helpers
+from lightly_train._commands.common_helpers import ModelFormat
+from lightly_train._configs.config import PydanticConfig
+
+
+class ModelExportArgs(PydanticConfig):
+    every_n_epochs: int | None = None
```
guarin: Shouldn't this be:
```suggestion
    every_n_epochs: int = 1
```

Then it is clear that it is exported after every epoch by default.

JonasWurst: yes, I only took inspiration from how it was done in checkpoint:    [every_n_epochs: int | None = None](https://github.com/lightly-ai/lightly-train/blob/ccc8ed10406918974ab2c96004207255f5531149/src/lightly_train/_callbacks/checkpoint.py#L33)

But I'm happy to change it, since it would also remove some logic.

PR #47, opened 2025-05-01
Unify environment variables

guarin:
## What has changed and why?

* Moves all environment variables and their default values to the same place

## How has it been tested?

* Added unit tests

- [ ] Yes


src/lightly_train/_modules/teachers/dinov2/models/__init__.py
```diff
@@ -17,7 +17,7 @@ def build_model(
 ) -> Union[Tuple[Module, int], Tuple[Module, Module, int]]:
     suffix = "_memeff"
     if args.arch.endswith(suffix):
-        args.arch = args.arch[:-len(suffix)]
+        args.arch = args.arch[: -len(suffix)]
```
liopeer, line 20: Assuming this is caused by the formatting. Would be great to quickly hint at this in the PR descriptition.

liopeer (approved):
LGTM!

src/lightly_train/_modules/teachers/dinov2/models/__init__.py
guarin: Ah yes, not sure why this was formatted. Might have an outdated ruff version.

PR #50, opened 2025-05-02
Change to Lightly cache in dockerfile and in makefile

JonasWurst:
## What has changed and why?
With the latest changes from #3 , the caching of the hugging-face dinov2 model is not required anymore. Instead we need to cache the model from the facebook implementation. This PR changes the required cache dir setup in the docker image and the download of the model in the docker creation in the makefile.

- [ ] Yes


guarin (approved):
LGTM!

PR #53, opened 2025-05-02
Improved logging for xformers

JonasWurst:
## What has changed and why?

The logging around xFormers availability was overloaded. Hence in this PR all logging related to the availability of xFormers is moved to a single place (vision transformer). 

## How has it been tested?

Pending

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_modules/teachers/dinov2/models/vision_transformer.py
+++ b/src/lightly_train/_modules/teachers/dinov2/models/vision_transformer.py
@@ -9,13 +9,16 @@
 #   https://github.com/facebookresearch/dino/blob/main/vision_transformer.py
 #   https://github.com/rwightman/pytorch-image-models/tree/master/timm/models/vision_transformer.py
 
+import importlib
 import logging
 import math
+import os
 from functools import partial
 from typing import Callable, Sequence, Tuple, Union
 
 import torch
 import torch.nn as nn
+from pytorch_lightning.utilities import rank_zero_only
 from torch.nn.init import trunc_normal_
 
 from lightly_train._modules.teachers.dinov2.layers import (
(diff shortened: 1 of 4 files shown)


src/lightly_train/_modules/teachers/dinov2/models/vision_transformer.py
```diff
@@ -30,6 +33,23 @@
 
 logger = logging.getLogger(__name__)
 
+XFORMERS_INSTALLED = importlib.util.find_spec("xformers") is not None
```
liopeer: I think our go-to way is using `try` and `except ImportError` statements. It is imo a bit safer, since it not only checks if it is installed, but also whether it can be imported. Particularly C++/CUDA libraries often cause issues only upon import.

liopeer (commented):
Small comment, the rest looks good!

src/lightly_train/_modules/teachers/dinov2/models/vision_transformer.py
guarin: Either try/except or RequirementCache, see here for an example: https://github.com/lightly-ai/lightly-train/blob/3aca12f29a441a23eaac47fe784708e7aa38d37d/src/lightly_train/_models/ultralytics/ultralytics.py#L47-L49

JonasWurst: The static-checks are failing with the try block since the xformers is not sued in this highlevel module itself... I'll go ahead with RequirementCache

src/lightly_train/_modules/teachers/dinov2/models/vision_transformer.py
```diff
@@ -30,6 +32,19 @@
 
 logger = logging.getLogger(__name__)
 
+XFORMERS_INSTALLED = RequirementCache("xformers>=0.0.18") is not None
```
liopeer:
```suggestion
XFORMERS_INSTALLED = RequirementCache("xformers>=0.0.18")
```
A `RequirementCache` object should never be `None`. It behaves like a boolean.

liopeer (changes requested):
Thanks for adding the `RequirementCache`! Since we now check the availability of `xformers` (and whether it is enabled) in 3 files, I would recommend moving those constants into a submodule `_modules/teachers/dinov2/_dependency.py` and use them from there. This will make the imports a bit cleaner.

src/lightly_train/_modules/teachers/dinov2/models/vision_transformer.py
JonasWurst: yep, god catch. Forgot to remove it.

liopeer (approved):
LGTM!

PR #52, opened 2025-05-02
add unit tests and get_model

liopeer:
## What has changed and why?
- add get_model to FeatureExtractor interface
- implement get_model method for all models

## How has it been tested?
- added unit tests for all models and also added some other missing unit tests for torchvision

## Did you update [CHANGELOG.md](../CHANGELOG.md)?

- [ ] Yes
- [x] Not needed (internal change)

## Did you update the documentation?

- [ ] Yes
- [x] Not needed (internal change without effects for user)

PR #60, opened 2025-05-07
Trn 396 dinov2 allow vit model training

JonasWurst:
## What has changed and why?
This PR brings a new model package that wraps the dinov2_vit.

## How has it been tested?
Unit tests are in place

- [ ] Yes

- [ ] Yes


JonasWurst:
NOTE: exporting and testing of it, not yet done.

src/lightly_train/_models/dinov2_vit/dinov2_vit_package.py
```diff
@@ -0,0 +1,105 @@
...
+        return DINOv2ViTModelWrapper(model)
+
+    @classmethod
+    def export_model(cls, model: Module, out: Path, log_example: bool = True) -> None:
+        torch.save(model.state_dict(), out)
+        #TODO: finish implement export_model
+
+# Create singleton instance of the package. The singleton should be used whenever
+# possible.
+DINOV2_VIT_Package_PACKAGE = DINOv2ViTPackage()
```
guarin:
```suggestion
DINOV2_VIT_PACKAGE = DINOv2ViTPackage()
```

src/lightly_train/_models/dinov2_vit/dinov2_vit.py
```diff
@@ -0,0 +1,63 @@
...
+    def forward_pool(self, x: ForwardFeaturesOutput) -> ForwardPoolOutput:
+        return {"pooled_features": self._pool(x["features"])}
+
+    def make_teacher(self) -> None:
+        if self._model.chunked_blocks:
+            for chunked_blocks in self._model.blocks:
+                update_blocks_student_to_teacher(chunked_blocks)
+        else:
+            update_blocks_student_to_teacher(self._model.blocks)             
+
```
guarin: Please add `get_model` function as in https://github.com/lightly-ai/lightly-train/pull/52

guarin (changes requested):

tests/_models/dinov2_vit/test_dinov2_vit.py
```diff
@@ -0,0 +1,97 @@
...
+from lightly_train._modules.teachers.dinov2.layers.drop_path import DropPath
+from lightly_train._modules.teachers.dinov2.models.vision_transformer import (
+    DinoVisionTransformer,
+)
+from lightly_train._modules.teachers.dinov2.models.vision_transformer import (
+    vit_small as vit_small_untyped,
+)
+
+
+def vit_small() -> DinoVisionTransformer:
```
yutong-xiang-97: Should we alternatively add typing to the original functions?

src/lightly_train/_models/dinov2_vit/dinov2_vit.py
```diff
@@ -0,0 +1,71 @@
...
+            raise ValueError(
+                f"Unexpected shape for x_norm_patchtokens: {rt['x_norm_patchtokens'].shape}"
+            )
+        return {"features": features_reshaped, "cls_token": rt["x_norm_clstoken"]}
+    
+    def forward_pool(self, x: ForwardFeaturesOutput) -> ForwardPoolOutput:
+        return {"pooled_features": self._pool(x["features"])}
+    
+    def get_model(self) -> DinoVisionTransformer:
+        return cast(DinoVisionTransformer, self._model)
```
yutong-xiang-97: Personally, I would prefer
```
model: DinoVisionTransformer = self._model
return model
```
instead of casting to be more type safe

src/lightly_train/_models/dinov2_vit/dinov2_vit.py
```diff
@@ -0,0 +1,71 @@
...
+        super().__init__()
+        self._model = model
+        self._feature_dim = int(self._model.embed_dim)
+        self._pool = AdaptiveAvgPool2d((1, 1))
+    
+    def feature_dim(self) -> int:
+        return self._feature_dim
+
+    def forward_features(self, x: Tensor) -> ForwardFeaturesOutput:
+        rt = self._model(x, is_training=True) # forcing to return all patches
```
yutong-xiang-97: Personally I would prefer directly using
```
self._model.forward_features()
```
https://github.com/lightly-ai/lightly-train/blob/35f3004f495430b2056eb1ecc4fa11052ea6547c/src/lightly_train/_modules/teachers/dinov2/models/vision_transformer.py#L308C9-L308C25

src/lightly_train/_models/dinov2_vit/dinov2_vit.py
```diff
@@ -0,0 +1,71 @@
...
+        if rt["x_norm_patchtokens"].dim() == 3:
+            patches_resolution = self._model.patch_embed.patches_resolution
+            features_reshaped = rt["x_norm_patchtokens"].reshape(
+                rt["x_norm_patchtokens"].shape[0],
+                rt["x_norm_patchtokens"].shape[2],
+                patches_resolution[0],
+                patches_resolution[1],
+            )
+        elif rt["x_norm_patchtokens"].dim() == 4:
```
yutong-xiang-97, line 41: Just curious: in which case would the dim be 4?

src/lightly_train/_models/dinov2_vit/dinov2_vit_package.py
```diff
@@ -0,0 +1,121 @@
...
+            "vit_giant2": vits.vit_giant2,
+        }
+        model_builder = model_builders.get(cfg.student.arch, None)
+        if model_builder is None:
+            raise TypeError(
+                f"Unsupported architecture type {cfg.student.arch}."
+            )
+        
+        # Cast the model builder to the correct type
+        model_builder = cast(
```
yutong-xiang-97: Same question as before, should we cast it dynamically or we use static type checking for the model_builder functions?

src/lightly_train/_models/dinov2_vit/dinov2_vit_package.py
```diff
@@ -0,0 +1,121 @@
...
+            Callable[...,DinoVisionTransformer],
+            model_builder
+        )
+        
+        model = model_builder(
+            img_size=cfg.crops.global_crops_size,
```
yutong-xiang-97, line 70: The `DinoVisionTransformer` has more init params than listed here. Is it deliberate that we fixed the ones that are configurable and use the defaults for the rest? If not, I think using `**kwargs` makes sense here instead of using exact match of kwargs.

src/lightly_train/_modules/teachers/dinov2/configs/__init__.py
```diff
@@ -9,6 +9,39 @@
 
 from omegaconf import OmegaConf
 
+MODELS = {
```
yutong-xiang-97, line 16: I don't see serious issues with putting the config loading logic in `__init__.py` for now, esp. when it is fixed, but `__init__.py` was not supposed to handle code as it might have some implications.

src/lightly_train/_models/dinov2_vit/dinov2_vit.py
JonasWurst: Cast is simply telling for unytped vars, what to expect. So in some place, we have to perform the cast, if we are not changing the typing in the base implementation. So far we decided to not dive into that rabbit hole.

src/lightly_train/_models/dinov2_vit/dinov2_vit.py
JonasWurst: If already reshaped

tests/_models/dinov2_vit/test_dinov2_vit.py
JonasWurst: Initially I refrained doin it, since we also did not touch the vit code in the previous PRs. However, I decided it is simpler to break that rule, so the remaining code becomes simpler.

src/lightly_train/_models/dinov2_vit/dinov2_vit.py
JonasWurst: See comment above, I've changed the type in vit code, hence cast is not required anymore

src/lightly_train/_models/dinov2_vit/dinov2_vit.py
JonasWurst: I personally see no benefit in doing it. The typical thing which is used is the forward. I would keep it as is

src/lightly_train/_models/dinov2_vit/dinov2_vit_package.py
JonasWurst: resolved, since took the liberty to change the vit code

src/lightly_train/_models/dinov2_vit/dinov2_vit_package.py
JonasWurst: I'm now combining the two dicts, however I think it is not a good idea to play around with that and overwrite the cfg.

tests/_models/dinov2_vit/test_dinov2_vit.py
guarin: > Initially I refrained doin it, since we also did not touch the vit code in the previous PRs. However, I decided it is simpler to break that rule, so the remaining code becomes simpler.

Please add disclaimers in the code in case you modify any code that is not from us

src/lightly_train/_modules/teachers/dinov2/configs/__init__.py
JonasWurst: This is again down to not refactoring the vit codebase much. They already had logic related to cfg loading. Hence that was the best suited place w/o messing with the codebase.

But in general, I agree we should keep it clean.

tests/_models/dinov2_vit/test_dinov2_vit.py
JonasWurst: Last time I've changed things, I'v mentioned it in the Notice file. Is that sufficient in your opinion?

guarin: If it is bigger things (add/remove/modify function) I would add a note in the code.

src/lightly_train/_models/dinov2_vit/dinov2_vit.py
yutong-xiang-97: In which case is it necessary to reshape the features? In original DINOv2 it is used directedly: https://github.com/facebookresearch/dinov2/blob/e1277af2ba9496fbadf7aec6eba56e8d882d1e35/dinov2/train/ssl_meta_arch.py#L165

src/lightly_train/_models/dinov2_vit/dinov2_vit_package.py
JonasWurst: Remark: the kwargs matching is copied from the original codebase

yutong-xiang-97 (approved):

PR #62, opened 2025-05-08
Harmonize ModelWrapper Interface

liopeer:
## What has changed and why?
- Splits the package interface into two different ones: `LimitedPackage` if it can't implement methods such as `get_model_wrapper` and `FrameworkPackage` for the full interface. This fixes logic, where the `CustomPackage` has to implement methods that make no sense.
- allow `str | ModelWrapper` in the `lightly_train.train` API, wrap the model immediately and prefer passing the wrapped model to downstream functions (and then use its `get_model()` method where necessary).

## How has it been tested?
 - currently only passes static checking

- [ ] Yes

- [ ] Yes


src/lightly_train/_commands/train.py
```diff
@@ -46,7 +46,7 @@
 def train(
     out: PathLike,
     data: PathLike | Sequence[PathLike],
-    model: str | Module,
+    model: str | ModelWrapper,
```
guarin:
```suggestion
    model: str | ModelWrapper | Any,
```

We need `Any` because some packages do not inherit from any base class. For example RF-DETR.

src/lightly_train/_commands/train.py
```diff
@@ -250,11 +250,12 @@ def train_from_config(config: TrainConfig) -> None:
         scaling_info = train_helpers.get_scaling_info(
             dataset=dataset, epochs=config.epochs
         )
-        model_instance = package_helpers.get_model(
+        wrapped_model = package_helpers.get_wrapped_model(
             model=config.model, model_args=config.model_args
         )
+        model = wrapped_model.get_model()
```
guarin:
```suggestion
        model_instance = wrapped_model.get_model()
```
Otherwise there might be confusion between the input `model` argument and the instance.

src/lightly_train/_commands/train.py
```diff
@@ -341,7 +343,7 @@ def train_from_config(config: TrainConfig) -> None:
         train_helpers.load_checkpoint(
             checkpoint=config.checkpoint,
             resume=config.resume,
-            model=model_instance,
+            model=model,
```
guarin: I think this we'll also have to refactor to take a `model_wrapper` as input. Maybe we can do it in a follow-up PR?

src/lightly_train/_commands/train.py
```diff
@@ -356,7 +358,7 @@ def train_from_config(config: TrainConfig) -> None:
         trainer_instance.save_checkpoint(out_dir / "checkpoints" / "last.ckpt")
     logger.info("Training completed.")
     common_helpers.export_model(
-        model=model_instance,
+        model=model,
```
guarin: This we should also refactor to take `model_wrapper` instead. Can also be done in follow-up

src/lightly_train/_commands/train.py
```diff
@@ -373,7 +375,7 @@ def train_from_dictconfig(config: DictConfig) -> None:
 class TrainConfig(PydanticConfig):
     out: PathLike
     data: PathLike | Sequence[PathLike]
-    model: str | Module
+    model: str | ModelWrapper
     method: str = "distillation"
```
guarin:
```suggestion
    model: str | ModelWrapper | Any
    method: str = "distillation"
```

src/lightly_train/_models/package.py
```diff
@@ -7,38 +7,37 @@
...
 
-class Package(ABC):
-    """Interface for a package that provides models and feature extractors that are
-    compatible with lightly_train.
-
-    Every package must implement this interface.
-    """
-
+@runtime_checkable
+class LimitedPackage(Protocol):
```
guarin: Why Protocol instead of ABC? ABC provides better type-checking and runtime check guarantees. As all packages are implemented internally and not exposed to users we have full control and can benefit more from ABCs than Protocols.

src/lightly_train/_models/package.py
```diff
@@ -7,38 +7,37 @@
...
     name: str  # The name of the package.
 
     @classmethod
-    @abstractmethod
-    def list_model_names(cls) -> list[str]:
-        """List all supported models by this package."""
-        ...
+    def export_model(
+        cls, model: Module, out: Path, log_example: bool = True
```
guarin, line 25:
```suggestion
        cls, model_wrapper: ModelWrapper, out: Path, log_example: bool = True
```

src/lightly_train/_models/package.py
guarin: Should we call the classes `BasePackage` and `Package`? `FrameworkPackage` is a tautology.

src/lightly_train/_models/package_helpers.py
```diff
@@ -58,7 +65,20 @@ def list_model_names() -> list[str]:
...
+
+def get_wrapped_model(
+    model: str | ModelWrapper, model_args: dict[str, Any] | None = None
+) -> ModelWrapper:
+    """Returns a wrapped model instance given a model name or instance."""
+    if isinstance(model, ModelWrapper):
+        return model
+
+    model = get_model(model=model, model_args=model_args)
+    return get_model_wrapper(model=model)
```
guarin: This is confusing, we have a function called `get_wrapped_model` and one called `get_model_wrapper`. I wonder whether we can refactor `get_model_wrapper`, `get_wrapped_model`, and `get_model` into a single `get_wrapped_model` function. All the remaining logic would either be in this function or the respective Package implementations.

tests/_callbacks/test_callback_helpers.py
```diff
@@ -69,7 +69,8 @@ def test_get_callbacks__default(tmp_path: Path) -> None:
     callbacks = callback_helpers.get_callbacks(
         callback_args=callback_args,
         out=tmp_path,
-        model=model,
+        wrapped_model=model,
+        model=model.get_model(),
```
guarin:
```suggestion
        wrapped_model=model,
```

We can call `wrapped_model.get_model()` inside the function.

src/lightly_train/_commands/train.py
liopeer: Yes, did not want to include this right now, but that would be added later.

src/lightly_train/_commands/train.py
liopeer: I would actually prefer the name `package_model` for any underlying model. Wdyt?

src/lightly_train/_commands/train.py
liopeer: Yes, did not want to change too much now, but I'd refactor this as well for sure.

src/lightly_train/_models/package.py
liopeer: Wasn't happy either with my naming choices. Your suggestion would definitely be better.

src/lightly_train/_callbacks/callback_helpers.py
```diff
@@ -62,7 +62,7 @@ def get_callbacks(
     if callback_args.model_export is not None:
         callbacks.append(
             ModelExport(
-                model=model,
+                wrapped_model=wrapped_model,
                 out_dir=out / "exported_models",
                 **callback_args.model_export.model_dump(),
             )
```
liopeer, line 67: See below: `ModelCheckpoint` and `CheckpointLightlyTrainModels` is currently really hard to convert to the model wrapper. I would not do this here.

src/lightly_train/_models/package.py
liopeer: This is unfortunately really difficult to change, that's why I left it. Reason: The `export` command expects to be able to use this directly from the model that's included in the checkpoint. Wrapping it again before exporting would not be very smart.

src/lightly_train/_callbacks/checkpoint.py
```diff
@@ -95,7 +95,9 @@ def on_load_checkpoint(
             )
         else:
             # Load the state dicts from the checkpoint in-place.
-            self._models.model.load_state_dict(_checkpoint.models.model.state_dict())
```
guarin, line 98: If we keep `model` in the checkpoint we should also load its state dict. IMO we can drop it completely from the checkpoint.

src/lightly_train/_commands/train.py
```diff
@@ -250,11 +250,14 @@ def train_from_config(config: TrainConfig) -> None:
...
         )
-        model_instance = package_helpers.get_model(
-            model=config.model, model_args=config.model_args
-        )
+        if not isinstance(config.model, ModelWrapper):
+            wrapped_model = package_helpers.get_wrapped_model(
+                model=config.model, model_args=config.model_args
+            )
+        else:
+            wrapped_model = config.model
```
guarin: Could we move the if/else logic into the `package_helpers.get_wrapped_model` function? Then it is a single function call and doesn't bloat the main `train_from_config` function that much.

src/lightly_train/_models/custom/custom_package.py
```diff
@@ -9,42 +9,24 @@
...
-    @classmethod
-    def get_model_wrapper(cls, model: Module) -> ModelWrapper:
-        if not isinstance(model, ModelWrapper):
-            raise TypeError(
-                "Unsupported model type: Model does not implement FeatureExtractor interface."
-            )
-        return model
+        """Check if the model is supported by this package."""
+        # Custom package supports all models.
+        return isinstance(model, Module)
```
guarin:
```suggestion
        return isinstance(model, ModelWrapper)
```
This must check if the model implements the `ModelWrapper` interface.

src/lightly_train/_models/embedding_model.py
```diff
@@ -39,7 +39,7 @@ def __init__(
                 the features are embedded and returned without pooling.
         """
         super().__init__()
-        self.model_wrapper = model_wrapper
+        self.model_wrapper = wrapped_model
```
guarin:
```suggestion
        self.wrapped_model = wrapped_model
```

src/lightly_train/_models/model_wrapper.py
```diff
@@ -7,11 +7,19 @@
...
 
-from typing import Protocol, runtime_checkable
+from typing import (
+    Any,
+    Dict,
+    Iterator,
+    Mapping,
+    Protocol,
+    overload,
+    runtime_checkable,
```
guarin, line 17:
```suggestion
    Any,
    Iterator,
    Mapping,
    Protocol,
    overload,
    runtime_checkable,
```

Use `dict` instead

src/lightly_train/_models/package.py
guarin: IMO it only makes sense to pass a ModelWrapper here. We'll then call `get_model` internally and save the model. I don't think the package should implement saving either EmbeddingModel or ModelWrapper itself as those classes are defined by LightlyTrain.

src/lightly_train/_models/package_helpers.py
```diff
@@ -40,10 +40,14 @@ def list_packages() -> list[Package]:
     ]
 
 
-def get_package(package_name: str) -> Package:
+def list_full_packages() -> list[Package]:
```
guarin:
```suggestion
def list_packages() -> list[Package]:
```

src/lightly_train/_models/package_helpers.py
```diff
@@ -58,32 +62,28 @@ def list_model_names() -> list[str]:
...
 
-    raise UnknownModelError(f"Unknown model: '{model.__class__.__name__}'")
+def get_wrapped_model(
+    model: str, model_args: dict[str, Any] | None = None
+) -> ModelWrapper:
+    """Returns a wrapped model instance given a model name or instance."""
+    package_name, model_name = _parse_model_name(model)
+    package = get_package(package_name)
+    # This can never be a BasePackage, because calling get_wrapped_model() makes no sense
+    # for a BasePackage.
```
guarin:
```suggestion
    package = get_package(package_name)
    # This can never be a BasePackage, because calling get_wrapped_model() makes no sense
    # for a BasePackage.
```

This looks to me like `get_package` must have return type `Package` instead of `BasePackage`.

tests/_callbacks/test_callback_helpers.py
```diff
@@ -120,23 +117,6 @@ def test_get_callbacks__user_config(tmp_path: Path) -> None:
...
-            return torch.zeros(1)
-
-        def forward_pool(self, x: Tensor) -> Tensor:
-            return torch.zeros(1)
-
     model = DummyCustomModel()
-    checkpoint_model = callback_helpers.get_checkpoint_model(model=model)
-    assert model == checkpoint_model
-    assert isinstance(checkpoint_model, DummyCustomModel)
+    checkpoint_model = model.get_model()
```
guarin: Can't we remove `test_get_checkpoint_model` completely as `get_checkpoint_model` was removed?

guarin:
This PR or a follow-up has to take https://github.com/lightly-ai/lightly-train/pull/70 into account

src/lightly_train/_commands/train.py
liopeer: Completely removed anyway from the train command.

src/lightly_train/_commands/train.py
liopeer: changed

src/lightly_train/_commands/train.py
liopeer: done

src/lightly_train/_callbacks/checkpoint.py
liopeer: Thanks for catching this! I wanted to drop it as well, but how it is currently with `ModelPart` and the export functions it would require a lot of changes: The export is currently set up such that we export a `ModelPart` (usually `model` or `embedding_model` from the checkpoint). If we now change `ModelPart` to be `wrapped_model` and `embedding_model` we could not use the same API to export the true `model` anymore.
In short: `ModelPart` and the `CheckpointLightlyTrainModels` are synced in their attributes. We would need to change both and additionally write logic that still let's us export the true `model`, which is not included directly in either.

src/lightly_train/_commands/train.py
liopeer: Done!

src/lightly_train/_models/custom/custom_package.py
liopeer: I would honestly refrain from doing that. Reasoning: All other packages run that check on the underlying model (i.e. the model returned by `wrapped_model.get_model()`). Either we change it for all packages or none. But I would do this separately. The package interface is already confusing enough as it is, since the custom package is not capable of fully implementing it.

src/lightly_train/_models/embedding_model.py
liopeer: Thanks!

src/lightly_train/_models/model_wrapper.py
liopeer: Unfortunately you can only use `dict` with the `from __future__ import annotations` in Python < 3.9 inside type hints, but not to define types, which we do in the `TypeVar`. Same issue btw in `lightly_train._types` where we have to use `Dict` as well.

src/lightly_train/_models/package_helpers.py
liopeer: True, thanks for the catch.

src/lightly_train/_callbacks/checkpoint.py
liopeer: But I am now definitely loading the state dict.

src/lightly_train/_commands/train.py
```diff
@@ -330,7 +330,7 @@ def train_from_config(config: TrainConfig) -> None:
             method_args=config.method_args,
             scaling_info=scaling_info,
             optimizer_args=config.optim_args,
-            model=model_instance,
+            model=wrapped_model.get_model(),
```
liopeer, line 334: Here we should also pass the wrapped model. Will be done in a follow up and is related to #70

src/lightly_train/_models/package.py
liopeer: See reasoning above. I agree, but it would require quite a few changes more.

src/lightly_train/_models/super_gradients/super_gradients.py
```diff
@@ -7,18 +7,16 @@
...
-from abc import ABC, abstractmethod
+from typing import Protocol
 
 from torch.nn import Module
 
 from lightly_train._models.model_wrapper import ModelWrapper
 
 
-class SuperGradientsModelWrapper(ABC, ModelWrapper):
+class SuperGradientsModelWrapper(ModelWrapper, Protocol):
```
liopeer, line 17: Had to change this since ModelWrapper is also a Protocol.

tests/_callbacks/test_callback_helpers.py
liopeer: Done!

tests/_models/super_gradients/test_segmentation_module.py
```diff
@@ -42,7 +42,7 @@ def test_forward_pool(self) -> None:
         fe = SegmentationModuleFeatureExtractor(model)
         x = torch.rand(1, 3, 224, 224)
         out = fe.forward_pool({"features": x})["pooled_features"]
-        assert out.shape == (1, 1024)
+        assert out.shape == (1, 3, 1, 1)
```
liopeer, line 45: Quickyl fixed this unit test. The result is now correct according to the corresponding protocol:
```python
@runtime_checkable
class ForwardPool(Protocol):
    def forward_pool(self, x: ForwardFeaturesOutput) -> ForwardPoolOutput:
        """Pools features, should be called after `forward_features`.

        Args:
            x:
                Output of `forward_features` method. Must be a dict with a "features"
                entry containing the extracted features with shape
                (B, feature_dim, H_in, W_in).

        Returns:
            Dict with "pooled_features" entry containing the pooled features with shape
            (B, feature_dim, H_out, W_out). H_out and W_out depend on the pooling
            strategy but are usually 1.
        """
        ...
```

liopeer: I wonder why this was never caught before. Those should run in our nightly tests.

src/lightly_train/_callbacks/export.py
```diff
@@ -26,11 +26,11 @@ class ModelExportArgs(PydanticConfig):
 class ModelExport(Callback):
     def __init__(
         self,
-        model: Module,
+        wrapped_model: ModelWrapper,
         out_dir: Path,
         every_n_epochs: int = 1,
     ):
-        self._model = model
+        self._model = wrapped_model
```
guarin: Nit:

```suggestion
        self._wrapped_model = wrapped_model
```

src/lightly_train/_callbacks/checkpoint.py
guarin: Sounds good, let's address this in a follow-up. As long as `wrapped_model` is in `CheckPointLightlyTrainModels` we can always access `model` as well with `wrapped_model.get_model()`. Note that we always store the full model instance and not only the state dict in the checkpoints.

src/lightly_train/_commands/train_helpers.py
```diff
@@ -395,18 +396,21 @@ def load_checkpoint(
...
 def load_state_dict(
-    model: Module, embedding_model: EmbeddingModel, method: Method, checkpoint: PathLike
+    wrapped_model: ModelWrapper,
+    embedding_model: EmbeddingModel,
+    method: Method,
+    checkpoint: PathLike,
 ) -> None:
     ckpt = Checkpoint.from_path(Path(checkpoint))
-    model.load_state_dict(ckpt.lightly_train.models.model.state_dict())
+    wrapped_model.load_state_dict(ckpt.lightly_train.models.wrapped_model.state_dict())
```
guarin, line 413: Hmm this looks dangerous because we no longer load the state dict of the actual model. So some things are loaded from the checkpoint but not all. I suggest the following as a quick fix:
```suggestion
    wrapped_model.load_state_dict(ckpt.lightly_train.models.wrapped_model.state_dict())
    model = wrapped_model.get_model()
    if isinstance(model, Module):
        model.load_state_dict(ckpt.lightly_train.models.model.state_dict())
```

This doesn't handle the cases where model doesn't inherit from `Module` though. So in the long term we need a solution that can handle both. It seems to me we need an `export_model` and corresponding `import_model` for all models. For custom models we can handle it automatically if they inherit from `Module`. If they don't we should raise an error and let the user know that they have to implement these extra methods.

src/lightly_train/_models/custom/custom_package.py
guarin: Hmm doesn't `get_package_from_model` in `package_helpers.py` not work anymore? Because it would then return the `CustomPackage` for any models inheriting `Module` but we actually require models to implement `ModelWrapper`.

tests/_commands/test_export.py
```diff
@@ -217,15 +236,23 @@ def test_export__parameters() -> None:
...
+@pytest.mark.parametrize(
+    "part",
+    [
+        "model",
+        "wrapped_model",
+        "embedding_model",
+    ],
+)
+def test_export_from_dictconfig(tmp_path: Path, part: str) -> None:
     ckpt_path, ckpt = _get_checkpoint(tmp_path)
```
guarin, line 248: Nice!

tests/_models/super_gradients/test_segmentation_module.py
guarin: Good catch!

guarin (commented):
This looks much better than we had before!

src/lightly_train/_commands/train_helpers.py
liopeer: Ah right. I must have not paid attention to the fact that not all underlying `model`s inherit from `Module`. If they did, then for sure the `wrapped_model` would include all the parameters of the `model` as well. Your solution is indeed the safest for now.

And your suggestion seems reasonable. I think we can definitely constrain custom models to inherit from Module. Most will anyway.

src/lightly_train/_callbacks/export.py
liopeer: Done

liopeer: Thx! Should definitely change it while we're at it.

src/lightly_train/_models/custom/custom_package.py
liopeer: So, now it should work!

guarin (approved):
LGTM!

src/lightly_train/_commands/train.py
liopeer: issue created

PR #63, opened 2025-05-08
Distillationv2 interface

stegmuel:
## What has changed and why?

Made DistillationV2 visible.

## How has it been tested?

Updated existing test to check the list of method and the mapping from method names to method classes.

- [x] Yes

- [x] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support lists of files and folders as input to the `embed` and `train` commands.
 - Faster dataset initialization with SLURM and Windows.
 - Model is exported after every n-th epoch instead of only at the end. Default n=1
+- Add DistillationV2 to the methods.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support lists of files and folders as input to the `embed` and `train` commands.
 - Faster dataset initialization with SLURM and Windows.
 - Model is exported after every n-th epoch instead of only at the end. Default n=1
-- Add `DistillationV2` to the methods.
-- `DistillationV2` is used with `method="distillation"` or `method="distillationv2"`.
+- New, improved distillationv2 method that achieves higher accuracy and trains up to 3x faster than distillationv1.
+  The new method is selected as default by LightlyTrain with `method="distillation"`. The old distillation method
+  can still be used with `method="distillationv1"` for backwards compatibility.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,12 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Support lists of files and folders as input to the `embed` and `train` commands.
-- Faster dataset initialization with SLURM and Windows.
-- Model is exported after every n-th epoch instead of only at the end. Default n=1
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support lists of files and folders as input to the `embed` and `train` commands.
 - Faster dataset initialization with SLURM and Windows.
 - Model is exported after every n-th epoch instead of only at the end. Default n=1
+- Add DistillationV2 to the methods.
```
guarin: We should also mention that it is now the default distillation method that is selected with `method="distillation"`. For the old method `method="distillationv1"` must be used.

tests/_methods/test_method_helpers.py
```diff
@@ -38,6 +41,7 @@ def test_list_methods_private() -> None:
         "densecl",
         "dino",
         "distillation",
+        "distillationv2",
         "simclr",
```
guarin, line 46: I think `distillationv1` must also be returned from here. Otherwise you cannot set `method="distillationv1"` when training.

tests/_methods/test_method_helpers.py
```diff
@@ -46,5 +50,6 @@ def test_list_methods_public() -> None:
     assert method_helpers.list_methods() == [
         "dino",
         "distillation",
+        "distillationv2",
```
guarin, line 55: Should we only return `distillation` here? This functions returns the methods "available" for users. We can either treat distillationv1/v2 as hidden methods or not. IMO it doesn't really matter but both should behave the same way.

tests/_methods/test_method_helpers.py
stegmuel: Updated.

tests/_methods/test_method_helpers.py
stegmuel: Updated.

CHANGELOG.md
```diff
@@ -12,11 +12,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support lists of files and folders as input to the `embed` and `train` commands.
 - Faster dataset initialization with SLURM and Windows.
 - Model is exported after every n-th epoch instead of only at the end. Default n=1
+- Add `DistillationV2` to the methods.
+- `DistillationV2` is used with `method="distillation"` or `method="distillationv2"`.
```
JonasWurst, line 16: This statement seems to wrong

```suggestion
- New, improved distillationv2 method that achieves higher accuracy and trains up to 3x faster than distillationv1.
  The new method is selected as default by LightlyTrain with `method="distillation"`. The old distillation method
  can still be used with `method="distillationv1"` for backwards compatibility.
```

I would also put this entry at the top of the list as it is the main feature for the new release.

docs/source/methods/distillation.md
```diff
@@ -40,3 +40,8 @@ The following are the default augmentation settings for Distillation. To learn h
 
 ```{include} _auto/distillation_transform_args.md
 ```
+
+## Note
+
+Starting from **LightlyTrain 0.7**, `method="distillation"` uses a new, improved `v2` implementation.\
+The previous version is still available via `method="distillationv1"` for backward compatibility.
```
guarin:
````suggestion
```{note}
Starting from **LightlyTrain 0.7.0***, `method="distillation"` uses a new, improved `v2` implementation
that achieves higher accuracy and trains up to 3x faster. The previous version is still available via
`method="distillationv1"` for backward compatibility.
```
````

Lets put this at the top right after the `# Distillation (recommended)` intro paragraph.

guarin (approved):

CHANGELOG.md
guarin: Could you put this at the top of the changelog as it is the main feature for the new release?

PR #64, opened 2025-05-09
Test checkpoint loading

guarin:
## What has changed and why?

* Add test that checkpoints are loaded correctly

## How has it been tested?

* Updated unit tests
* Made test fail by introducing wrong checkpoint loading that was fixed in a previous PR

- [ ] Yes

- [ ] Yes

### DIFF
--- a/tests/_commands/test_train.py
+++ b/tests/_commands/test_train.py
@@ -88,6 +88,8 @@ def test_train(
 
     # Check that we can resume training
     last_ckpt_path = out / "checkpoints" / "last.ckpt"
+    first_ckpt = Checkpoint.from_path(checkpoint=last_ckpt_path)
+
     with caplog.at_level(logging.INFO):
         train.train(
             out=out,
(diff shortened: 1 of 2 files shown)


tests/_commands/test_train.py
```diff
@@ -388,8 +414,10 @@ def test_train__checkpoint(mocker: MockerFixture, tmp_path: Path) -> None:
         num_workers=0,
         epochs=0,
         accelerator="cpu",
+        devices=1,
```
guarin, line 420: Fix devices to not accidentally test on multi-GPU

tests/_commands/test_train.py
```diff
@@ -106,6 +108,27 @@ def test_train(
...
+    # Check that exported checkpoint weights changed between first and second run.
+    second_ckpt = Checkpoint.from_path(checkpoint=last_ckpt_path)
+    first_state_dict = first_ckpt.lightly_train.models.model.state_dict()
+    second_state_dict = second_ckpt.lightly_train.models.model.state_dict()
+    assert first_state_dict.keys() == second_state_dict.keys()
+    for key in first_state_dict.keys():
+        if key.startswith("fc."):
+            # Skip the last layer as it is not pretrained.
+            continue
+        assert not torch.equal(first_state_dict[key], second_state_dict[key])
```
JonasWurst, line 120: In the unit tests for the export we used `torch.allclose`. I'm fine with having this more strict check, but wanted to raise awareness about this.

JonasWurst (approved):
LGTM

Do we need a separate test, when using distillation? Since we should not saving/loading the teacher?

tests/_commands/test_train.py
guarin: Good point, I think stricter is better. Updated the export tests to `torch.equal`.

PR #67, opened 2025-05-13
Export and remove models from rank 0 only

stegmuel:
## What has changed and why?

Ensures the model is only exported and removed from rank 0.

## How has it been tested?

Successfully runs for 100 epochs on 4 GPUs without crashing.

- [ ] Yes

- [ ] Yes


JonasWurst (approved):
LGTM

PR #74, opened 2025-05-15
Add Windows CI to standard unit tests

liopeer:
## What has changed and why?
 - add Windows unit tests to PR CI
 - fixed a few things with logging
 - fixed a unit test concerning logging that cause permission issues on Windows
 - added unit tests for file paths (was also an issue under Windows)

- [ ] Yes

- [ ] Yes


src/lightly_train/_data/image_dataset.py
```diff
@@ -80,7 +80,8 @@ def list_image_filenames(
             The root directory to scan for images.
 
     Returns:
-        An iterable of image filenames relative to `image_dir`.
+        An iterable of image filenames relative to `image_dir` or absolute paths
```
liopeer, line 92: This was not properly documented until now. Also extended the corresponding unit tests.

tests/test__logging.py
```diff
@@ -78,19 +77,55 @@ def test__remove_handlers_by_type() -> None:
 
 
 def test_set_up_file_logging() -> None:
-    with tempfile.NamedTemporaryFile() as file:
```
liopeer, line 81: This test had to be adapted in order to be Windows compatible.

src/lightly_train/_logging.py
```diff
@@ -26,6 +26,7 @@
 
 # Set up the logger for the lightly_train package.
 lightly_logger = logging.getLogger("lightly_train")
+lightly_logger.setLevel(logging.DEBUG)
```
liopeer, line 29: Has to be global. `test_set_up_file_logging` fails otherwise when ran as a single test, since it does not get properly set up.

src/lightly_train/_logging.py
```diff
@@ -157,7 +157,7 @@ def set_up_file_logging(log_file_path: Path) -> None:
     ]
 
     for file_logger in file_loggers:
-        _remove_handlers(file_logger, LightlyTrainRotatingFileHandler)
+        _remove_handlers(file_logger, logging.FileHandler)
```
guarin: Why is this change needed? We added `LightlyTrainRotatingFileHandler` to have a special class just for Lightly and not remove any other handlers that users might have attached.

liopeer: That was indeed something that I forgot to clean up. I initially changed quite a few things until I realized that the main issue is the setup of the `lightly_logger`, which should be global.

liopeer: It's reverted now

guarin (approved):

PR #76, opened 2025-05-16
Fixed bug getting pool layer with timm

Uzarel:
## What has changed and why?

When loading a timm-wrapped model, it fails if the `global_pool` layer is within `model.head`.

## How has it been tested?

If you now try to use `convnext_base` as a timm model, it now works, while it didn't before.

- [ ] Yes

- [ ] Yes


stegmuel (approved):
Thanks for your contribution.
Tried the changes and it does indeed fixes the problem!

PR #79, opened 2025-05-19
Refactor teachers to vit_package

JonasWurst:
## What has changed and why?

In this refactoring PR, the distillation teacher loading is switched to the newly available dinov2_vit package loading to remove redundant code. 

## How has it been tested?
Unit test 

- [ ] Yes

- [ ] Yes


src/lightly_train/_models/dinov2_vit/dinov2_vit_package.py
```diff
@@ -54,13 +53,17 @@ def updated_cfg_to_match_pretrained(
...
+    # check if model_name has _pretrain suffix
+    if not model_name.endswith("_pretrain"):
+        model_name_pretrained = model_name + "_pretrain"
+
+    if model_name_pretrained not in VIT_MODELS:
         logger.warning(
-            f"Model {model_name} does not have a pretrained version."
+            f"Model {model_name_pretrained} does not have a pretrained version."
             "The non pretrained version will be used."
         )
```
guarin: This seems risky, we should raise an error instead and print the list of available pretrained models.

src/lightly_train/_methods/distillationv2/distillationv2.py
```diff
@@ -36,14 +36,23 @@
 logger = logging.getLogger(__name__)
 
 
+def get_teacher(teacher_name: str) -> Module:
+    wrapped_model = package_helpers.get_wrapped_model(
+        model=teacher_name, model_args={"pretrained": True}
```
guarin: Can we drop the `pretrained` argument as it is now already encoded in the `_pretrain` suffix in the model name? I would prefer if there is only a single way to load pretrained models.
```suggestion
        model=teacher_name
```

src/lightly_train/_models/dinov2_vit/dinov2_vit_package.py
```diff
@@ -99,8 +102,11 @@ def get_model(
         # Get the model cfg and update if required.
         config_path = get_config_path(VIT_MODELS[model_name]["config"])
         cfg = load_and_merge_config(str(config_path))
+        url = VIT_MODELS[model_name]["url"]
 
-        if load_pretrained:
+        if load_pretrained and not model_name.endswith("_pretrain"):
```
guarin: Can we simplify the logic here as well and ignore `model_args={"pretrained": True}` and only rely on the `_pretrained` suffix in the model name instead?

src/lightly_train/_models/dinov2_vit/dinov2_vit_package.py
JonasWurst: adressed

src/lightly_train/_methods/distillationv2/distillationv2.py
JonasWurst: adressed

src/lightly_train/_models/dinov2_vit/dinov2_vit_package.py
JonasWurst: adressed

guarin (approved):
LGTM!

PR #80, opened 2025-05-20
Override the loading of checkpoints to allow missing teacher keys.

stegmuel:
## What has changed and why?

Override the loading of checkpoint in distillation methods to allow missing teacher-related keys.

## How has it been tested?

Tried continuing pre-training from checkpoints with both distillation methods.
Two tests are added to verify that the distillation methods correctly load  from checkpoints.

- [ ] Yes

- [ ] Yes


guarin (commented):
Code LGTM!

Could you add a unit test?

src/lightly_train/_methods/distillation/distillation.py
```diff
@@ -267,3 +267,24 @@ def on_load_checkpoint(self, checkpoint: dict[str, Any]) -> None:
...
+            state_dict, strict=False
+        )
+
+        # Filter out teacher-related keys from the list of missing keys.
+        missing_keys = [
+            k for k in missing_keys if not k.startswith("teacher_embedding_model.")
+        ]
+
+        # No key should be missing besides the ones related to the teacher model.
+        if missing_keys or unexpected_keys:
```
guarin: This should only raise an error if `strict=True`

stegmuel: Updated.

guarin (approved):
LGTM

PR #81, opened 2025-05-20
Document method args

guarin:
## What has changed and why?

* Document how to overwrite method args
* Add default method args for all methods

I didn't create a new page for the method args as there is not much content. 

## How has it been tested?
[Train - LightlyTrain documentation method args.pdf](https://github.com/user-attachments/files/20344869/Train.-.LightlyTrain.documentation.method.args.pdf)
[DINO - LightlyTrain documentation _13.pdf](https://github.com/user-attachments/files/20344870/DINO.-.LightlyTrain.documentation._13.pdf)

- [ ] Yes

- [x] Yes


docs/prebuild.py
```diff
@@ -43,26 +38,41 @@ def build_changelog_html(source_dir: Path) -> None:
...
-        ("simclr", SimCLRTransformArgs),
-        ("distillation", DistillationTransformArgs),
-    ]:
-        # dump transform args
-        args = common_helpers.pretty_format_args(transform_args().model_dump())
+def dump_transform_args_for_methods(dest_dir: Path) -> None:
+    dest_dir.mkdir(parents=True, exist_ok=True)
+    for method in method_helpers.list_methods():
+        if method in {"distillationv1", "distillationv2"}:
+            continue
```
guarin, line 45: Changed code to automatically include new methods

docs/source/train/index.md
```diff
@@ -290,18 +290,71 @@ resolution of 224x224 pixels is used. A custom resolution can be set like this:
...
+lightly-train train out="out/my_experiment" data="my_data_dir" model="torchvision/resnet18" transform_args.image_size="[448,448]"
+```
 ````
 
 ```{warning}
 Not all models support all image sizes.
 ```
 
+### Image Transforms
+
```
yutong-xiang-97, line 316: Suggestion: here we could also note the user that in most cases, it is not necessary to modify the default transform args, like in below for the method args

yutong-xiang-97 (approved):

docs/source/train/index.md
guarin: Good point, there is actually already a warning on the image transforms page. This is why I didn't add another one here as well.

PR #82, opened 2025-05-20
DINOv2 Method

yutong-xiang-97:
## What has changed and why?

Added DINOv2 method based on the original DINOv2 impl.

#### Added `dinov2.py`
- Adapted from https://github.com/facebookresearch/dinov2/blob/main/dinov2/train/ssl_meta_arch.py#L372
- init `method_args` by `resolve_auto` based on the model config 

#### Added `dinov2_optim.py`

- Two `OptimArgs` classes for ViT small-base and large-giant respectively (to be continued)
- Layerwise LR decay and weight decay cancellation, adapted from https://github.com/facebookresearch/dinov2/blob/main/dinov2/utils/param_groups.py

#### Added `utils.py`
- Masking, adapted from https://github.com/facebookresearch/dinov2/blob/main/dinov2/data/masking.py and https://github.com/facebookresearch/dinov2/blob/main/dinov2/data/collate.py
- `linear_warmup_schedule` copy-catted from LightlySSL for DINO loss' teacher temp

#### Modified `dinov2_loss.py`

- Input type based on original DINOv2 impl.

#### Modified `dinov2_transform.py`

- Two `TransformArgs` classes for ViT small-base and large-giant respectively (to be continued)

### Notes:

- `n_local_crops` is set to 8 manually. It should match with ` DINOv2.transform_cls().transform_args_cls().transform_args.local_view.num_views`
- All parameters follow the original config file atm. The param scaling is temporarily removed
- In `_forward_teacher()` and `_forward_student_global()` when we use identical DINO and iBOT heads, we concatenate the cls and patch tokens, put them in a buffer, and do one forward pass. This follows the original DINOv2 impl. except that we did a separate pass with `_forward_student_local()` so that we could allow using global crops only, as compared to the original impl. where all student output are concatenated and passed in one go.
- Some of the vars are renamed to make the code clearer and more concise.

### TODOs

- Scale the params with `resolve_auto`
- Use the latest `linear_warmup_schedule` from LightlySSL after the latter gets a new release.
- Figure out how to feed params based on model config to `OptimArgs` and `TransformArgs`.

## How has it been tested?

N/A (test to be followed)

- [x] Yes

- [ ] Yes
- [x] Not needed (separate issue)


tests/_methods/dinov2/test_dinov2_transform.py
```diff
@@ -11,7 +11,7 @@
 
 import numpy as np
 
-from lightly_train._methods.dinov2.dinov2_transform import DINOv2Transform
+from lightly_train._methods.dinov2.dinov2_transform import DINOv2ViTSBTransform
```
JonasWurst: Testing LG as well

src/lightly_train/_methods/method.py
```diff
@@ -54,7 +54,7 @@ def __init__(
     ):
         super().__init__()
         self.global_batch_size = global_batch_size
-        self.optimizer_args: OptimizerArgs = optimizer_args
```
JonasWurst: why is this required?

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+    # crops
+    n_global_crops: int = 2
+    n_local_crops: int = (
+        8  # transform_cls().transform_args_cls().transform_args.local_view.num_views
+    )
+
+    # projection head
+    ibot_separate_head: bool = False
+    hidden_dim: int = 2048
+    bottleneck_dim: int = 256
```
JonasWurst, line 77: we need a separate `bootleneck_dim` for example in L, they are different (384,256)

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+    ) -> None:
+        # Determine the args based on the model architecture
+        depth: int = model.n_blocks
+        num_heads: int = model.num_heads
+        self.embed_dim: int = model._embed_dim
+        if depth == 40 and num_heads == 24 and self.embed_dim == 1536:  # giant
+            # projection head
+            self.ibot_separate_head = True
+            self.bottleneck_dim = 384
+            self.bottleneck_dim_ibot = 256
```
JonasWurst, line 150: here it is used, but it is missing in the top init

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+        self,
+        scaling_info: ScalingInfo,
+        optimizer_args: OptimizerArgs,
+        model: Module,
+    ) -> None:
+        # Determine the args based on the model architecture
+        depth: int = model.n_blocks
+        num_heads: int = model.num_heads
+        self.embed_dim: int = model._embed_dim
+        if depth == 40 and num_heads == 24 and self.embed_dim == 1536:  # giant
```
JonasWurst: We have conflicting concepts here. https://github.com/lightly-ai/lightly-train/blob/693c6b598b117c165ac1fc53b3835945c6fec82a/src/lightly_train/_methods/dino/dino.py#L71

We need to think about a solution that is also considering the dataset size. The `output_dim` is simply to large for small datasets

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+    layerwise_decay: float = 0.9
+    patch_embed_lr_multiplier: float = 0.2
+
+    # weight decay scheduler
+    weight_decay_end: float = 0.4
+
+    # gradient clipping
+    gradient_clip_val: float = 3.0
+
+    def resolve_auto(
```
JonasWurst, line 134: We need to take the learnings which we had for DINO into consideration here: https://github.com/lightly-ai/lightly-train/blob/693c6b598b117c165ac1fc53b3835945c6fec82a/src/lightly_train/_methods/dino/dino.py#L66

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+    center_momentum: float = 0.9
+
+    # teacher momentum
+    momentum_start: float = 0.992
+    momentum_end: float = 1.0
+
+    # teacher temp scheduler
+    start_teacher_temp: float = 0.04
+    end_teacher_temp: float = 0.07
+    warmup_teacher_temp_epochs: int = 30
```
JonasWurst: If you check https://github.com/lightly-ai/lightly-train/blob/693c6b598b117c165ac1fc53b3835945c6fec82a/src/lightly_train/_methods/dino/dino.py#L66, there they are all resolved auto --> we should discuss this

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+    koleo_loss_weight: float = 0.1
+
+    student_temp: float = 0.1
+
+    centering: Literal["softmax", "sinkhorn_knopp"] = "softmax"
+    center_momentum: float = 0.9
+
+    # teacher momentum
+    momentum_start: float = 0.992
+    momentum_end: float = 1.0
```
JonasWurst: also resolve auto candidates: https://github.com/lightly-ai/lightly-train/blob/693c6b598b117c165ac1fc53b3835945c6fec82a/src/lightly_train/_methods/dino/dino.py#L66

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+            self.weight_decay_end = 0.2
+
+        elif depth == 12 and num_heads == 12 and self.embed_dim == 768:  # base
+            pass
+        elif depth == 12 and num_heads == 6 and self.embed_dim == 384:  # small
+            pass
+        else:
+            raise ValueError("Unsupported model configs.")
+
+        # TODO: scale some params based on the scaling info
```
JonasWurst: Is this planned?

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+            # loss
+            self.centering = "sinkhorn_knopp"
+            # teacher momentum
+            self.momentum_start = 0.994
+            # lr scheduler
+            self.warmup_epochs = 80
+            # lr decay
+            self.layerwise_decay = 1.0
+            # weight decay scheduler
+            self.weight_decay_end = 0.2
```
JonasWurst: Completely identical to the upper --> please merge

the cfgs only differ in batchsize and model arch.

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+
+        elif depth == 12 and num_heads == 12 and self.embed_dim == 768:  # base
+            pass
+        elif depth == 12 and num_heads == 6 and self.embed_dim == 384:  # small
+            pass
+        else:
+            raise ValueError("Unsupported model configs.")
```
JonasWurst: instead warn and still pass. --> more flexible

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+        global_batch_size: int,
+    ):
+        super().__init__(
+            method_args=method_args,
+            optimizer_args=optimizer_args,
+            embedding_model=embedding_model,
+            global_batch_size=global_batch_size,
+        )
+
+        # Load configs based on the model architecture
```
JonasWurst: This comment is not matching the code

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+        self.method_args = method_args
+        self.optimizer_args = optimizer_args
+
+        # Calculate the number of crops
+        self.n_global_crops = self.method_args.n_global_crops
```
JonasWurst: Note: Will it be checked if == 2 ?

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+        self.n_global_crops_loss_terms = (self.n_global_crops - 1) * self.n_global_crops
+        self.n_local_crops_loss_terms = max(self.n_local_crops * self.n_global_crops, 1)
+
+        # Create teacher and student embedding models
+        model_wrapper: DINOv2ViTModelWrapper = embedding_model.model_wrapper
+        self.teacher_embedding_model_wrapper = model_wrapper
+        self.student_embedding_model_wrapper = copy.deepcopy(
+            self.teacher_embedding_model_wrapper
+        )
+
```
JonasWurst, line 279: at default, the model is created as student. Hence you need to convert after deepcopy:

```suggestion
        self.teacher_embedding_model_wrapper.make_teacher()

```

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+        # Load configs based on the model architecture
+        self.method_args = method_args
+        self.optimizer_args = optimizer_args
+
+        # Calculate the number of crops
+        self.n_global_crops = self.method_args.n_global_crops
+        self.n_local_crops = self.method_args.n_local_crops
+        self.n_global_crops_loss_terms = (self.n_global_crops - 1) * self.n_global_crops
+        self.n_local_crops_loss_terms = max(self.n_local_crops * self.n_global_crops, 1)
```
JonasWurst: Those numbers could also be determined on the fly in the train step as in the original code -> no need to include them in the args

JonasWurst: Note also in the dino implementation we used hard coded 2 globals + even the dinov2 asserts on global 2--> i would remove all this and hard code to 2, no need to have it in the args

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+
+        # Create teacher and student embedding models
+        model_wrapper: DINOv2ViTModelWrapper = embedding_model.model_wrapper
+        self.teacher_embedding_model_wrapper = model_wrapper
+        self.student_embedding_model_wrapper = copy.deepcopy(
+            self.teacher_embedding_model_wrapper
+        )
+
+        # Create teacher and student dino heads
+        self.teacher_dino_head = DINOProjectionHead(
```
JonasWurst: Consider using partial(...)

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+            output_dim=self.method_args.output_dim,
+            batch_norm=self.method_args.batch_norm,
+            freeze_last_layer=self.method_args.student_freeze_last_layer_epochs,
+            norm_last_layer=self.method_args.norm_last_layer,
+        )
+
+        # Create teacher and student iBOT head
+        self.ibot_separate_head: bool = self.method_args.ibot_separate_head
+        if self.ibot_separate_head:
+            self.teacher_ibot_head = DINOProjectionHead(
```
JonasWurst: consider using partial(...)

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+        # watch out: these are chunked and cat'd in reverse so A is matched to B in the global crops dino loss
+        cls_tokens = torch.cat(
+            (cls_tokens_chunked[1], cls_tokens_chunked[0])
+        )  # [G*B, C]
+
+        # process the patch tokens
+        patches = patches.flatten(2).permute(0, 2, 1)  # [G*B, H*W, C]
+
+        # forward through the teacher dino and ibot heads
+        if not self.ibot_separate_head:
```
JonasWurst: why differ here? we have the heads anyway

src/lightly_train/_methods/dinov2/dinov2.py
JonasWurst: Note: what todo if we loaded from pretrained?

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+        ibot_loss = training_step_log.ibot_loss
+        koleo_loss = training_step_log.koleo_loss
+
+        log_dict = {
+            "dino_global_loss": dino_global_loss,
+            "ibot_loss": ibot_loss,
+            "koleo_loss": koleo_loss,
+        }
+
+        if dino_local_loss is not None:
```
JonasWurst: can't we either calculate or put to 0.0. We will avoid quite some logic.

JonasWurst: then the training_step can be simply used from the base class

src/lightly_train/_methods/dinov2/dinov2.py
JonasWurst: We might also discard the gradients in the teacher

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+            max_steps=self.trainer.estimated_stepping_batches,
+            start_value=no_auto(self.method_args.momentum_start),
+            end_value=self.method_args.momentum_end,
+        )
+        update_momentum(
+            self.student_embedding_model_wrapper._model,
+            self.teacher_embedding_model_wrapper._model,
+            m=momentum,
+        )
+        update_momentum(self.student_dino_head, self.teacher_dino_head, m=momentum)
```
JonasWurst: Is the ibot head not updated ?

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+        self, batch: Batch, batch_idx: int
+    ) -> DINOv2TrainingStepResult:
+        # Momentum update teacher.
+        momentum = cosine_schedule(
+            step=self.trainer.global_step,
+            max_steps=self.trainer.estimated_stepping_batches,
+            start_value=no_auto(self.method_args.momentum_start),
+            end_value=self.method_args.momentum_end,
+        )
+        update_momentum(
```
JonasWurst: In the original code, the teacher is updated after the first iteration. I guess we can also keep it like this. but it is different.

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+                / self.trainer.max_epochs
+                * self.trainer.estimated_stepping_batches
+            ),
+            start_value=self.method_args.start_teacher_temp,
+            end_value=self.method_args.end_teacher_temp,
+        )
+
+        views = batch["views"]
+        global_views = torch.cat(
+            views[: self.n_global_crops]
```
JonasWurst: i would fix it as 2 in the top of the method --> lot of logic/deps can be removed

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+
+        # Masking
+        self.n_crops = global_views.shape[0]  # G*B
+        self.n_channels = global_views.shape[1]  # C
+        h = global_views.shape[2]
+        w = global_views.shape[3]
+
+        mask_generator = MaskingGenerator(
+            input_size=(h, w),
+            max_num_patches=int(0.5 * h * w),
```
JonasWurst: hard coded value. This can be moved to the config. --> does not collide with any other assumptions

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+        mask_ratio_max = self.method_args.mask_ratio_max
+        masks = create_collated_masks(
+            mask_ratio_min=mask_ratio_min,
+            mask_ratio_max=mask_ratio_max,
+            n_masked_crops=n_masked_crops,
+            n_crops=self.n_crops,
+            mask_generator=mask_generator,
+        )
+
+        self.collated_masks = masks["collated_masks"].to(
```
JonasWurst:
```suggestion
        collated_masks = masks["collated_masks"].to(
```

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+            device=self.device, non_blocking=True
+        )
+        self.mask_indices_list = masks["mask_indices_list"].to(
+            device=self.device, non_blocking=True
+        )
+        self.masks_weight = masks["masks_weight"].to(
```
JonasWurst: var

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+        h = global_views.shape[2]
+        w = global_views.shape[3]
+
+        mask_generator = MaskingGenerator(
+            input_size=(h, w),
+            max_num_patches=int(0.5 * h * w),
+        )
+        n_masked_crops = int(self.n_crops * self.method_args.mask_probability)
+        mask_ratio_min = self.method_args.mask_ratio_min
+        mask_ratio_max = self.method_args.mask_ratio_max
```
JonasWurst: why not simply use the attributes instead of this remapping?

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+            device=self.device, non_blocking=True
+        )
+        self.mask_indices_list = masks["mask_indices_list"].to(
+            device=self.device, non_blocking=True
+        )
+        self.masks_weight = masks["masks_weight"].to(
+            device=self.device, non_blocking=True
+        )
+        self.n_masked_patches = self.mask_indices_list.shape[0]
+        self.n_masked_patches_max = int(
```
JonasWurst: I would prefer those things to be given as input params. Makes it more clear what is actually processed by what.

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+            x
+        )  # input [G*B, C, ...]
+        patches = tokens["features"]  # [G*B, C, H, W]
+        cls_tokens = tokens["cls_token"]  # [G*B, C]
+
+        # process the cls tokens
+        cls_tokens_chunked = cls_tokens.chunk(self.n_global_crops)  # [G, B, C]
+        # watch out: these are chunked and cat'd in reverse so A is matched to B in the global crops dino loss
+        cls_tokens = torch.cat(
+            (cls_tokens_chunked[1], cls_tokens_chunked[0])
```
JonasWurst: Here 2 global crops are implicitly assumed!

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+        )  # [G*B, C]
+
```
JonasWurst:
```suggestion
        # watch out: these are chunked and cat'd in reverse so A is matched to B in the global crops dino loss
        cls_tokens = torch.cat((cls_tokens[B:], cls_tokens[:B]))  # [G*B, C]

```

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+            )  # [M, C]
+
+            tokens_after_head = self.teacher_dino_head.forward(
+                buffer_tokens
+            )  # [G*B+M_max, D]
+
+            cls_tokens_after_dino = tokens_after_head[: self.n_crops]  # [G*B, D]
+            masked_patch_tokens_after_ibot = tokens_after_head[
+                self.n_crops : self.n_crops + self.n_masked_patches
+            ]  # [M, D]
```
JonasWurst: Did you try using simply the ibot head instead? is it much slower --> ho much? code would much simpler

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+        )
+        self.mask_indices_list = masks["mask_indices_list"].to(
+            device=self.device, non_blocking=True
+        )
+        self.masks_weight = masks["masks_weight"].to(
+            device=self.device, non_blocking=True
+        )
+        self.n_masked_patches = self.mask_indices_list.shape[0]
+        self.n_masked_patches_max = int(
+            (n_masked_crops + 1) * h * w * (mask_ratio_min + mask_ratio_max) / 2.0
```
JonasWurst: should it be: `(n_masked_crops + 1) * h * w * mask_ratio_max`? that would be the actual max

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+            (cls_tokens_chunked[1], cls_tokens_chunked[0])
+        )  # [G*B, C]
+
+        # process the patch tokens
+        patches = patches.flatten(2).permute(0, 2, 1)  # [G*B, H*W, C]
+
+        # forward through the teacher dino and ibot heads
+        if not self.ibot_separate_head:
+            buffer_tokens = patches.new_zeros(
+                self.n_crops + self.n_masked_patches_max, self.n_channels
```
JonasWurst: why using this instead of the `self.n_masked_patches` as in the orig impl?

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+        if self.centering == "softmax":
+            cls_tokens_centered = self.dino_loss.softmax_center_teacher(
+                cls_tokens_after_dino, teacher_temp=self.teacher_temp
+            ).view(
+                self.n_global_crops, -1, *cls_tokens_after_dino.shape[1:]
+            )  # [G, B, D]
+            self.dino_loss.update_center(cls_tokens_after_dino)
+
+            masked_patch_tokens_centered = self.ibot_loss.softmax_center_teacher(
+                masked_patch_tokens_after_ibot[: self.n_masked_patches],
```
JonasWurst: I saw the orig impl has this as well, but why is `[: self.n_masked_patches]` required. it should be of that shape anyway

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+            )
+        elif self.centering == "sinkhorn_knopp":
+            cls_tokens_centered = self.dino_loss.sinkhorn_knopp_teacher(
+                cls_tokens_after_dino, teacher_temp=self.teacher_temp
+            ).view(
+                self.n_global_crops, -1, *cls_tokens_after_dino.shape[1:]
+            )  # [G, B, D]
+
+            masked_patch_tokens_centered = self.ibot_loss.sinkhorn_knopp_teacher(
+                masked_patch_tokens_after_ibot,
```
JonasWurst, line 557: Here, it is not btw.

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+            x
+        )  # input [G*B, C, ...]
+        patches = tokens["features"]  # [G*B, C, H, W]
+        cls_tokens = tokens["cls_token"]  # [G*B, C]
+
+        # process the patch tokens
+        patches = patches.flatten(2).permute(0, 2, 1)  # [G*B, H*W, C]
+
+        # forward through the student dino and ibot heads
+        if not self.ibot_separate_head:
```
JonasWurst: Similar to the teacher comment

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+
+        return cls_tokens_after_dino, masked_patch_tokens_after_ibot
+
+    def _forward_student_local(self, x: Tensor) -> Tensor:
+        tokens = self.student_embedding_model_wrapper.forward_features(
+            x
+        )  # input [L*B, C, ...]
+        cls_tokens = tokens["cls_token"]  # [L*B, C]
+
+        # forward through the teacher dino and ibot heads
```
JonasWurst: comment needs to be updated

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+        if self.n_local_crops > 0:
+            dino_local_loss = (
+                self.dino_loss.forward(
+                    student_output_list=student_cls_tokens_local.chunk(
+                        self.n_local_crops
+                    ),  # [L, B, D]
+                    teacher_out_softmaxed_centered_list=teacher_cls_tokens_centered,  # [G, B, D]
+                )
+                / (self.n_global_crops_loss_terms + self.n_local_crops_loss_terms)
+            )
```
JonasWurst: group with upper if

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+            * self.n_global_crops
+            / (self.n_global_crops_loss_terms + self.n_local_crops_loss_terms)
+        )
+        if self.n_local_crops > 0:
+            dino_local_loss = (
+                self.dino_loss.forward(
+                    student_output_list=student_cls_tokens_local.chunk(
+                        self.n_local_crops
+                    ),  # [L, B, D]
+                    teacher_out_softmaxed_centered_list=teacher_cls_tokens_centered,  # [G, B, D]
```
JonasWurst, line 457: This input should be a list, but it seems it is not

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+    @staticmethod
+    def method_args_cls() -> type[DINOv2Args]:
+        return DINOv2Args
+
+    @staticmethod
+    def optimizer_args_cls(
+        optim_type: OptimizerType | Literal["auto"],
+    ) -> type[DINOv2AdamWViTSBArgs | DINOv2AdamWViTLGArgs]:
+        classes: dict[OptimizerType | Literal["auto"], type[OptimizerArgs]] = {
+            OptimizerType.ADAMW: DINOv2AdamWViTSBArgs,
```
JonasWurst: "auto" is missing

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+
+    def configure_gradient_clipping(
+        self,
+        optimizer: Optimizer,
+        gradient_clip_val: int | float | None = None,
+        gradient_clip_algorithm: str | None = None,
+    ) -> None:
+        self.clip_gradients(
+            optimizer=optimizer,
+            gradient_clip_val=gradient_clip_val,
```
JonasWurst: `self.method_args.` ?

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,678 @@
...
+                self.student_embedding_model_wrapper._model,
+                self.student_dino_head,
+                self.student_ibot_head,
+            ],  # decay is realized in get_optimizer_with_decay
+        )
+
+    # Ignore the return type, because pytorch-lightning types it wrongly.
+    # See https://github.com/Lightning-AI/pytorch-lightning/issues/20106
+    def configure_optimizers(self) -> OptimizerLRScheduler:
+        optim = get_optimizer_with_decay(
```
JonasWurst, line 636: can we rearrange this?

pull the optim code in and call the optimizer from here as it is done in the other methods? I know this file will get huge but at least we have the same style everywhere?

src/lightly_train/_methods/dinov2/dinov2_optim.py
```diff
@@ -0,0 +1,138 @@
...
+        trainable_modules (TrainableModules): trainable modules.
+        lr_scale (float): learning rate scale.
+        layerwise_decay (float): base lr decay rate.
+        patch_embed_lr_multiplier (float): multiplier for patch embedding layer.
+    Returns:
+        Optimizer: optimizer with decay.
+    """
+
+    all_param_groups: List[Dict[str, Any]] = []
+    for module in trainable_modules.modules:  # TODO: FSDP sharding
```
JonasWurst: will you create a follow up for this?

src/lightly_train/_methods/dinov2/dinov2_optim.py
JonasWurst, line 1: Mostly copied

src/lightly_train/_methods/dinov2/utils.py
JonasWurst, line 1: mostly copied

JonasWurst (changes requested):
This looks already pretty good. I have some general points besides the comments:

- lic headers need to be updated
- do we want to have separate files for utils and optim
- we should see if we actually need the model for the transform args, the differences are very little (local crop size 96 vs 98 --> maybe not worth the effort atm?)
- for the optimzer args, the solution is not that straight forward--> i think for now we need to stick to one fixed. Create a follow up for the issues, if we change the interface? @guarin

src/lightly_train/_methods/method.py
yutong-xiang-97: Ah this is an artifact created for testing stuff. Will be reverted.

src/lightly_train/_methods/dinov2/dinov2.py
yutong-xiang-97: It's simply a placeholder for `resolve_auto`

src/lightly_train/_methods/dinov2/dinov2.py
yutong-xiang-97: That would be an overestimation if you look into the masking creation logic.

src/lightly_train/_methods/dinov2/dinov2.py
yutong-xiang-97: The argument is named as such. In the global loss it is a list, in the local loss it is a Tensor, which is also reflected in the typing. It followed the original code w/o changing the types.

src/lightly_train/_methods/dinov2/dinov2.py
yutong-xiang-97: The original impl. indeed uses `self.n_masked_patches_max` (called upperbound)

src/lightly_train/_methods/dinov2/dinov2.py
yutong-xiang-97: removed

src/lightly_train/_methods/dinov2/dinov2_optim.py
yutong-xiang-97: This is from @stegmuel and I am not aware what changes should be made. See the conversation here: https://lightly-ai.slack.com/archives/C07J7HHCJHK/p1747404957696759?thread_ts=1747308653.453729&cid=C07J7HHCJHK

src/lightly_train/_methods/dinov2/dinov2.py
yutong-xiang-97: What do you mean by "discard the gradients in the teacher"? for now `_forward_teacher` runs under `@torch.no_grad()`

src/lightly_train/_methods/dinov2/dinov2.py
yutong-xiang-97: Sorry, missed it. Will add.

src/lightly_train/_methods/dinov2/dinov2.py
yutong-xiang-97: This is easy to add so no worries.

src/lightly_train/_methods/dinov2/dinov2.py
yutong-xiang-97: Could you elaborate on which to move? The `0.5` is hard-coded in the original code though. https://github.com/facebookresearch/dinov2/blob/4d83ffd4f4b39df112cef7ed7833de7ebf3a202e/dinov2/train/train.py#L172

src/lightly_train/_methods/dinov2/dinov2.py
yutong-xiang-97: Could you elaborate?

src/lightly_train/_methods/dinov2/dinov2.py
yutong-xiang-97: Would result in the same model if `model_ema = model` but I would put it to the end of training_step_impl to avoid useless calculation

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,691 @@
...
+                / self.trainer.max_epochs
+                * self.trainer.estimated_stepping_batches
+            ),
+            start_value=no_auto(self.method_args.start_teacher_temp),
+            end_value=no_auto(self.method_args.end_teacher_temp),
+        )
+
+        # Get the views
+        # Calculate the number of crops
+        n_local_crops = self.method_args.n_local_crops
```
JonasWurst, line 379: i would still have set it here and reuse
`n_global_crops = 2`

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,691 @@
...
+
+        koleo_loss = sum(
+            self.koleo_loss(token) for token in student_cls_tokens_global.chunk(2)
+        )  # [G, B, D], only use global views
+
+        loss = (
+            self.method_args.dino_loss_weight * dino_global_loss
+            + self.method_args.dino_loss_weight * dino_local_loss
+            if dino_local_loss is not None
+            else 0.0
```
JonasWurst: I would put `dino_local_loss = torch.tensor(0.0)` infront of `if n_local_crops > 0:` in line 439, and remove this if, and also the one in the return

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -0,0 +1,691 @@
...
+            m=momentum,
+        )
+        update_momentum(self.student_dino_head, self.teacher_dino_head, m=momentum)
+        if self.ibot_separate_head:
+            update_momentum(self.student_ibot_head, self.teacher_ibot_head, m=momentum)
+
+        return DINOv2TrainingStepResult(
+            loss=loss,
+            dino_global_loss=dino_global_loss,
+            dino_local_loss=dino_local_loss if n_local_crops > 0 else torch.tensor(0.0),
```
JonasWurst: Remove as explained above

PR #83, opened 2025-05-20
Add LIGHTLY_TRAIN_TMP_DIR variable

guarin:
## What has changed and why?

* Allow overwriting the temporary directory used by LightlyTrain

The directory can be overwritten with the `LIGHTLY_TRAIN_TMP_DIR` variable.
Added this as it is also helpful for testing. 

## How has it been tested?

* Unit tests

- [ ] Yes

- [ ] Yes


stegmuel (approved):
LGTM. What is the convention regarding the double "_"?

guarin:
> What is the convention regarding the double "_"?

I assume the question is with respect to `test_get_tmp_dir__custom`. The double underscore is the separator between the name of the function that is tested and any extra information. We use the following format: `test_<function name>__<other info>`.

PR #89, opened 2025-05-21
Relax Model Constraints

liopeer:
## What has changed and why?
- allow `Any` types for models
- make compatibility checks more sophisticated
- make export more sophisticated

## How has it been tested?
- adapted existing unit tests
- created new unit tests to cover the relaxed typing

- [ ] Yes

Implicitly, we always supported directly passing models, however, we don't have it anywhere in our documentation yet. I would make a Changelog entry once we adapt the documentation.

- [ ] Yes


Makefile
```diff
@@ -133,7 +133,7 @@ endif
 
 # RFDETR is not compatible with Python<3.9. Therefore we exclude it from the
 # default extras.
-EXTRAS_PY38 := [dev,notebook,supergradients,tensorboard,timm,ultralytics,wandb]
+EXTRAS_PY38 := [dev,notebook,super-gradients,tensorboard,timm,ultralytics,wandb]
```
liopeer: This was actually an error I detected during testing. Very likely that we never actually tested the `super-gradients` stuff in CI.

src/lightly_train/_models/custom/custom_package.py
```diff
@@ -24,12 +25,22 @@ class CustomPackage(BasePackage):
...
     @classmethod
-    def is_supported_model(cls, model: Module | ModelWrapper) -> bool:
-        """Check if the model is supported by this package."""
+    def is_supported_model(cls, model: Module | ModelWrapper | Any) -> bool:
         return isinstance(model, ModelWrapper)
 
     @classmethod
-    def export_model(cls, model: Module, out: Path, log_example: bool = True) -> None:
+    def export_model(
+        cls, model: Module | ModelWrapper | Any, out: Path, log_example: bool = True
```
liopeer, line 33: Typing here could of course also just be `Any`, but this makes it more understandable what is requested.

Makefile
guarin: Does uv not raise an error if you request a dependency group that does not exist 🫨 ?

src/lightly_train/_models/rfdetr/rfdetr_package.py
```diff
@@ -12,7 +12,14 @@
 from typing import Any
 
 import torch
-from torch.nn import Module
+
+try:
+    from rfdetr.detr import RFDETR
+except ImportError:
+
+    class RFDETR:  # type: ignore[no-redef]
```
liopeer, line 20: Needed because we use `RFDETR` for instance checks. Type hints would work even if the import failed, but instance checks don't.

Makefile
liopeer: No, it just gives a warning 😅

tests/_models/torchvision/__init__.py
```diff
@@ -0,0 +1,7 @@
+#
```
liopeer, line 1: Required to import the `helpers` inside the `test_torchvision_package`.

src/lightly_train/_models/rfdetr/rfdetr_package.py
```diff
@@ -72,40 +73,62 @@ def get_model(
...
         except ImportError:
             raise ValueError(
                 f"Cannot create model because '{cls.name}' is not installed."
             )
-        if not isinstance(model, LWDETR):
-            raise ValueError(f"Model must be of type 'LWDETR', got {type(model)}")
 
-        torch.save({"model": model.state_dict()}, out)
+        if isinstance(model, RFDETRModelWrapper):
+            rfdetr_model = model.get_model()
```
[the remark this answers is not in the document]
liopeer: Fixed, but differently.

src/lightly_train/_methods/distillation/distillation.py
```diff
@@ -40,7 +40,7 @@
 logger = logging.getLogger(__name__)
 
 
-def get_teacher(teacher_name: str) -> Module:
+def get_teacher(teacher_name: str) -> Any:
```
guarin: This must still be a `Module`, otherwise you can't call `teacher_embedding_model.eval()`. In these cases it is IMO ok to add an assert as we know which types of teacher we expect.

src/lightly_train/_models/custom/custom_package.py
```diff
@@ -24,12 +25,23 @@ class CustomPackage(BasePackage):
...
+    ) -> None:
+        if isinstance(model, ModelWrapper):
+            model = model.get_model()
+        elif isinstance(model, Module):
+            model = model
+        else:
+            raise ValueError(
+                f"CustomPackage only supports exportint ModelWrapper or torch.nn.Module, "
+                f"but got {type(model)}"
```
guarin:
```suggestion
                f"CustomPackage only supports exporting ModelWrapper or torch.nn.Module, "
                f"but got {type(model)}"
```

src/lightly_train/_models/package.py
```diff
@@ -11,24 +11,36 @@
...
 from typing import Any
 
-from torch.nn import Module
-
 from lightly_train._models.model_wrapper import ModelWrapper
 
+# PackageModels should be either a ModelWrapper or the package's underlying models, such
+# as a ResNet (torchvision.models.resnet.ResNet) or an RFDETR (rfdetr.detr.RFDETR).
+# Most of these will be torch.nn.Module, but e.g. RFDETR is a custom class.
+PackageModel = Any
```
guarin: This is a bit confusing to me. I think we should differentiate between model and wrapped model. `get_model` returns a `PackageModel` whereas `get_wrapped_model` returns a `ModelWrapper`.

I would also define `PackageModel` only once and then import it everywhere. You can put it into `types.py` if you need it in many places. The alternative is to keep `Any` everywhere. Now it is a bit weird because there is a mix between `Any` and `PackageModel`. Maybe we can also use `Any` for the external facing code and `PackageModel` in the internal code?

src/lightly_train/_models/package.py
```diff
@@ -11,24 +11,36 @@
...
 class BasePackage(ABC):
     name: str  # The name of the package.
 
     @classmethod
     @abstractmethod
     def export_model(
-        cls, model: Module, out: Path, log_example: bool = True
-    ) -> None: ...
+        cls,
+        model: PackageModel | ModelWrapper | Any,
```
guarin, line 25:
```suggestion
        model: PackageModel | ModelWrapper,
```

src/lightly_train/_models/package.py
```diff
@@ -11,24 +11,36 @@
...
+        """Export the model in the package's format. The model can be either a
+        ModelWrapper or one of the package's underlying models.
+        """
+        ...
 
     @classmethod
     @abstractmethod
-    def is_supported_model(cls, model: Module | ModelWrapper) -> bool:
-        """Check if the model is supported by this package."""
+    def is_supported_model(cls, model: PackageModel | ModelWrapper | Any) -> bool:
```
guarin, line 36:
```suggestion
    def is_supported_model(cls, model: PackageModel | ModelWrapper) -> bool:
```

src/lightly_train/_models/rfdetr/rfdetr_package.py
```diff
@@ -72,40 +73,62 @@ def get_model(
...
-        model_full = model_rfdetr.model.model  # The actual LWDETR model, which is a submodule of nn.Module, is stored in RFDETR().model.model
-        if isinstance(model_full, LWDETR):
-            return model_full  # type: ignore
-        else:
-            raise ValueError(f"Model must be of type 'LWDETR', got {type(model_full)}")
+        return model_rfdetr
 
     @classmethod
-    def get_model_wrapper(cls, model: Module) -> ModelWrapper:
+    def get_model_wrapper(cls, model: RFDETR) -> ModelWrapper:
```
guarin: Might want to update the others as well:

```suggestion
    def get_model_wrapper(cls, model: RFDETR) -> RDETRModelWrapper:
```

src/lightly_train/_models/rfdetr/rfdetr_package.py
```diff
@@ -72,40 +73,62 @@ def get_model(
...
+
+        if not cls.is_supported_model(model):
+            raise ValueError(
+                f"Model must be of type 'RFDETR' or 'RFDETRModelWrapper', got {type(model)}"
+            )
+
+        lwdetr_model = model.model.model
+        assert isinstance(lwdetr_model, LWDETR)
+
+        # Set encoder state to eval mode
```
guarin: Is the conversion to eval mode missing?

src/lightly_train/_models/ultralytics/ultralytics_package.py
```diff
@@ -15,6 +15,11 @@
 import torch
 from torch.nn import Module
 
+try:
+    from ultralytics import YOLO
+except ImportError:
+    pass
```
guarin: If the class is only used for type checking you can import it with:
```
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from ultralytics import YOLO
```

tests/_models/rfdetr/test_rfdetr.py
```diff
@@ -71,8 +73,7 @@ def test_forward_pool(self) -> None:
         assert pool.shape == (1, expected_dim, 1, 1)
 
     def test_get_model(self) -> None:
-        model_instance = RFDETRBase()  # type: ignore[no-untyped-call]
-        model = model_instance.model.model
+        model = RFDETRBase()  # type: ignore[no-untyped-call]
         feature_extractor = RFDETRModelWrapper(model=model)
```
guarin: Rename `feature_extractor` to `wrapped_model`

src/lightly_train/_models/package.py
liopeer: I now put `PackageModel = Any` into `types.py`. I would also use `PackageModel` for inwards-facing code, since it's more expressive & we could decide to later constrain it again to e.g. just `PackageModel = Module | RFDETR`.

src/lightly_train/_models/package.py
liopeer: I would actually keep the `Any` here. IMO it makes it clear that while implementing the method one has to check for 3 cases, the `PackageModel` (usually a `Module`), the `ModelWrapper` and any other input. E.g. in the `CustomPackage`, the typing should then look like:
```python
    @classmethod
    def export_model(
        cls, model: Module | ModelWrapper | Any, out: Path, log_example: bool = True
    ) -> None:
        if isinstance(model, ModelWrapper):
            model = model.get_model()
        elif isinstance(model, Module):
            model = model
        else:
            raise ValueError(
                f"CustomPackage only supports exporting ModelWrapper or torch.nn.Module, "
                f"but got {type(model)}"
            )
```
To me, this is the clearest possible option. If you disagree I will change it.

src/lightly_train/_models/package.py
liopeer: Same here as above.

src/lightly_train/_models/rfdetr/rfdetr_package.py
liopeer: Did it in all of them, thx!

src/lightly_train/_models/rfdetr/rfdetr_package.py
liopeer: That's a comment from an experiment I ran. I removed it now.

src/lightly_train/_models/ultralytics/ultralytics_package.py
liopeer: fixed.

tests/_models/rfdetr/test_rfdetr.py
liopeer: Changed all of them in that module. There are some remaining ones in the ultralytics package and torchvision packages. will do those in a follow-up.

src/lightly_train/_models/ultralytics/ultralytics_package.py
liopeer: fixed.

src/lightly_train/types.py
```diff
@@ -16,6 +16,8 @@
 from torch import Tensor
 from typing_extensions import NotRequired
 
+PackageModel = Any
```
guarin, line 22: Nit: Could you add a comment explaining this type and why it is Any?

src/lightly_train/_methods/distillation/distillation.py
```diff
@@ -40,10 +40,11 @@
 logger = logging.getLogger(__name__)
 
 
-def get_teacher(teacher_name: str) -> Any:
+def get_teacher(teacher_name: str) -> Module:
     wrapped_model = package_helpers.get_wrapped_model(model=teacher_name)
     teacher_embedding_model = wrapped_model.get_model()
     teacher_embedding_model.eval()
+    assert isinstance(teacher_embedding_model, Module)
```
guarin, line 46: Nit: If it is not a module the code will most likely already fail in the line above the assert when it tries to call `.eval()`.

src/lightly_train/types.py
liopeer: Done

src/lightly_train/_methods/distillation/distillation.py
liopeer: Done.

PR #91, opened 2025-05-21
Use smaller ViT for unit tests

guarin:
## What has changed and why?

* Use ViT small instead of base for unit tests

Small optimization to speed up unit tests and use less RAM

## How has it been tested?

* CI

- [ ] Yes

- [ ] Yes


stegmuel (approved):
Approved.

PR #94, opened 2025-05-22
mlflow logger

JonasWurst:
## What has changed and why?

This PR brings a mlflow logger for lightly-train. It will log model metrics and images as the other loggers do. Only the log_model is not yet supported. Follow up required!

## How has it been tested?

unit tests in the style of the other loggers are in place

- [x] Yes

- [x] Yes


src/lightly_train/_loggers/mlflow.py
```diff
@@ -0,0 +1,73 @@
...
+from pytorch_lightning.utilities import rank_zero_only
+from torch import Tensor
+
+from lightly_train._configs.config import PydanticConfig
+
+
+class MLFlowLoggerArgs(PydanticConfig):
+    experiment_name: str = ""
+    run_name: str | None = None
+    tracking_uri: str | None = os.getenv("MLFLOW_TRACKING_URI")
```
liopeer, line 24: Do we not want to expose `synchronous`, `save_dir` and `checkpoint_path_prefix` to the user?

src/lightly_train/_loggers/mlflow.py
```diff
@@ -0,0 +1,73 @@
...
+            prefix=prefix,
+            artifact_location=artifact_location,
+            run_id=run_id,
+            synchronous=synchronous,
+        )
+        self.save_temp_dir = str(save_dir)
+
+    @rank_zero_only  # type: ignore[misc]
+    def log_image(
+        self, key: str, images: list[Tensor], step: int | None = None
```
liopeer: Does this really support tensors? In `_methods/method.py` you apparently pass PIL images. And also the docs suggest otherwise: https://mlflow.org/docs/latest/api_reference/python_api/mlflow.html#mlflow.log_image

docs/source/train/index.md
```diff
@@ -182,6 +184,42 @@ loggers={"jsonl": None}
 loggers.jsonl=null
 ````
 
+(mlflow)=
+
+### MLflow
+
+```{important}
+MLflow must be installed with `pip install "lightly-train[mlflow]"`.
```
liopeer, line 192: We don't have this optional dependency atm.

src/lightly_train/_callbacks/mlflow_logging.py
```diff
@@ -0,0 +1,47 @@
+#
+# Copyright (c) Lightly AG and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+#
+
+from mlflow.system_metrics.system_metrics_monitor import SystemMetricsMonitor
```
liopeer: I don't see this class anywhere in mlflow's docs. Why can't we enable it like mentioned here: https://mlflow.org/docs/latest/system-metrics/

src/lightly_train/_loggers/mlflow.py
liopeer: Found out that `save_dir` is coming from us.

src/lightly_train/_methods/method.py
```diff
@@ -161,7 +162,9 @@ def _log_example_views(self, train_batch: Batch) -> None:
                     img_tensor=example_aug_tensor,
                     global_step=self.global_step,
                 )
-            elif isinstance(logger, LightningWandbLogger):
```
liopeer, line 164: We could think about coming up with an interface for the loggers. Does not have to be right now, but it would probably simplify things.

src/lightly_train/_loggers/mlflow.py
JonasWurst: >  Do we not want to expose synchronous, save_dir and checkpoint_path_prefix to the user?

I removed completely, since it is not available in pytorch_lightning 2.1

-> resolved

src/lightly_train/_loggers/mlflow.py
JonasWurst: Fixed. (old version)

docs/source/train/index.md
JonasWurst: fixed

src/lightly_train/_callbacks/mlflow_logging.py
JonasWurst: I tried this way initially, however that did not work. I assume it has to do with the fact that the logger is based on the client, and is not starting a run via context manager.

src/lightly_train/_methods/method.py
JonasWurst: As discussed: I would propose to do so, as soon as we have a second case like this

Makefile
```diff
@@ -133,20 +133,20 @@ endif
 
 # RFDETR is not compatible with Python<3.9. Therefore we exclude it from the
 # default extras.
-EXTRAS_PY38 := [dev,notebook,supergradients,tensorboard,timm,ultralytics,wandb]
+EXTRAS_PY38 := [dev,mlflow,notebook,supergradients,tensorboard,timm,ultralytics,wandb]
```
liopeer, line 136: `mlflow` requires Python >=3.9. Could you remove it and add it to the comment above?

src/lightly_train/_callbacks/callback_helpers.py
```diff
@@ -8,27 +8,37 @@
...
 from lightly_train._configs import validate
+from lightly_train._loggers.jsonl import JSONLLogger
+from lightly_train._loggers.mlflow import MLFlowLogger
+from lightly_train._loggers.tensorboard import TensorBoardLogger
+from lightly_train._loggers.wandb import WandbLogger
 from lightly_train._models.embedding_model import EmbeddingModel
 from lightly_train._models.model_wrapper import ModelWrapper
 from lightly_train._transforms.transform import NormalizeArgs
 
+AnyLoggerType = TypeVar(
```
liopeer, line 38: Nit: I would use a Union-type, since you're not creating a generic function.

liopeer (approved):
LGTM overall, one thing to fix in the pinned dependencies for Python 3.8, the other one is nitpicking. Feel free to address, resolve and merge.

src/lightly_train/_callbacks/callback_helpers.py
JonasWurst: I went for that initially, but mypy was complaining. So i used this solution

Makefile
JonasWurst: mlflow 2.17 is still supporting python 3.8, since we have mlflow >= 2.0, having it in the pinned 3.8 dependencies is fine

PR #96, opened 2025-05-22
Update custom model docs

guarin:
## What has changed and why?

* Update custom model docs to include `get_model` info
* Update error message when a custom model is missing some methods

## How has it been tested?

* Added unit test
* Tested manually

- [x] Yes

- [x] Yes

### DIFF
--- a/docs/source/models/rtdetr.md
+++ b/docs/source/models/rtdetr.md
@@ -59,6 +59,9 @@ class RTDETRModelWrapper(Module):
         self._model = model
         self._pool = AdaptiveAvgPool2d((1, 1))
 
+    def get_model(self) -> Module:
+        return self._model
+
     def forward_features(self, x: Tensor) -> Dict[str, Tensor]:
         features = self._model.backbone(x)[-1]
         return {"features": features}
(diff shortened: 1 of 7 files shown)


docs/source/models/rtdetr.md
```diff
@@ -59,6 +59,9 @@ class RTDETRModelWrapper(Module):
         self._model = model
         self._pool = AdaptiveAvgPool2d((1, 1))
 
+    def get_model(self) -> Module:
```
stegmuel, line 62: What about:

```
class RTDETRModelWrapper(Module):
    def __init__(self, model: Module):
        super().__init__()
        self._model = [model]
        self.backbone = model.backbone
        self._pool = AdaptiveAvgPool2d((1, 1))

    def forward_features(self, x: Tensor) -> Dict[str, Tensor]:
        features = self._model[0].backbone(x)[-1]
        return {"features": features}

    def forward_pool(self, x: Dict[str, Tensor]) -> Dict[str, Tensor]:
        return {"pooled_features": self._pool(x["features"])}

    def feature_dim(self) -> int:
        return self._model[0].backbone.out_channels[-1]

    def get_model(self) -> Module:
        return self._model[0]

```

This way the number of parameters displayed by LightlyTrain corresponds to the number actually being trained.

stegmuel (approved):
Approved with a minor comment/suggestion.

PR #104, opened 2025-05-26
change dataset

liopeer:
## What has changed and why?
- adjusted datasets and paths

## How has it been tested?
- dataset-related cells were run

- [ ] Yes

- [ ] Yes

### DIFF
--- a/examples/notebooks/yolov12.ipynb
+++ b/examples/notebooks/yolov12.ipynb
@@ -13,7 +13,7 @@
    "id": "1",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


examples/notebooks/yolov12.ipynb
```diff
@@ -245,7 +249,7 @@
     "if __name__ == \"__main__\":\n",
     "    lightly_train.train(\n",
     "        out=\"out/my_experiment\",  # Output directory.\n",
-    "        data=f\"{settings['datasets_dir']}/coco8/images/train\",  # Directory with images.\n",
+    "        data=f\"{settings['datasets_dir']}/VOC/images/train\",  # Directory with images, no labels!\n",
```
yutong-xiang-97: The data dir is not correct. Should be e.g. `train2007`

yutong-xiang-97 (changes requested):
All good but I would suggest we still use separate sections for pretraining and fine-tuning and process the respective datasets under these two sections, just like in Ultralytics. It is one of the major points for updating the tutorial. WDYT?

examples/notebooks/yolov12.ipynb
liopeer: changed.

PR #107, opened 2025-05-26
Add RT-DETR ModelWrapper to `model_wrappers`

liopeer:
## What has changed and why?
- adds the custom wrapper for RT-DETR to a new module `lightly_train.model_wrappers`

## How has it been tested?
- was tested locally. unit tests make no sense since RT-DETR is not pip-installable

- [x] Yes

- [x] Yes

### DIFF
--- a/src/lightly_train/_models/rtdetr/rtdetr.py
+++ b/src/lightly_train/_models/rtdetr/rtdetr.py
@@ -1,0 +1,23 @@
+from __future__ import annotations
+
+from torch.nn import Module, AdaptiveAvgPool2d
+from torch import Tensor
+
+class RTDETRModelWrapper(Module):
+    def __init__(self, model: Module):
+        super().__init__()
+        self._model = model
+        self._pool = AdaptiveAvgPool2d((1, 1))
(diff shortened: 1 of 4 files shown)


src/lightly_train/_models/rtdetr/rtdetr.py
```diff
@@ -0,0 +1,39 @@
...
+from torch.nn import AdaptiveAvgPool2d, Module
+
+from lightly_train._models.model_wrapper import (
+    ForwardFeaturesOutput,
+    ForwardPoolOutput,
+    ModelWrapper,
+)
+
+
+class RTDETRModelWrapper(Module, ModelWrapper):
```
stegmuel, line 20: I'm in favor of using: `self._model = [model]` and `self.backbone = model.backbone` to have the number of trainable parameters correctly displayed. What do you think?

liopeer: Agree. Changed it!

stegmuel (approved):
Approved.

PR #112, opened 2025-05-28
Correct the log_model in case of pytorch_lightning

JonasWurst:
## What has changed and why?

Due to a bug in pytorch-lightning, logging models with MLFlowLogger is not possible with the current 2.5.1 version. This PR is a fix, which should be removed as soon as a fix bi pytorch-lightning is provided.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_loggers/mlflow.py
+++ b/src/lightly_train/_loggers/mlflow.py
@@ -12,7 +12,7 @@ import os
 from pathlib import Path
 from typing import Any, Literal
 
-import pytorch_lightning
+from lightning_utilities.core.imports import RequirementCache
 from PIL.Image import Image
 from pydantic import validator
 from pytorch_lightning.loggers import MLFlowLogger as LightningMLFlowLogger
@@ -22,6 +22,8 @@ from lightly_train._configs.config import PydanticConfig
 
 logger = logging.getLogger(__name__)
 
+PYTORCH_LIGHTNING_BUG_VERSION = "2.5.1"
+PL_BUG_VERSION_INSTALLED = RequirementCache(f"pytorch-lightning=={PYTORCH_LIGHTNING_BUG_VERSION}")
 
 class MLFlowLoggerArgs(PydanticConfig):
     experiment_name: str = ""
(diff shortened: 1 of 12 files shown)


src/lightly_train/_loggers/mlflow.py
```diff
@@ -28,6 +33,18 @@ class MLFlowLoggerArgs(PydanticConfig):
     artifact_location: str | None = None
     run_id: str | None = None
 
+    @validator("log_model") # type: ignore[untyped]
```
liopeer, line 38: Why do we need this validator and the type ignore? Is it because literals are not compatible with Python 3.8?

src/lightly_train/_loggers/mlflow.py
```diff
@@ -28,6 +33,18 @@ class MLFlowLoggerArgs(PydanticConfig):
...
+    def validate_log_model(cls, v: Literal[True, False, "all"]) -> Literal[True, False, "all"]:
+        if v not in [True, False, "all"]:
+            raise ValueError("log_model must be one of True, False or 'all'")
+        if v is not False:
```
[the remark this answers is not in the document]
liopeer: Yes, should be addressed!

src/lightly_train/_loggers/mlflow.py
```diff
@@ -28,6 +33,18 @@ class MLFlowLoggerArgs(PydanticConfig):
...
+            if pytorch_lightning.__version__ == "2.5.1": # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
liopeer: I would prefer the `RequirementCache`.

src/lightly_train/_loggers/mlflow.py
```diff
@@ -28,6 +33,18 @@ class MLFlowLoggerArgs(PydanticConfig):
...
+                logger.warning(
+                    "Due to a bug in pytorch_lightning 2.5.1, logging models with MLFlowLogger is not possible.")
```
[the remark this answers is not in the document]
liopeer: Agree. I would include the referencing issue and a possible mitigation, i.e. a line `pip install "pytorch_lignthning>=2.1,!=2.5.1"`. Alternatively, we could also just directly avoid 2.5.1 in the `pyproject.toml`'s `mflow` dependencies with the same formula.

```toml
mlflow = [
    "mlflow>=2.0.0", # contain fixes for numpy>=2.0 support
    "pytorch_lightning!=2.5.1",
]
```
I think that would actually be the best solution. But make sure to add a comment.

liopeer (changes requested):
I would do it via the dependencies as mentioned in one of the comments.

src/lightly_train/_loggers/mlflow.py
JonasWurst: It is the pydanitc way to check the args. In this case it is used to correct the args, such that the displayed config is actually valid.

src/lightly_train/_loggers/mlflow.py
```diff
@@ -42,7 +65,6 @@ def __init__(
         artifact_location: str | None = None,
         run_id: str | None = None,
     ) -> None:
-        os.environ["MLFLOW_ENABLE_SYSTEM_METRICS_LOGGING"] = "true"
```
liopeer, line 45: Why are we not using this anymore? Also could you move the environment variables to `_env.py`. Sorry that I did not see this during the initial reviews.

liopeer (changes requested):
If you could answer my last question & move the env variables to `_env.py`, then we could merge.

src/lightly_train/_loggers/mlflow.py
JonasWurst: This should not have been here. In the initial commit (mlflow logger) I did not remove it after unsuccessfully testing it.

liopeer (approved):
LGTM, thanks for the patience!

PR #113, opened 2025-05-30
Make Arguments Resolution more sophisticated (draft)

liopeer:
## What has changed and why?
- <none>

## How has it been tested?
- <noneY

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/common_helpers.py
+++ b/src/lightly_train/_commands/common_helpers.py
@@ -26,7 +26,6 @@ from pytorch_lightning.accelerators.cuda import CUDAAccelerator
 from pytorch_lightning.accelerators.mps import MPSAccelerator
 from pytorch_lightning.strategies.strategy import Strategy
 from torch.nn import Module
-from torch.utils.data import Dataset
 
 from lightly_train import _distributed as distributed_helpers
 from lightly_train._data import image_dataset
@@ -40,7 +39,7 @@ from lightly_train._models.custom.custom_package import CUSTOM_PACKAGE
 from lightly_train._models.embedding_model import EmbeddingModel
 from lightly_train._models.model_wrapper import ModelWrapper
 from lightly_train._models.package import BasePackage
-from lightly_train.types import DatasetItem, PathLike, Transform
+from lightly_train.types import PathLike, Transform
 
 logger = logging.getLogger(__name__)
 
@@ -430,14 +429,10 @@ def get_dataset_mmap_filenames(
 
 
 def get_dataset(
-    data: PathLike | Sequence[PathLike] | Dataset[DatasetItem],
-    transform: Transform,
+    data: PathLike | Sequence[PathLike],
+    transform: Transform | None,
     mmap_filepath: Path | None,
-) -> Dataset[DatasetItem]:
-    if isinstance(data, Dataset):
-        logger.debug("Using provided dataset.")
-        return data
-
+) -> ImageDataset:
     if mmap_filepath is None:
         raise ValueError("Memory-mapped file path must be provided.")
 
--- a/src/lightly_train/_commands/train.py
+++ b/src/lightly_train/_commands/train.py
@@ -237,12 +237,6 @@ def train_from_config(config: TrainConfig) -> None:
     _system.log_system_information(system_information=system_information)
 
     pytorch_lightning.seed_everything(seed=config.seed, workers=True)
-    config.transform_args = train_helpers.get_transform_args(
-        method=config.method, transform_args=config.transform_args
-    )
-    transform_instance = train_helpers.get_transform(
-        method=config.method, transform_args_resolved=config.transform_args
-    )
     config.float32_matmul_precision = (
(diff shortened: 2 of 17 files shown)


src/lightly_train/_commands/common_helpers.py
```diff
@@ -430,14 +429,10 @@ def get_dataset_mmap_filenames(
 
 
 def get_dataset(
-    data: PathLike | Sequence[PathLike] | Dataset[DatasetItem],
```
liopeer, line 433: We've been wanting to get rid of this for a long time. Now is a good time, since we need any dataset to have the `set_transform` method.

src/lightly_train/_commands/common_helpers.py
```diff
@@ -430,14 +429,10 @@ def get_dataset_mmap_filenames(
...
-    transform: Transform,
+    data: PathLike | Sequence[PathLike],
+    transform: Transform | None,
     mmap_filepath: Path | None,
-) -> Dataset[DatasetItem]:
-    if isinstance(data, Dataset):
-        logger.debug("Using provided dataset.")
-        return data
-
+) -> ImageDataset:
```
liopeer, line 435: Restricting here to datasets that have the `set_transform` method. Not worth creating an interface for now.

src/lightly_train/_commands/train.py
```diff
@@ -259,17 +253,27 @@ def train_from_config(config: TrainConfig) -> None:
...
+        config.transform_args = train_helpers.get_transform_args(
+            method=config.method,
+            wrapped_model=wrapped_model,
+            dataset_size=len(dataset),
+            transform_args=config.transform_args,
+        )
+        transform_instance = train_helpers.get_transform(
+            method=config.method, transform_args_resolved=config.transform_args
+        )
+        dataset.set_transform(transform_instance)
```
liopeer, line 273: We need the dataset size for determininig the transform, i.e. the transform has to be set post-init.

src/lightly_train/_data/image_dataset.py
```diff
@@ -56,6 +59,8 @@ def __getitem__(self, idx: int) -> DatasetItem:
             mask = _open_image(self.mask_dir / maskname, mode="L")
             input["mask"] = mask
 
+        if self.transform is None:
+            raise ValueError("Transform for the dataset is not set.")
```
liopeer, line 63: The error here should probably be a more custom error and not a `ValueError`.

src/lightly_train/_data/image_dataset.py
```diff
@@ -34,14 +34,17 @@ def __init__(
...
-        transform: Transform,
+        transform: Transform | None,
         mask_dir: Path | None = None,
     ):
         self.image_dir = image_dir
         self.image_filenames = image_filenames
         self.mask_dir = mask_dir
         self.transform = transform
 
+    def set_transform(self, transform: Transform) -> None:
```
liopeer, line 45: I prefer explicit setter methods over just setting attributes.

src/lightly_train/_transforms/transform.py
```diff
@@ -131,9 +130,6 @@ class MethodTransformArgs(PydanticConfig):
     solarize: SolarizeArgs | None
 
 
-_T = TypeVar("_T", covariant=True)
```
liopeer, line 134: Saw that this is not used anymore, so I remove it. (unrelated)

src/lightly_train/_commands/train.py
```diff
@@ -259,17 +253,27 @@ def train_from_config(config: TrainConfig) -> None:
...
         dataset = common_helpers.get_dataset(
             data=config.data,
-            transform=transform_instance,
+            transform=None,
             mmap_filepath=mmap_filepath,
         )
+        config.transform_args = train_helpers.get_transform_args(
+            method=config.method,
+            wrapped_model=wrapped_model,
+            dataset_size=len(dataset),
```
JonasWurst, line 267: I think this not required. At least I'm not aware of any scaling based on the dataset size in the transforms

src/lightly_train/_commands/train.py
```diff
@@ -259,17 +253,27 @@ def train_from_config(config: TrainConfig) -> None:
...
+        wrapped_model = package_helpers.get_wrapped_model(
+            model=config.model, model_args=config.model_args
+        )
         dataset = common_helpers.get_dataset(
             data=config.data,
-            transform=transform_instance,
+            transform=None,
             mmap_filepath=mmap_filepath,
         )
+        config.transform_args = train_helpers.get_transform_args(
```
JonasWurst, line 264: what if we have a auto resolver inside transform_args, which also accepts the model args --> with some helper we determine if it is a VIT. If so we validate if the image size and view size is // Patch_size.

The resolved cfg can then be created before the model.

src/lightly_train/_data/image_dataset.py
JonasWurst: agree

PR #114, opened 2025-06-02
Test dinov2 implementation

stegmuel:
## What has changed and why?

Draft fixes for DINOv2.

## How has it been tested?

Tested by running and comparing metrics.
This PR is not intended to be merged.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ dependencies = [
     "pyarrow>=15.0.0",
     "pytorch_lightning>=2.1",
     "tensorboard>=2.10.0",
-    "torch>=2.1.0,<2.6",        # PyTorch Lightning is not yet compatible with PyTorch 2.6
+    "torch>=2.1.0",        # PyTorch Lightning is not yet compatible with PyTorch 2.6
     "torchvision>=0.16.0",      # torchvision 0.15.0 was yanked
     "tqdm>=4.0.0",
 ]
--- a/src/lightly_train/_methods/dinov2/dinov2.py
+++ b/src/lightly_train/_methods/dinov2/dinov2.py
@@ -490,12 +490,12 @@ class DINOv2(Method):
             student_masks_flat=collated_masks,
             n_masked_patches=n_masked_patches,
             masks_weight=masks_weight,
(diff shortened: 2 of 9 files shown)


pyproject.toml
```diff
@@ -32,7 +32,7 @@ dependencies = [
     "pyarrow>=15.0.0",
     "pytorch_lightning>=2.1",
     "tensorboard>=2.10.0",
-    "torch>=2.1.0,<2.6",        # PyTorch Lightning is not yet compatible with PyTorch 2.6
```
yutong-xiang-97, line 35: So does removing this constraint work for you? For me on B200, I still need this constraint to let `make install-dev` work.

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -466,12 +494,12 @@ def training_step_impl(
             student_masks_flat=collated_masks,
             n_masked_patches=n_masked_patches,
             masks_weight=masks_weight,
-        )
```
yutong-xiang-97, line 469: In the original impl. there is a tiny difference between the loss value for logging and backprop: https://github.com/facebookresearch/dinov2/blob/4d83ffd4f4b39df112cef7ed7833de7ebf3a202e/dinov2/train/ssl_meta_arch.py#L317. I tried to follow the original flavor but happy to discuss if we need to adjust this.

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -466,12 +494,12 @@ def training_step_impl(
...
+        ) / n_global_crops
 
         koleo_loss = sum(
             self.koleo_loss(token) for token in student_cls_tokens_global.chunk(2)
-        )  # [G, B, D], only use global views
```
yutong-xiang-97, line 473: Same as above.

pyproject.toml
guarin: Please don't allow torch 2.6 as PyTorch Lightning is not yet fully compatible with it. You can always override this constraint with UV when installing the package.

PR #118, opened 2025-06-04
Unit Tests for DINOv2 Masking

yutong-xiang-97:
## What has changed and why?

Unit tests for `MaskingGenerator` and `create_collated_masks`. Inspired by: https://github.com/lightly-ai/lightly/blob/master/tests/models/test_ModelUtils.py

Note that even with the min and max mask ratio, the algorithm does not guarantee a min or max under certain circumstances.

## How has it been tested?

N/A

- [ ] Yes

- [ ] Yes

### DIFF
--- a/tests/_methods/dinov2/test_utils.py
+++ b/tests/_methods/dinov2/test_utils.py
@@ -8,15 +8,16 @@
 
 import re
 
+import numpy as np
 import pytest
+import torch
 
 from lightly_train._methods.dinov2.utils import MaskingGenerator, create_collated_masks
 
 
 class TestMaskingGenerator:
-    # def setup_method(self) -> None:
-    #     self.grid_size = 16
-    #     self.max_num_patches=int(0.5 * self.grid_size**2)
+    def setup_method(self) -> None:
+        self.grid_size = 16
 
     @pytest.mark.parametrize("grid_size", [14, 16])
     def test_get_shape_and_repr(self, grid_size: int) -> None:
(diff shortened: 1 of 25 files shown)


tests/_methods/dinov2/test_utils.py
```diff
@@ -0,0 +1,190 @@
...
+
+        masking_generator = MaskingGenerator(
+            input_size=(self.grid_size, self.grid_size),
+            max_num_patches=n_masked_patch_tokens,
+            min_aspect=min_aspect_ratio,
+            max_aspect=max_aspect_ratio,
+        )
+
+        mask = masking_generator(n_masked_patch_tokens)
+        assert mask.sum() > 0
```
JonasWurst: I do not think that this test properly tests the aspect ratios. maybe let's just validate the extrem cases as done below?

tests/_methods/dinov2/test_utils.py
```diff
@@ -0,0 +1,190 @@
...
+            n_masked_crops=2,
+            n_crops=4,
+            mask_generator=self.masking_generator,
+        )
+
+        collated_masks = masks["collated_masks"]
+        for mask in collated_masks:
+            n_patch_tokens = mask.numel()
+            n_masked_patch_tokens = mask.sum().item()
+            if n_masked_patch_tokens == 0:
```
JonasWurst, line 206: This will also be the case for the (0.0, 0.0).

Maybe an additional test for the both extrem cases to verify that no or exactly the half is masked?
Though I'm not sure if the exactly 0.5 will be valid because of the lower bound issue

tests/_methods/dinov2/test_utils.py
yutong-xiang-97: I refined the tests to check if the `num_masked_patch_tokens` and `aspect_ratio` result in valid masks. Yet still, with some of the param combinations we will get no masking. E.g. when aspect_ratio is set to 2 and we mask 10%, i.e. 25 patches with `grid_size` 16x16, we will have `h=7` and `w=4` for masks, but since `h*w=28` is bigger than 25, the mask will not be generated. 

Imho, in any case the validity test should still exist, as it tests the supposed logic. However, some of the theoretically possible values may fail the actual test, and it is more because of the suboptimal impl of masking in the original repo. We could ofc fix it easily, but that would deviate ourselves from the original impl.

Do you think we should go with the current solution, or we really want the masking to work as it should?

tests/_methods/dinov2/test_utils.py
yutong-xiang-97: I did a separate test for (0.0, 0.0). However, with masking ratio 0.5 we indeed don't get 128 all the time. Sometimes it will be 127. I would say we just leave it as it is now. There are no hard constraints for #patches masked even with the extreme values.

tests/_methods/dinov2/test_utils.py
JonasWurst: For now I would stick with the original repo. Only if there is proper demand, we can think of adopting.

tests/_methods/dinov2/test_utils.py
```diff
@@ -0,0 +1,216 @@
...
+
+    @pytest.mark.parametrize(
+        "aspect_ratio, masking_percentage, is_masked",
+        [
+            (0.1, 0.005, False),
+            (0.1, 0.05, True),
+            (0.1, 0.5, False),
+            (2.0, 0.001, False),
+            (2.0, 0.01, True),
+            (2.0, 1.0, False),
```
JonasWurst: please add comments to relate to your explanation

JonasWurst (approved):
LGTM

PR #126, opened 2025-06-05
ci: add macOS CI

liopeer:
## What has changed and why?
- still a draft

### DIFF
--- a/.github/workflows/test_unit.yml
+++ b/.github/workflows/test_unit.yml
@@ -75,6 +75,11 @@ jobs:
       - name: Set Up Environment
         run: |
           make install-pinned-${{ matrix.python }}              # Install the dependencies
-      - name: Run Pytest
+      - name: Run Pytest on Ubuntu
+        if: ${{ matrix.runners == 'ubuntu-latest' }}
         run: |
           make test-ci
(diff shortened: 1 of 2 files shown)


JiwaniZakir (commented):
The `ifeq ($(OS),Linux)` check in the Makefile (around line 173) is likely broken. On Linux systems, the `OS` environment variable is not set by the OS itself (unlike Windows where it's `Windows_NT`), so `$(OS)` will be empty on Ubuntu CI runners. This means `PINNED_TORCH_VERSION_PY38`, `PINNED_TORCH_VERSION_PY312`, and related variables will be undefined when running on Linux CI, silently falling back to whatever default behavior the install targets have. The correct approach would be `ifeq ($(shell uname -s),Linux)` or checking `RUNNER_OS` (which GitHub Actions sets to `"Linux"` or `"macOS"`).

Additionally, the `test_unit.yml` file is missing a newline at the end of file (noted in the diff), which is a minor but easily fixed style issue. It's also worth considering whether the "Setup cmake" step should be gated with `if: ${{ matrix.runners == 'ubuntu-latest' }}` the same way FFmpeg is — if cmake isn't needed or behaves differently on macOS, this could cause unexpected failures.

PR #130, opened 2025-06-06
Add weights_only=True to torch.load

yutong-xiang-97:
## What has changed and why?

Add weights_only=True to wherever possible when doing `torch.load()`

## How has it been tested?

- passed unit tests

- [ ] Yes

- [ ] Yes


docs/source/export.md
```diff
@@ -84,10 +84,13 @@ formats are supported.
   from torchvision.models import resnet50
 
   model = resnet50()
-  model.load_state_dict(torch.load("my_exported_model.pt"))
-  ```
+  model.load_state_dict(torch.load("my_exported_model.pt", weights_only=True))
```
guarin, line 87:
````suggestion
  model.load_state_dict(torch.load("my_exported_model.pt", weights_only=True))
  ```
````

File: docs/source/export.md [hunk 2]
```diff
@@ -84,10 +84,13 @@ formats are supported.
...
+
   ````
 
+  ```
+  ```
```
guarin: Remove

docs/source/export.md
```diff
@@ -137,7 +140,7 @@ formats are supported.
   ```python
   import torch
 
-  model = torch.load("my_exported_model.pt")
+  model = torch.load("my_exported_model.pt", weights_only=True)
```
guarin:
```suggestion
  model = torch.load("my_exported_model.pt")
```

I think `weights_only=True` should only be used when loading a state dict. In this case here we are actually loading a full model class which will fail with `weights_only=True`.

guarin (changes requested):

PR #137, opened 2025-06-11
Thomas trn 479 change lr scaling in distillation methods

stegmuel:
## What has changed and why?

The scaling of the learning is changed from linear to "square root" to prevent too small learning with small batches.

## How has it been tested?

Tests will be added to verify that the resulting learning rate is unchanged for the recommended batchsize.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_methods/distillation/distillation.py
+++ b/src/lightly_train/_methods/distillation/distillation.py
@@ -8,10 +8,13 @@
 from __future__ import annotations
 
 import logging
+import math
 from typing import Any, Literal, Mapping, cast
 
 import torch
 import torch.nn.functional as F
+from lightly.utils.scheduler import CosineWarmupScheduler
+from pytorch_lightning.utilities.types import OptimizerLRScheduler
 from torch import Tensor
 from torch.nn import Flatten, Linear, Module, init
 from torch.nn.modules.module import _IncompatibleKeys
@@ -28,6 +31,7 @@ from lightly_train._methods.method_args import MethodArgs
 from lightly_train._models import package_helpers
 from lightly_train._models.embedding_model import EmbeddingModel
 from lightly_train._models.model_wrapper import ModelWrapper
+from lightly_train._optim import optimizer_helpers
 from lightly_train._optim.lars_args import LARSArgs
 from lightly_train._optim.optimizer_args import OptimizerArgs
 from lightly_train._optim.optimizer_type import OptimizerType
@@ -95,7 +99,7 @@ class DistillationArgs(MethodArgs):
 
 
 class DistillationLARSArgs(LARSArgs):
-    lr: float = 0.3
+    lr: float = 1.8  # 1.8 = 0.3 * 1536 / 256
     momentum: float = 0.9
     dampening: float = 0
     weight_decay: float = 1e-6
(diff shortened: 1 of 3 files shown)


src/lightly_train/_methods/distillation/distillation.py
```diff
@@ -246,6 +250,36 @@ def trainable_modules(self) -> TrainableModules:
             modules=[self.student_embedding_model, self.student_projection_head]
         )
 
+    def configure_optimizers(self) -> OptimizerLRScheduler:
+        lr_scale = math.sqrt(self.global_batch_size / 1536)  # type: ignore[attr-defined]
```
guarin: Does sqrt scaling also make sense for large batch sizes?

src/lightly_train/_methods/distillation/distillation.py
```diff
@@ -246,6 +250,36 @@ def trainable_modules(self) -> TrainableModules:
             modules=[self.student_embedding_model, self.student_projection_head]
         )
 
+    def configure_optimizers(self) -> OptimizerLRScheduler:
```
guarin: Could we add a `lr_scale_method = "linear" | "sqrt"` argument to the optimizer args and handle the scaling in `Method.configure_optimizers` instead of copying the `configure_optimizers` for all method implementations?

guarin (changes requested):

src/lightly_train/_methods/distillation/distillation.py
stegmuel: We change from linear to sqrt because we think linear is too sensitive to the batch size. That holds for both directions, no?

src/lightly_train/_methods/distillation/distillation.py
stegmuel: I added the argument to `method_args` instead.

src/lightly_train/_methods/method.py
```diff
@@ -85,10 +86,15 @@ def trainable_modules(self) -> TrainableModules:
     # Ignore the return type, because pytorch-lightning types it wrongly.
     # See https://github.com/Lightning-AI/pytorch-lightning/issues/20106
     def configure_optimizers(self) -> OptimizerLRScheduler:
+        # Scale the learning rate based on the global batch size.
+        lr_scale = self.global_batch_size / self.method_args.reference_batch_size
+        if self.method_args.lr_scale_method == "sqrt":
+            lr_scale = math.sqrt(lr_scale)  # type: ignore[attr-defined]
```
JonasWurst: can't you avoid this by adding typing to line 90?

`lr_scale: float = self.global_batch_size / self.method_args.reference_batch_size`

JonasWurst (commented):
Also in dinov2 it needs to be modified + tested

tests/_methods/test_method.py
```diff
@@ -0,0 +1,94 @@
...
+        embedding_model = EmbeddingModel(wrapped_model=DummyCustomModel(embed_dim))
+
+        # Compute expected learning rate.
+        base_lr = optimizer_args.lr
+        expected_scale = global_batch_size / reference_batch_size
+        if lr_scale_method == "sqrt":
+            expected_scale = math.sqrt(expected_scale)
+        expected_lr = base_lr * expected_scale
+
+        # Instantiate distillation method.
```
JonasWurst:
```suggestion
        # Instantiate dummy method.
```

src/lightly_train/_methods/method.py
stegmuel: Changed

PR #144, opened 2025-06-20
Add ChannelDrop augmentation

guarin:
## What has changed and why?

* Add channel drop augmentation
* Add option to load multi-channel images with LIGHTLY_TRAIN_IMAGE_MODE="UNCHANGED"

Note that we still assume that all images have 3-channels for the augmentations. So `LIGHTLY_TRAIN_IMAGE_MODE="UNCHANGED"` must be used in combination with `channel_drop=dict(num_channels_keep=3, weight_drop=(...))`. To make sure that only 3 channels are kept. 

Will add proper support for multi-channel later. 

## How has it been tested?

* Added unit tests
* Ran training with NRGB images and the two new options

- [x] Yes

- [x] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add channel drop transform.
+- Add option to load multi-channel images with `LIGHTLY_TRAIN_IMAGE_MODE="UNCHANGED"`.
+
 ### Changed
 
 ### Deprecated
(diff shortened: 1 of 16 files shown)


stegmuel (approved):
Looks good.
I have two questions:
- We support `num_channels_keep != 3`, but that would in any cases causes error currently. Should we temporarily remove the option to set it?
- Can we think of realistic scenarios where one would load the images in `RGB` mode and drop channels? Can't we infer the `image_mode` from channels dropping being used or not?  I would expect users to forget to set the env variable.

PR #149, opened 2025-06-23
Allow DINOv2 weights loading for Distillation

yutong-xiang-97:
## What has changed and why?

Support loading DINOv2 teacher weights from a pretrained checkpoint by allowing such input

```
lightly_train.train(
  ...,
  method="distillation",
  method_args={
    "teacher": "dinov2_vit/vitb14",
    "teacher_weights": "out/dinov2/exported_models/exported_last.pt", # pretrained `dinov2_vit/vitb14` weights 
  }
)
```

Also added unit tests for this. More specifically, to enable the test, we

- renamed `vit_tiny__testing` to `_vit_test` and added it to the available model list so that it can be loaded via `lightly_train.train(model="dinov2_vit/_vit_test")
- moved `dummy_vit_model` to `helpers` so that it can be used across all the tests

## How has it been tested?

By running the unit tests.

- [x] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_methods/distillation/distillation.py
+++ b/src/lightly_train/_methods/distillation/distillation.py
@@ -47,8 +47,10 @@ def get_teacher(teacher_name: str, teacher_weights: str | Path | None = None) ->
     teacher_embedding_model = wrapped_model.get_model()
     assert isinstance(teacher_embedding_model, Module)
 
-    # If a path to the teacher weights is provided, load them.
-    if teacher_weights is not None:
+    # If a path to the pretrained teacher weights is provided, load them.
+    # Ignore the weights when the pretrained teacher is provided so as to prevent loading
+    # the weights twice.
+    if (not teacher_name.endswith("_pretrain")) and (teacher_weights is not None):
         if not Path(teacher_weights).exists():
             raise FileNotFoundError(
                 f"Teacher weights file {teacher_weights} does not exist."
--- a/src/lightly_train/_methods/distillation/distillation.py
+++ b/src/lightly_train/_methods/distillation/distillation.py
@@ -47,10 +47,8 @@ def get_teacher(teacher_name: str, teacher_weights: str | Path | None = None) ->
     teacher_embedding_model = wrapped_model.get_model()
     assert isinstance(teacher_embedding_model, Module)
 
-    # If a path to the pretrained teacher weights is provided, load them.
-    # Ignore the weights when the pretrained teacher is provided so as to prevent loading
-    # the weights twice.
-    if (not teacher_name.endswith("_pretrain")) and (teacher_weights is not None):
+    # If a path to the teacher weights is provided, load them.
+    if teacher_weights is not None:
         if not Path(teacher_weights).exists():
             raise FileNotFoundError(
                 f"Teacher weights file {teacher_weights} does not exist."
--- a/src/lightly_train/_methods/distillation/distillation.py
+++ b/src/lightly_train/_methods/distillation/distillation.py
@@ -8,6 +8,7 @@
 from __future__ import annotations
 
 import logging
+from pathlib import Path
 from typing import Any, Literal, Mapping, cast
 
 import torch
@@ -41,11 +42,24 @@ from lightly_train.types import Batch
 logger = logging.getLogger(__name__)
 
 
-def get_teacher(teacher_name: str) -> Module:
+def get_teacher(teacher_name: str, teacher_weights: str | Path | None = None) -> Module:
     wrapped_model = package_helpers.get_wrapped_model(model=teacher_name)
     teacher_embedding_model = wrapped_model.get_model()
     assert isinstance(teacher_embedding_model, Module)
+
+    # If a path to the teacher weights is provided, load them.
+    if teacher_weights is not None:
+        if not Path(teacher_weights).exists():
+            raise FileNotFoundError(
+                f"Teacher weights file {teacher_weights} does not exist."
+            )
+
+        state_dict = torch.load(teacher_weights, weights_only=True)
+        teacher_embedding_model.load_state_dict(state_dict)
+        logger.debug(f"Loaded teacher weights from {teacher_weights}.")
+
     teacher_embedding_model.eval()
+
     return teacher_embedding_model
 
 
@@ -60,6 +74,9 @@ class DistillationArgs(MethodArgs):
 
     # Default teacher.
     teacher: str = "dinov2_vit/vitb14_pretrain"
+
+    # Optional teacher weight path.
+    teacher_weights: str | Path | None = None
 
     # Scaling method for the learning rate.
     lr_scale_method: Literal["linear", "sqrt"] = "sqrt"
@@ -123,7 +140,9 @@ class Distillation(Method):
             global_batch_size=global_batch_size,
         )
         # Get the teacher model.
-        self.teacher_embedding_model = get_teacher(method_args.teacher)
+        self.teacher_embedding_model = get_teacher(
+            method_args.teacher, method_args.teacher_weights
+        )
 
         # Store the student model.
         self.student_embedding_model = embedding_model
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Allow loading pretrained DINOv2 teacher weights for distillation methods with an extra `teacher_weights` argument in `method_args`.
+
 ### Deprecated
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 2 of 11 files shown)


src/lightly_train/_methods/distillation/distillation.py
```diff
@@ -41,11 +42,24 @@
 logger = logging.getLogger(__name__)
 
 
-def get_teacher(teacher_name: str) -> Module:
+def get_teacher(teacher_name: str, teacher_weights: str | Path | None = None) -> Module:
     wrapped_model = package_helpers.get_wrapped_model(model=teacher_name)
```
JonasWurst, line 46: in case `"_pretrained"` is used, this will cause loading twice: from the original weights and from the new loading logic here. We could prevent that by dropping the postfix in case we have a valid path

src/lightly_train/_methods/distillation/distillation.py
```diff
@@ -41,11 +42,24 @@
...
+    if teacher_weights is not None:
+        if not Path(teacher_weights).exists():
+            raise FileNotFoundError(
+                f"Teacher weights file {teacher_weights} does not exist."
+            )
+
+        state_dict = torch.load(teacher_weights, weights_only=True)
+        teacher_embedding_model.load_state_dict(state_dict)
+        logger.debug(f"Loaded teacher weights from {teacher_weights}.")
+
```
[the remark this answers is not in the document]
JonasWurst: At the moment it is accepted since we intentionally decided for this route, refactoring will happen later.

yutong-xiang-97:
> I think we can add a test: we have a [tiny vit model](https://github.com/lightly-ai/lightly-train/blob/c63a7e52852d1d09f58d8e483d292174e3df3d94/src/lightly_train/_models/dinov2_vit/dinov2_vit_src/models/vision_transformer.py#L461) defined for testing:
> 
> 1. initalize dinov2 train
> 2. force export
> 3. Load as teacher
> 
> we would need to add a config for it, but that should be minor effort

Then how should we assert that it works as expected? I thought about adding unit tests and was not able to find a way without doing actual training.

JonasWurst:
Yes, the model will not be meaningful, however we would make sure that the state_dict has the keys as expected

src/lightly_train/_methods/distillation/distillation.py
yutong-xiang-97: The `vits14` and `vits14_pretrained` already have different key names. Also, in this case, the `teacher_weights` can be the `.pt` files pretrained by LightlyTrain with other data on top of the original `vits14_pretrained`?

JonasWurst: yes, you are right. I'm referring to the scenario where a user does not change let's say `vits14_pretrained` and provides a `.pt`as well. Then first the official weighty will be loaded. and then the local weights will be loaded. Hence, the first loading is unnecessary and might seem odd if someone investigates the log.

yutong-xiang-97:
> Yes, the model will not be meaningful, however we would make sure that the state_dict has the keys as expected

The `vitb14` and `vitb14_pretrained` have slight difference in keys and params:
```
RuntimeError: Error(s) in loading state_dict for DinoVisionTransformer:
        Missing key(s) in state_dict: "blocks.0.mlp.fc1.weight", "blocks.0.mlp.fc1.bias", "blocks.0.mlp.fc2.weight", "blocks.0.mlp.fc2.bias", "blocks.1.mlp.fc1.weight", "blocks.1.mlp.fc1.bias", "blocks.1.mlp.fc2.weight", "blocks.1.mlp.fc2.bias", "blocks.2.mlp.fc1.weight", "blocks.2.mlp.fc1.bias", "blocks.2.mlp.fc2.weight", "blocks.2.mlp.fc2.bias", "blocks.3.mlp.fc1.weight", "blocks.3.mlp.fc1.bias", "blocks.3.mlp.fc2.weight", "blocks.3.mlp.fc2.bias", "blocks.4.mlp.fc1.weight", "blocks.4.mlp.fc1.bias", "blocks.4.mlp.fc2.weight", "blocks.4.mlp.fc2.bias", "blocks.5.mlp.fc1.weight", "blocks.5.mlp.fc1.bias", "blocks.5.mlp.fc2.weight", "blocks.5.mlp.fc2.bias", "blocks.6.mlp.fc1.weight", "blocks.6.mlp.fc1.bias", "blocks.6.mlp.fc2.weight", "blocks.6.mlp.fc2.bias", "blocks.7.mlp.fc1.weight", "blocks.7.mlp.fc1.bias", "blocks.7.mlp.fc2.weight", "blocks.7.mlp.fc2.bias", "blocks.8.mlp.fc1.weight", "blocks.8.mlp.fc1.bias", "blocks.8.mlp.fc2.weight", "blocks.8.mlp.fc2.bias", "blocks.9.mlp.fc1.weight", "blocks.9.mlp.fc1.bias", "blocks.9.mlp.fc2.weight", "blocks.9.mlp.fc2.bias", "blocks.10.mlp.fc1.weight", "blocks.10.mlp.fc1.bias", "blocks.10.mlp.fc2.weight", "blocks.10.mlp.fc2.bias", "blocks.11.mlp.fc1.weight", "blocks.11.mlp.fc1.bias", "blocks.11.mlp.fc2.weight", "blocks.11.mlp.fc2.bias". 
        Unexpected key(s) in state_dict: "blocks.0.mlp.w12.weight", "blocks.0.mlp.w12.bias", "blocks.0.mlp.w3.weight", "blocks.0.mlp.w3.bias", "blocks.1.mlp.w12.weight", "blocks.1.mlp.w12.bias", "blocks.1.mlp.w3.weight", "blocks.1.mlp.w3.bias", "blocks.2.mlp.w12.weight", "blocks.2.mlp.w12.bias", "blocks.2.mlp.w3.weight", "blocks.2.mlp.w3.bias", "blocks.3.mlp.w12.weight", "blocks.3.mlp.w12.bias", "blocks.3.mlp.w3.weight", "blocks.3.mlp.w3.bias", "blocks.4.mlp.w12.weight", "blocks.4.mlp.w12.bias", "blocks.4.mlp.w3.weight", "blocks.4.mlp.w3.bias", "blocks.5.mlp.w12.weight", "blocks.5.mlp.w12.bias", "blocks.5.mlp.w3.weight", "blocks.5.mlp.w3.bias", "blocks.6.mlp.w12.weight", "blocks.6.mlp.w12.bias", "blocks.6.mlp.w3.weight", "blocks.6.mlp.w3.bias", "blocks.7.mlp.w12.weight", "blocks.7.mlp.w12.bias", "blocks.7.mlp.w3.weight", "blocks.7.mlp.w3.bias", "blocks.8.mlp.w12.weight", "blocks.8.mlp.w12.bias", "blocks.8.mlp.w3.weight", "blocks.8.mlp.w3.bias", "blocks.9.mlp.w12.weight", "blocks.9.mlp.w12.bias", "blocks.9.mlp.w3.weight", "blocks.9.mlp.w3.bias", "blocks.10.mlp.w12.weight", "blocks.10.mlp.w12.bias", "blocks.10.mlp.w3.weight", "blocks.10.mlp.w3.bias", "blocks.11.mlp.w12.weight", "blocks.11.mlp.w12.bias", "blocks.11.mlp.w3.weight", "blocks.11.mlp.w3.bias". 
        size mismatch for pos_embed: copying a param with shape torch.Size([1, 257, 768]) from checkpoint, the shape in current model is torch.Size([1, 1370, 768]).
```
If users provides a mismatched input, they should receive a runtime error like above. 

What would be the benefit in testing the matching keys in this case?

JonasWurst:
Yes, that is fine. This is due to the different crop size, hence we might need to adjust the base config. At some point we need to look into the image size, since so far, the image size specified in the transform, is not linked here. -> can you create a follow up to look into this?

To solve your current testing issue: you could create a config for tiny. I highly recommned not using the b for testing.

yutong-xiang-97:
> Yes, that is fine. This is due to the different crop size, hence we might need to adjust the base config. At some point we need to look into the image size, since so far, the image size specified in the transform, is not linked here. -> can you create a follow up to look into this?
> 
> To solve your current testing issue: you could create a config for tiny. I highly recommned not using the b for testing.

For this tiny vit to be loaded as teacher (which is a string in the method_args, it has to have a model name exposed in `list_model_names`). 

Imho it does not worth that much effort to create a whole test just to make sure the keys match. As I said, a runtime error should be sufficient. WDYT?

src/lightly_train/_methods/distillation/distillation.py
```diff
@@ -47,8 +47,10 @@ def get_teacher(teacher_name: str, teacher_weights: str | Path | None = None) ->
     teacher_embedding_model = wrapped_model.get_model()
     assert isinstance(teacher_embedding_model, Module)
 
-    # If a path to the teacher weights is provided, load them.
-    if teacher_weights is not None:
+    # If a path to the pretrained teacher weights is provided, load them.
+    # Ignore the weights when the pretrained teacher is provided so as to prevent loading
+    # the weights twice.
+    if (not teacher_name.endswith("_pretrain")) and (teacher_weights is not None):
```
JonasWurst: If we do skip the provided weights, we should warn about it, and hint that instead a different model should be used.

However for this we need to adjust the train config.

JonasWurst (approved):
LGTM

Issue #151, opened 2025-06-25
[QUESTION] It's time-consuming to write the filename every time


Devoe-97:
### 🤔 What’s your question?

Every time training need to write all the filenames to the temp file, this process is very time-consuming, may I ask if it is possible to load the temp file that has been generated by the last training, so as to save the time of re-writing the filenames every time?

### 🧠 What did you try?
I tried backing up the temp file from the last training, but the temp file name is different for every training

liopeer:
Hi @Devoe-97 and thanks for the question!

The way how the file is generated is by using the following logic:

https://github.com/lightly-ai/lightly-train/blob/22c855df2dfc89996436bbe0324d7de06803fe71/src/lightly_train/_commands/common_helpers.py#L396-L398

So, we generate the name/path of the tempfile based on the output directory that you specified and the node rank (every node gets its dedicated file). This means that whenever the output directory changes (which is usually the case for new runs), the path will also change.

## Why did we implement it like this?
Usually, we want to check during the setup of the training which images are actually in the specified folders and then we want to create the file with the paths. Of course this will take some time to set up when you have a lot of images, but we thought it's the safest way. Otherwise if there are small differences in the dataset this might cause a runtime error at some point during the epoch because the dataloader did e.g. not find an image that is still listed in the file.

## What we could consider changing
For advanced users that all the time use the exact same dataset, the setup is indeed quite long and time-consuming. I don't think we want to expose passing a path to a tempfile directly in the `lightly_train.train` command, however we could think about exposing an environment variable instead, e.g. a `LIGHTLY_TRAIN_MMAP_FILEPATH`. Optimally this file would know through which `data` argument to the `train` command it was generated and like this we could quickly check if the configuration still makes sense and otherwise throw an error.

How does that sound for you? Also, may I ask how large your dataset is and how long it takes to create the memory-mapped file?

Devoe-97:
> Hi [@Devoe-97](https://github.com/Devoe-97) and thanks for the question!
> 
> The way how the file is generated is by using the following logic:
> 
> [lightly-train/src/lightly_train/_commands/common_helpers.py](https://github.com/lightly-ai/lightly-train/blob/22c855df2dfc89996436bbe0324d7de06803fe71/src/lightly_train/_commands/common_helpers.py#L396-L398)
> 
> Lines 396 to 398 in [22c855d](/lightly-ai/lightly-train/commit/22c855df2dfc89996436bbe0324d7de06803fe71)
> 
[...]

PR #155, opened 2025-06-30
Trn 617 reuse memory mapped file

JonasWurst:
## What has changed and why?

For large detests reusing the mmap file can mean a high speed up for the dataset initialization. #151 

This PR brings the required changes. Usage:
To enable the permanent saving of the mmap file, the environment variable LIGHTLY_TRAIN_MMAP_REUSE_FILE is introduced. If this is true, an existing mmap file for the same input data_arg will be reused if it exists, else it will be created.
`os.environ["LIGHTLY_TRAIN_MMAP_REUSE_FILE"] = "1"`

## How has it been tested?

Unit tests are introduced.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/common_helpers.py
+++ b/src/lightly_train/_commands/common_helpers.py
@@ -375,14 +375,15 @@ def _get_package(model: Module) -> BasePackage:
 
 
 @contextlib.contextmanager
-def get_dataset_temp_mmap_path(out: Path) -> Generator[Path, Any, Any]:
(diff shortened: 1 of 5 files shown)


JonasWurst:
> Thanks, this looks pretty good! However, I have one concern regarding the hashing directly from data: Did you try hashing if someone passes e.g. a very long sequence of folders/images into the train or embed commands (~ several thousands)?

Good point.
I tested hashing:
- a list of 70_000 folders: 0.01 seconds
- a list of 2_863_670 images: 0.47 seconds

I think this is fine.

liopeer (approved):
LGTM. Thanks for the additional check!

PR #156, opened 2025-06-30
Trn 617 reuse memory mapped file

JonasWurst:
## What has changed and why?

This PR aims to bring improvements to the dataset preparation. It focuses on the file listing iterator.

### DIFF
--- a/src/lightly_train/_commands/common_helpers.py
+++ b/src/lightly_train/_commands/common_helpers.py
@@ -480,8 +480,6 @@ def get_dataset(
 
     mask_dir = Env.LIGHTLY_TRAIN_MASK_DIR.value
 
-    # NOTE(Jonas 06/2025): If we would use abspaths on both cases, we could only use the
-    #  _get_image_filepaths for both cases. This would remove some logic.
     if isinstance(data, (str, Path)):
         data = Path(data).resolve()
         if not data.exists():
@@ -492,6 +490,9 @@ def get_dataset(
             raise ValueError(f"Data directory '{data}' is empty!")
         # Use relative paths as filenames when a single directory or file is provided to
         # reduce the file size.
+        # NOTE(Guarin, 01/25): The bottleneck for dataset initialization is filename
+        # listing and not the memory mapping. Listing the train set from ImageNet takes
+        # about 30 seconds. This is mostly because os.walk is not parallelized.
         filenames = image_dataset.list_image_filenames(image_dir=data)
         return ImageDataset(
             image_dir=data,
--- a/src/lightly_train/_commands/common_helpers.py
+++ b/src/lightly_train/_commands/common_helpers.py
@@ -375,14 +375,15 @@ def _get_package(model: Module) -> BasePackage:
 
 
 @contextlib.contextmanager
-def get_dataset_temp_mmap_path(out: Path) -> Generator[Path, Any, Any]:
+def get_dataset_temp_mmap_path(data: PathLike | Sequence[PathLike]) -> Generator[Path, Any, Any]:
     """Generate file in temporary directory to be used for memory-mapping the dataset.
 
-    Creates a unique filename for the memory-mapped file based on the out path.
-    We use the out path as a deterministic value that is consistent across all ranks
-    on the same node.
+    Creates a unique filename for the memory-mapped file based on the data arg.
+    We use the data arg as a deterministic value that is consistent across all ranks
+    on the same node. Additionally, we can cache the file if required, since the hash 
+    directly reflects the used config.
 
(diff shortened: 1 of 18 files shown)


src/lightly_train/_data/image_dataset.py
```diff
@@ -130,14 +133,48 @@ def list_image_files(imgs_and_dirs: Sequence[Path]) -> Iterable[Path]:
             raise ValueError(f"Invalid path: {img_or_dir}")
 
 
-def _get_image_filepaths(image_dir: Path) -> Iterable[Path]:
+def _get_image_filepaths(image_dir: Path) -> Iterable[str]:
```
JonasWurst: In dummy tests, those changes showed a strong improvement (5x). However, comparing the train command old vs new, we only save around 20%.

src/lightly_train/_data/image_dataset.py
```diff
@@ -99,17 +100,18 @@ def list_image_filenames(
             "Either `image_dir` or `files` must be provided, but not both."
         )
     elif files is not None:
-        return (ImageFilename(str(fpath.resolve())) for fpath in files)
```
JonasWurst, line 102: Since we resolve the paths on high level, we do not need to ensure this here. This had a very strong impact.

This change is independent of the others and could be kept.

src/lightly_train/_data/image_dataset.py
```diff
@@ -99,17 +100,18 @@ def list_image_filenames(
...
+        return (ImageFilename(fpath) for fpath in files)
     elif image_dir is not None:
+        image_dir_str = str(image_dir)
         return (
-            ImageFilename(str(fpath.relative_to(image_dir)))
+            ImageFilename(os.path.relpath(fpath, start=image_dir_str))
```
JonasWurst: if we opt for not converting to relpath we could get another speed improvement ->

no relpath conversion + avoid os walk: in total it would be around 3x faster than the current version

src/lightly_train/_data/image_dataset.py
guarin: Are the 20% for the mmap process or how exactly did you measure this? And single or multi-GPU?

src/lightly_train/_data/image_dataset.py
guarin: What does high level mean here? Where do we resolve the paths?

src/lightly_train/_commands/common_helpers.py
```diff
@@ -490,9 +492,6 @@ def get_dataset(
             raise ValueError(f"Data directory '{data}' is empty!")
         # Use relative paths as filenames when a single directory or file is provided to
         # reduce the file size.
-        # NOTE(Guarin, 01/25): The bottleneck for dataset initialization is filename
-        # listing and not the memory mapping. Listing the train set from ImageNet takes
-        # about 30 seconds. This is mostly because os.walk is not parallelized.
```
guarin: Is this not true anymore? What is the bottleneck now?

src/lightly_train/_data/image_dataset.py
guarin: IMO it is ok to return the relative path to `image_dir` from `_get_image_filepaths`. Then we can skip the relpath here. We'll still have to be careful to return the correct paths from `_get_image_filepaths` when handling nested directories.

JonasWurst: Realizing it in `_get_image_filepaths` would not make any difference. If relpaths are needed to keep the filesize low, then we need to accept the slower execution.

src/lightly_train/_commands/common_helpers.py
JonasWurst: os.walk is not used in this version. Hence i wanted to avoid confusion here. But the file listing is still the bottleneck yes.

src/lightly_train/_data/image_dataset.py
JonasWurst: resolving is done here: https://github.com/lightly-ai/lightly-train/blob/d941efb077c105d289a9ebffe42c4e6b44ff9def/src/lightly_train/_commands/common_helpers.py#L508

src/lightly_train/_data/image_dataset.py
JonasWurst: yes it is only for the mmap. However, I think we agree that anything around that number is not worth to trigger a change.

src/lightly_train/_data/image_dataset.py
guarin: But that resolve is only for the input `data` sequence, not for the listed files.

guarin (approved):

PR #157, opened 2025-07-01
Add train task structure

guarin:

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_data/mask_semantic_segmentation_dataset.py
+++ b/src/lightly_train/_data/mask_semantic_segmentation_dataset.py
@@ -1,0 +1,20 @@
+#
+# Copyright (c) Lightly AG and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+#
+import torch
+from torch.utils.data import Dataset
+
+from lightly_train.types import MaskSemanticSegmentationDatasetItem
+
+
+class MaskSemanticSegmentationDataset(Dataset[MaskSemanticSegmentationDatasetItem]):
+    def __getitem__(self, index: int) -> MaskSemanticSegmentationDatasetItem:
+        return {
+            "image_filename": "abc.jpg",
+            "image": torch.rand(3, 224, 224, dtype=torch.uint8),
+            "mask": torch.rand(5, 224, 224, dtype=torch.uint8),
+        }
--- a/src/lightly_train/_task_models/dinov2_semantic_segmentation/dinov2_semantic_segmentation.py
+++ b/src/lightly_train/_task_models/dinov2_semantic_segmentation/dinov2_semantic_segmentation.py
@@ -1,0 +1,26 @@
+#
+# Copyright (c) Lightly AG and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+#
+from __future__ import annotations
+
+from torch import Module, Tensor
+
+from lightly_train._models.dinov2_vit.dinov2_vit_src.models.vision_transformer import (
+    DinoVisionTransformer,
+)
+from lightly_train._task_models.task_model import TaskModel
+
+
+class DINOv2SemanticSegmentation(TaskModel):
+    def __init__(self) -> None:
+        super().__init__()
+        self.backbone: DinoVisionTransformer
+        self.head: Module
+
+    def forward(self, x: Tensor) -> Tensor:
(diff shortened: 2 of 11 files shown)


pyproject.toml
```diff
@@ -112,6 +112,7 @@ warn_unused_ignores = false
 module = [
     "albumentations.*",
     "lightly.*",
+    "lightning_fabric.*",
```
stegmuel, line 115: With fabric we will be able to use torch 2.7? Can we have temporarily two distinct set of requirements?

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,20 @@
...
+
+from lightly_train.types import MaskSemanticSegmentationDatasetItem
+
+
+class MaskSemanticSegmentationDataset(Dataset[MaskSemanticSegmentationDatasetItem]):
+    def __getitem__(self, index: int) -> MaskSemanticSegmentationDatasetItem:
+        return {
+            "image_filename": "abc.jpg",
+            "image": torch.rand(3, 224, 224, dtype=torch.uint8),
+            "mask": torch.rand(5, 224, 224, dtype=torch.uint8),
```
stegmuel, line 19: If we go for semantic segmentation we should got for a single channel and values in the range [0, C-1]. I have tried stacking everything once and the overhead was huge.

stegmuel: But we have to anticipate the case where some pixels could have multiple labels.

src/lightly_train/_task_models/dinov2_semantic_segmentation/dinov2_semantic_segmentation.py
```diff
@@ -0,0 +1,26 @@
...
+from lightly_train._models.dinov2_vit.dinov2_vit_src.models.vision_transformer import (
+    DinoVisionTransformer,
+)
+from lightly_train._task_models.task_model import TaskModel
+
+
+class DINOv2SemanticSegmentation(TaskModel):
+    def __init__(self) -> None:
+        super().__init__()
+        self.backbone: DinoVisionTransformer
```
stegmuel, line 21: Out of the box we will be able to load from `exported_last.pt`, right?

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
stegmuel: Throwing some random info: very often it is needed to ignore some labels as they are the border of objects or background.

src/lightly_train/_task_models/task_model.py
```diff
@@ -0,0 +1,18 @@
...
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+#
+from torch import Module
+
+
+class TaskModel(Module):
+    """Base class for task-specific models that the user interacts with.
+
+    Must implement the forward method for inference. Must be pure PyTorch and not rely
```
stegmuel, line 14: Why is the inference of the task-specific models not sufficient?

pyproject.toml
guarin: This change here is only to ignore some mypy warnings. Fabric and Lightning are tied to each other with the same versions. We'll have the same torch 2.7 support as we have now. We could add extra requirements with `lightly-train[task]` but I don't think we should do it, it will make installing the library more complicated and is harder to maintain.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
guarin: Ah yes good point, I just put something there 😅 

> Throwing some random info: very often it is needed to ignore some labels as they are the border of objects or background.

This we can handle by adding more fields to `data`. We don't need this info in the batches itself.

src/lightly_train/_task_models/dinov2_semantic_segmentation/dinov2_semantic_segmentation.py
guarin: Yes that should work as long as we load it directly into the backbone with `self.backbone.load_state_dict`. For proper load/export on the model level we'll have to make some more changes.

src/lightly_train/_task_models/task_model.py
guarin: Sufficient for what?

stegmuel: Answered myself.

PR #158, opened 2025-07-01
Add MaskSemanticSegmentationDataset

guarin:
## What has changed and why?

* Add MaskSemanticSegmentationDataset
* Add `_data/file_helpers.py`

## How has it been tested?

Added unit tests

- [ ] Yes

- [ ] Yes

### DIFF
--- a/tests/_data/test_mask_semantic_segmentation_dataset.py
+++ b/tests/_data/test_mask_semantic_segmentation_dataset.py
@@ -35,7 +35,7 @@ class DummyTransform(TaskTransform):
             [
                 A.Resize(32, 32),
                 A.Normalize(mean=(0.5, 0.5, 0.5), std=(0.5, 0.5, 0.5)),
-                A.ToTensorV2(),
+                A.pytorch.transforms.ToTensorV2(),
             ]
         )
 


src/lightly_train/_data/file_helpers.py
guarin, line 1: Everything in this file was moved from `image_dataset.py`

src/lightly_train/_data/image_dataset.py
```diff
@@ -77,99 +69,3 @@ def __getitem__(self, idx: int) -> DatasetItem:
 
     def __len__(self) -> int:
         return len(self.image_filenames)
-
-
-def list_image_filenames(
-    *, image_dir: Path | None = None, files: Iterable[Path] | None = None
```
guarin, line 83: Moved to `file_helpers.py`

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -5,16 +5,95 @@
...
+    MaskSemanticSegmentationDatasetItem,
+    PathLike,
+)
 
 
 class MaskSemanticSegmentationDataset(Dataset[MaskSemanticSegmentationDatasetItem]):
+    def __init__(
+        self,
+        dataset_args: MaskSemanticSegmentationDatasetArgs,
+        image_filenames: Sequence[ImageFilename],
```
guarin, line 30: `image_filenames` will be mmapped, this will happen outside the dataset class

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -5,16 +5,95 @@
...
-            "image_filename": "abc.jpg",
-            "image": torch.rand(3, 224, 224, dtype=torch.uint8),
-            "mask": torch.rand(5, 224, 224, dtype=torch.uint8),
+            "image_path": str(image_path),  # Str for torch dataloader compatibility.
+            "image": transformed["image"],
+            "mask": transformed["mask"],
         }
+
+
+class MaskSemanticSegmentationDatasetArgs(PydanticConfig):
```
guarin, line 56: The way the classes are structured it should be easy to go from user input to dataset like this:

```
data = {"train": {"images": ..., "masks": ...}, ...}
data_args: MaskSemanticSegmentationDataArgs = validate(data_args) # happens when we validate TrainTaskArgs
train_args = data_args.get_train_args()
train_filenames = train_args.list_image_filenames()
train_dataset_cls = train_args.get_dataset_cls()
train_dataset = train_dataset_cls(train_args, train_filenames, transform)

# Same for val...
# Later on we can add support for multiple validation sets
```

Validation is actually already tested in `test_train_task.py` so we know it works.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -5,16 +5,95 @@
...
+    def __len__(self) -> int:
+        return len(self.image_filenames)
+
     def __getitem__(self, index: int) -> MaskSemanticSegmentationDatasetItem:
+        image_filename = self.image_filenames[index]
+        image_path = self.args.image_dir / image_filename
+        mask_path = (self.args.mask_dir / image_filename).with_suffix(".png")
+
+        image = file_helpers.open_image(image_path=image_path, mode="RGB")
+        mask = file_helpers.open_image(image_path=mask_path, mode="L")
```
stegmuel: With this approach we cannot handle datasets with more than 256 classes, no?
Supposedly `mode="I"` is for 16-bit.
The assumption that we'll have the masks in a directory with the same name and in .png is quite strong, but I guess we'll update based on the needs.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
stegmuel: `data_args` is what the user will pass to `train_task` through the `data` argument?

tests/_data/test_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,71 @@
...
+            image_filenames=list(dataset_args.list_image_filenames()),
+            transform=transform,
+        )
+
+        assert len(dataset) == 2
+        for item in dataset:  # type: ignore[attr-defined]
+            assert item["image"].shape == (3, 32, 32)
+            assert item["mask"].shape == (32, 32)
+            assert item["mask"].min() >= 0
+            assert item["mask"].max() <= 1
```
stegmuel: The masks value are also scaled by 255?
This **[example](https://albumentations.ai/docs/3-basic-usage/semantic-segmentation/) seems to imply that the output masks are still in uint8.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
guarin: Updated to handle >256 classes. It actually worked already. I made it more explicit and added tests.

guarin: > The assumption that we'll have the masks in a directory with the same name and in .png is quite strong, but I guess we'll update based on the needs.

Yes let's see how far we can go with this. The alternative is that we allow users to specify a pattern how to build the mask name from the image name.

tests/_data/test_semantic_segmentation_dataset.py
guarin: No they are not. It was just a coincidence because I set `num_classes=2`. Made it clearer.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
guarin: Yes, this is actually a typo, I meant `data`

tests/_data/test_semantic_segmentation_dataset.py
stegmuel: Great!

PR #159, opened 2025-07-02
Add linear segmentation model.

stegmuel:
## What has changed and why?

Implementation of a linear segmentation model based on DINOv2.

## How has it been tested?

- Tested loaded an exported DINOv2 pre-trained model.
- Tested that the model's output shape is correct.

- [ ] Yes

- [ ] Yes


guarin (approved):
LGTM

PR #160, opened 2025-07-02
Disable FFmpeg install in CI

guarin:
## What has changed and why?

* Disable FFmpeg install in CI

Disabled because it fails too often and is only used for `extract_video_frames` which is not really used. 

- [ ] Yes

- [ ] Yes


liopeer (approved):
LGTM!

PR #161, opened 2025-07-02
Add cross-entropy loss for semantic segmentation

stegmuel:
## What has changed and why?

Added pixel-wise cross entropy loss for semantic segmentation.

## How has it been tested?

Dummy run using the linear segmentation model's output and dummy masks.
Proper tests will follow.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/.github/workflows/test_unit.yml
+++ b/.github/workflows/test_unit.yml
@@ -67,12 +67,13 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python }}
-      - name: Set Up FFmpeg
-        # FFMPEG is not installable on Windows using the setup-ffmpeg action.
-        if: ${{ matrix.runners == 'ubuntu-latest' }}
-        uses: FedericoCarboni/setup-ffmpeg@v3
-        with:
-          ffmpeg-version: 4.3.1
+      # Disabled because FFmpeg install fails too often in CI. See https://github.com/federicocarboni/setup-ffmpeg/issues/29
+      # - name: Set Up FFmpeg
+      #   # FFMPEG is not installable on Windows using the setup-ffmpeg action.
+      #   if: ${{ matrix.runners == 'ubuntu-latest' }}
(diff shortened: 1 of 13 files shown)


src/lightly_train/_task_models/dinov2_semantic_segmentation/dinov2_semantic_segmentation_ce_loss.py
```diff
@@ -0,0 +1,46 @@
...
+        Args:
+            logits: Tensor of shape (B, num_classes, H, W)
+                with raw model logits (before softmax).
+            targets: Tensor of shape (B, H, W) with class indices
+                in the range [0, num_classes - 1].
+
+        Returns:
+            Scalar tensor representing the pixel-wise cross-entropy loss.
+        """
+        return F.cross_entropy(
```
liopeer, line 59: I could possibly see us to use input assertions here to make sure that the right input shapes arrive here at the loss funciton.

src/lightly_train/_task_models/dinov2_semantic_segmentation/dinov2_semantic_segmentation_ce_loss.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class DINOv2SemanticSegmentationCrossEntropyLoss(Module):
+    """
+    Computes pixel-wise cross-entropy loss for semantic segmentation.
+
+    Expects model logits of shape (B, num_classes, H, W) and ground-truth
+    masks of shape (B, H, W) containing integer class indices.
+
+    The loss ignores the specified ignore_index pixels.
```
liopeer: I would be a bit more specific here: AFAIK it ignores the pixel that belong to classes in the `ignore_index`, right?

stegmuel: Yes.

PR #164, opened 2025-07-03
Iterate on train loop

guarin:
## What has changed and why?

* Init dataloader
* Init model
* Minimal loop

## How has it been tested?

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/common_helpers.py
+++ b/src/lightly_train/_commands/common_helpers.py
@@ -17,7 +17,7 @@ import time
 import warnings
 from enum import Enum
 from pathlib import Path
-from typing import Any, Generator, Iterable, Literal, Sequence
+from typing import Any, Generator, Iterable, Literal, Sequence, Sized, TypeVar
 
 import torch
 from pytorch_lightning.accelerators.accelerator import Accelerator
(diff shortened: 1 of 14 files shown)


src/lightly_train/_commands/common_helpers.py
```diff
@@ -542,3 +542,63 @@ def _unlink_and_ignore(path: Path) -> None:
         path.unlink(missing_ok=True)
     except OSError:
         pass
+
+
+_T = TypeVar("_T")
+
+
+def get_global_batch_size(
+    global_batch_size: int,
```
guarin, line 551: Moved from `train_helpers.py` to `common_helpers.py`

src/lightly_train/_commands/train_helpers.py
```diff
@@ -93,63 +93,6 @@ def get_total_num_devices(
     return total_num_devices
 
 
-def get_global_batch_size(
-    global_batch_size: int,
```
guarin, line 97: Moved to `common_helpers.py`

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -96,5 +127,88 @@ def get_dataset(
...
+
+
+def get_task_train_model_args(
+    task_args: dict[str, Any] | TaskTrainModelArgs | None,
+) -> TaskTrainModelArgs:
+    if isinstance(task_args, TaskTrainModelArgs):
+        return task_args
+    task_args = {} if task_args is None else task_args
+    task_cls = DINOv2SemanticSegmentationTrainArgs
+    args = validate.pydantic_model_validate(task_cls, task_args)
```
guarin, line 206: Will have to make this more flexible once we have more tasks

src/lightly_train/_data/infinite_cycle_iterator.py
```diff
@@ -0,0 +1,35 @@
...
+    def __iter__(self) -> Self:
+        return self
+
+    def __next__(self) -> _T:
+        if self._iter is None:
+            self._iter = iter(self.iterable)
+        try:
+            return next(self._iter)
+        except StopIteration:
+            self._iter = iter(self.iterable)
```
stegmuel, line 33: Is it correct to assume that fabric takes care of doing `self.iterable.sampler.set_epoch(epoch)` ?

stegmuel: Can we have a TODO to replace this with an infinite sampler _à la_ DINOv2 as this approach still has the same overhead as the epoch-based approach.

src/lightly_train/_commands/train_task.py
```diff
@@ -96,28 +108,83 @@ def train_task_from_config(config: TrainTaskConfig) -> None:
...
 
 
 class TrainTaskConfig(PydanticConfig):
     out: PathLike
     data: MaskSemanticSegmentationDataArgs
     model: str
     task: Literal["semantic_segmentation"]
+    steps: int | Literal["auto"] = "auto"
+    batch_size: int | Literal["auto"] = "auto"
+    num_workers: int | Literal["auto"] = "auto"
```
stegmuel, line 193: isn't this an attribute of `loader_args`?

src/lightly_train/_commands/train_task.py
```diff
@@ -96,28 +108,83 @@ def train_task_from_config(config: TrainTaskConfig) -> None:
...
     logger.info(
-        f"Resolved Args: {train_task_helpers.pretty_format_args(args=config.model_dump())}"
+        f"Resolved Args: {helpers.pretty_format_args(args=config.model_dump())}"
     )
+    logger.info(f"Starting training for {config.steps} steps...")
+    for step in range(config.steps):
+        batch = next(infinite_train_dataloader)
+        # TODO(Guarin, 07/25): Backprop.
+        # TODO(Guarin, 07/25): Log loss and metrics.
+        model.training_step(fabric=fabric, batch=batch)
```
stegmuel, line 174: `loss.backward()` and `optimizer.step()` are also implemented inside `model.trainig_step`, right?

stegmuel (approved):
Approved with some questions.

src/lightly_train/_data/infinite_cycle_iterator.py
guarin: > Is it correct to assume that fabric takes care of doing self.iterable.sampler.set_epoch(epoch) ?

Yes this happens here:  https://github.com/Lightning-AI/pytorch-lightning/blob/e29a8cbb83b9041ff7b7d10f24d9274f39d2358d/src/lightning/fabric/wrappers.py#L320-L321

> Can we have a TODO to replace this with an infinite sampler à la DINOv2 as this approach still has the same overhead as the epoch-based approach.

Added

src/lightly_train/_commands/train_task.py
guarin: It is the same as in `train.py`. `batch_size` and `num_workers` are important and often changed by users so we expose them directly. Rarely used arguments like `persistent_workers` are not directly exposed but can be set trough `loader_args`.

src/lightly_train/_commands/train_task.py
guarin: I would have put it outside together with `optimizer.zero_grad`, gradient clipping, gradient accumulation, etc. Otherwise we have to reimplement it in every model.

stegmuel: So do I, but I was confused to not see it.

guarin: Couldn't implement it yet because the forward pass is not implemented yet :)

PR #166, opened 2025-07-04
Update Documentation of RT-DETR Models

liopeer:
## What has changed and why?
- We are actually only supporting the RT-DETRv2 package, however this can also be used with the RT-DETRv1 configs.
- This PR streamlines the naming as follows: The package is always called RT-DETR, the model is always called RT-DETRv2
- I additionally included all the v2 configs that had previously not been in the documentation
- I excluded all the `hgnetv2` models, e.g. https://github.com/lyuwenyu/RT-DETR/blob/main/rtdetrv2_pytorch/configs/rtdetrv2/rtdetrv2_hgnetv2_h_6x_coco.yml, since they don't run with LightlyTrain.

@yutong-xiang-97 update: makes explicit that users can train v1 models with v2 package.

## How has it been tested?
- manually ran all the supported models once locally
- I would not implement unit tests for all the models, this would take forever

- [x] Yes

- [x] Yes

### DIFF
--- a/docs/source/index.md
+++ b/docs/source/index.md
@@ -48,7 +48,7 @@ a wide range of model architectures and use cases out of the box.
 :alt: benchmark results
 
 On COCO, YOLOv8-s models pretrained with LightlyTrain achieve high performance across all tested label fractions.
-These improvements hold for other architectures like YOLOv11, RT-DETRv2, and Faster R-CNN.
+These improvements hold for other architectures like YOLOv11, RT-DETR, and Faster R-CNN.
 See our [announcement post](https://www.lightly.ai/blog/introducing-lightly-train) for more details.
 ```
 
--- a/docs/source/index.md
+++ b/docs/source/index.md
@@ -122,7 +122,7 @@ Want to use your model to generate image embeddings instead? Check out the {ref}
 | Torchvision | ResNet, ConvNext, ShuffleNetV2 | [🔗](#models-torchvision) |
 | TIMM | All models | [🔗](#models-timm) |
 | Ultralytics | YOLOv5, YOLOv6, YOLOv8, YOLO11, YOLO12 | [🔗](#models-ultralytics) |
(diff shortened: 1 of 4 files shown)


yutong-xiang-97 (commented):
As someone who has little context: will I be able to train a v1 model with the v2 package and v1 config since you mentioned that the latter can be used? What model do I get if I do so? Imho we should also make this clear.

docs/source/index.md
```diff
@@ -48,7 +48,7 @@ a wide range of model architectures and use cases out of the box.
 :alt: benchmark results
 
 On COCO, YOLOv8-s models pretrained with LightlyTrain achieve high performance across all tested label fractions.
-These improvements hold for other architectures like YOLOv11, RT-DETR, and Faster R-CNN.
+These improvements hold for other architectures like YOLOv11, RT-DETRv2, and Faster R-CNN.
```
stegmuel: I would keep the previous version.
We tested RT-DETR, but not RT-DETRv2 I believe.

docs/source/models/rtdetr.md
```diff
@@ -8,7 +8,12 @@ with LightlyTrain.
 ```{note}
 RT-DETR is not a pip-installable Python package. For this reason,
 RT-DETR is not fully integrated with LightlyTrain and has to be
-installed manually.
+installed manually, however LightlyTrain provides a model wrapper for the RT-DETR models.
+```
+
+```{warning}
+Due to an incompatibility between the `torchvision` version requirements of RT-DETRv1 and
```
stegmuel: I find this sentence misleading.
Can we re-phrase to something like:
"Due to an incompatibility between the `torchvision` version requirements of RT-DETRv1, we support the RT-DETRv2 implementation that is backward compatible and allows the training of both RT-DETR(v1) and RT-DETRv2 models."

stegmuel (commented):
LGTM.
Left some comments.

PR #168, opened 2025-07-08
Set up ONNX test

yutong-xiang-97:
## What has changed and why?

Set up a test script that converts `DINOv2SemanticSegmentation` model to ONNX by

- exporting DINOv2 ViT backbones by `lightly_train.train()` (done by a separate script),
- loading a backbone to the `DINOv2SemanticSegmentation` model,
- exporting the segmentation model to ONNX format,
- loading the ONNX model with a sanity check,
- running one inference pass with `onnxruntime`, and
- checking that output is the same as the torch segmentation model.

Since we don't have the export function implemented yet, the test script is not part of the unit test yet, nor do we include onnx as a dependency. It will be worked on later soon.

## How has it been tested?

Tested for all ViT models.

- [ ] Yes

- [ ] Yes


guarin (commented):
LGTM! Nice that it works out of the box. I tested it out and had to reduce tolerance to 
```
    atol=1e-5,  # absolute tolerance
    rtol=1.0,
```

To make the tests reliably pass. This was with the additional change to set dynamic axes in the export:
```
torch.onnx.export(
    model,
    (dummy_input,),
    onnx_path,
    input_names=["input"],
    output_names=["output"],
    dynamic_axes={
        "input": {0: "batch_size", 2: "height", 3: "width"},
        "output": {0: "batch_size", 2: "height", 3: "width"},
    },
)
```

PR #170, opened 2025-07-09
Disable DeviceStatsMonitor by default

yutong-xiang-97:
## What has changed and why?

DeviceStatsMonitor is disabled by default. Can be enabled by

```py
lightly_train.train(
    ...,
    callbacks={
        "device_stats_monitor": {},
    },
)
```

also adjusted the unit tests accordingly.

## How has it been tested?

- Manually tested on local machine.
- Unit tests

- [x] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -9,10 +9,9 @@ from __future__ import annotations
 
 import json
 import logging
-from contextlib import contextmanager
 from json import JSONEncoder
 from pathlib import Path
-from typing import Any, Generator, Literal
+from typing import Any, Literal
 
 from lightning_fabric import Fabric
 from lightning_fabric import utilities as fabric_utilities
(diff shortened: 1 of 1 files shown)


src/lightly_train/_callbacks/callback_args.py
```diff
@@ -32,11 +32,10 @@ class CallbackArgs(PydanticConfig):
     learning_rate_monitor: LearningRateMonitorArgs | None = Field(
         default_factory=LearningRateMonitorArgs
     )
-    device_stats_monitor: DeviceStatsMonitorArgs | None = Field(
-        default_factory=DeviceStatsMonitorArgs
-    )
```
guarin, line 37: Why did you move it further down?

tests/_callbacks/test_callback_helpers.py
```diff
@@ -111,6 +108,25 @@ def test_get_callbacks__mlflow(tmp_path: Path) -> None:
...
+        callback_args=callback_args,
+        out=tmp_path,
+        wrapped_model=model,
+        embedding_model=embedding_model,
+        normalize_args=NormalizeArgs(),
+        loggers=[],
+    )
+    assert len(callbacks) == 6
+    assert any(isinstance(c, ModelCheckpoint) for c in callbacks)
+    assert not any(isinstance(c, MLFlowLogging) for c in callbacks)
```
guarin: Shouldn't this check that there is a devicestatsmonitor callback instead of checking `ModelCheckpoint` and `MLFlowLogging`?

src/lightly_train/_callbacks/callback_args.py
yutong-xiang-97: It is default to None which is different from other args? Anyway I moved it back to the original place.

PR #171, opened 2025-07-09
Task checkpoint and backbone handling

guarin:
## What has changed and why?

* Save/resume checkpoint
* Handle backbone loading and freezing

## How has it been tested?

* Manually

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/train_task.py
+++ b/src/lightly_train/_commands/train_task.py
@@ -28,6 +28,7 @@ from lightly_train._data.mask_semantic_segmentation_dataset import (
     MaskSemanticSegmentationDataArgs,
 )
 from lightly_train._loggers.task_logger_args import TaskLoggerArgs
+from lightly_train._task_checkpoint import TaskCheckpointArgs
 from lightly_train._task_models.task_train_model import TaskTrainModelArgs
 from lightly_train.types import PathLike
 
@@ -55,6 +56,7 @@ def train_task(
     logger_args: dict[str, Any] | None = None,
     task_args: dict[str, Any] | None = None,
     loader_args: dict[str, Any] | None = None,
+    checkpoint_args: dict[str, Any] | None = None,
 ) -> None:
     config = validate.pydantic_model_validate(TrainTaskConfig, locals())
     train_task_from_config(config=config)
@@ -110,6 +112,9 @@ def train_task_from_config(config: TrainTaskConfig) -> None:
         )
     )
     config.task_args = helpers.get_task_train_model_args(task_args=config.task_args)
+    config.checkpoint_args = helpers.get_checkpoint_args(
+        checkpoint_args=config.checkpoint_args
+    )
 
     # TODO(Guarin, 07/25): Verify out_dir same on all local ranks, see train.py. We can simplify this
     # here as distributed processing is already initialized with fabric.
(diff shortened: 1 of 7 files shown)


src/lightly_train/_task_models/dinov2_semantic_segmentation/dinov2_semantic_segmentation_train.py
```diff
@@ -98,3 +107,8 @@ def validation_step(
     def get_optimizer(self) -> Optimizer:
         # TODO(Guarin, 07/25): Handle weight decay for norm and bias parameters.
         return AdamW(self.parameters())
+
+    def set_train_mode(self) -> None:
+        self.train()
+        if self.task_args.freeze_backbone:
+            self.model.freeze_backbone()
```
stegmuel, line 114: Isn't this a bit overkill?
Why not just `self.model.backbone.eval()` ?

src/lightly_train/_commands/train_task.py
```diff
@@ -171,17 +174,46 @@ def train_task_from_config(config: TrainTaskConfig) -> None:
...
+
+    # TODO(Guarin, 07/25): Replace with infinite batch sampler instead to avoid
+    # reloading dataloader after every epoch? Is this preferred over persistent workers?
+    infinite_train_dataloader = InfiniteCycleIterator(
+        iterable=state["train_dataloader"]
+    )
+
+    for name, param in model.named_parameters():
+        logger.debug(f"grad={param.requires_grad} {name}")
+    for name, module in model.named_modules():
```
stegmuel, line 205: Honest question: Why do you print that right before changing the training flag?

src/lightly_train/_commands/train_task.py
```diff
@@ -171,17 +174,46 @@ def train_task_from_config(config: TrainTaskConfig) -> None:
...
+        logger.debug(f"train={module.training} {name}")
+
+    model.set_train_mode()
```
stegmuel: Is this second call to `model.set_train_mode()` needed?

stegmuel (approved):
Approved with comments/questions.

src/lightly_train/_task_models/dinov2_semantic_segmentation/dinov2_semantic_segmentation_train.py
guarin: You mean overkill because I also set `requires_grad=False`?

src/lightly_train/_commands/train_task.py
guarin: Good point, changed it.

src/lightly_train/_commands/train_task.py
guarin: I added it because I wanted to be 100% sure the model is in the right state before training starts. I removed it now and added some asserts in the checkpoint loading to make sure it doesn't mess up any of the objects.

src/lightly_train/_task_models/dinov2_semantic_segmentation/dinov2_semantic_segmentation_train.py
stegmuel: Yes.

PR #175, opened 2025-07-11
VFMF Add MLflow Logger

yutong-xiang-97:
## What has changed and why?

Use Lightning's MLflow logger for logging.

Note that:

1. The resolution of the accelerator and strategy does not work for now
2. We didn't add the corresponding unit test because the settings will be too cumbersome.

## How has it been tested?

run `test_train_task` with a script using the MLflow service.

- [ ] Yes

- [ ] Yes


guarin (approved):
LGTM!

PR #179, opened 2025-07-15
Thomas trn 637 dinov2 iteration based hp

stegmuel:
## What has changed and why?

Replacing epoch-based hyperparameters with step-based ones.

## How has it been tested?

Will be tested with a dummy training run.

- [x] Yes

- [x] Yes
- [] Not needed (internal change without effects for user)

### DIFF
--- a/docs/source/methods/dinov2.md
+++ b/docs/source/methods/dinov2.md
@@ -32,7 +32,7 @@ if __name__ == "__main__":
         data="my_data_dir",
         model="dinov2_vit/vitb14_pretrain",
         method="dinov2",
-        epochs=100, # We recommend epochs = -(-125000 * batch size // dataset size)
+        epochs=100, # We recommend epochs = 125000 * batch size // dataset size
     )
 ````
 
--- a/docs/source/methods/dinov2.md
+++ b/docs/source/methods/dinov2.md
@@ -20,11 +20,6 @@ if __name__ == "__main__":
         data="my_data_dir",
         model="dinov2_vit/vitb14_pretrain",
         method="dinov2",
-        method_args={
-            # Only set these arguments when starting from a pretrained model
-            "student_freeze_backbone_epochs": 1,  # Freeze the student backbone for 1 epoch
-            "student_freeze_last_layer_epochs": 0,  # Unfreeze the student last layer
-        },
     )
 ````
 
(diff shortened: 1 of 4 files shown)


src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -80,13 +81,13 @@ class DINOv2Args(MethodArgs):
...
     batch_norm: bool = False
-    student_freeze_last_layer_epochs: int = 1
+    student_freeze_last_layer_steps: int = 1250
 
     # freeze student backbone
     # Useful when starting from DINOv2 pretrained weights. This allows the projection
     # head to be trained while the backbone is frozen. This is important because the
     # DINOv2 pretrained weights do not include the projection head.
-    student_freeze_backbone_epochs: int = 0
+    student_freeze_backbone_steps: int = 0
```
guarin, line 90: Should we update the comment above?

src/lightly_train/_methods/dinov2/dinov2.py
```diff
@@ -657,3 +658,30 @@ def on_train_batch_end(
...
+
+    @rank_zero_only  # type: ignore[misc]
+    def warn_if_steps_too_low(self) -> None:
+        # Check if dataset size is set.
+        assert self.method_args.dataset_size is not None, (
+            "dataset_size must be set before training."
+        )
+
+        # Compute dataset-specific epoch recommendation.
+        steps_per_epoch = self.method_args.dataset_size // self.global_batch_size
```
guarin: I think you don't need to save the dataset size. You can calculate it with:
```
dataset_size = len(self.train_dataloader.dataset)
```

guarin (commented):
Two comments otherwise LGTM!

docs/source/methods/dinov2.md
```diff
@@ -32,11 +32,6 @@ if __name__ == "__main__":
         data="my_data_dir",
         model="dinov2_vit/vitb14_pretrain",
         method="dinov2",
-        method_args={
-            # Only set these arguments when starting from a pretrained model
-            "student_freeze_backbone_epochs": 1,  # Freeze the student backbone for 1 epoch
-            "student_freeze_last_layer_epochs": 0,  # Unfreeze the student last layer
-        },
```
guarin, line 39: Should we add here a `epochs=100      # We recommend epochs = 125000 * batch size / dataset size`

As a follow-up we should set `epoch = "auto"` by default and set it automatically for DINOv2.

src/lightly_train/_methods/dinov2/dinov2.py
stegmuel: IMO the comment is still valid, but we just don't recommend that setting as we've had success without that.

docs/source/methods/dinov2.md
```diff
@@ -32,11 +32,7 @@ if __name__ == "__main__":
...
+        epochs=100, # We recommend epochs = -(-125000 * batch size // dataset size)
```
guarin:
```suggestion
        epochs=100, # We recommend epochs = 125000 * batch size // dataset size
```

guarin: Don't think double negation is needed. If it is negative by accident users will get an error anyways.

PR #181, opened 2025-07-15
Add ONNX Export Function

yutong-xiang-97:
## What has changed and why?

NOTE: this PR is only a draft to showcase the idea on how to realize it. Please only focus on the functionality and ignore the implementation details.

Add `export_onnx` function and corresponding tests.

Design points:

- the `export_onnx` function is exposed to users and designed similarly to the `export` function in `lightly_train`
- we don't do `export_onnx` automatically after fine-tuning. Not every user would require ONNX support.
- More specifically, `export_onnx` loads the Lightning Fabric ckpt saved during fine-tuning and converts it to an ONNX model.
- I didn't include the unit test for numerical identicalness, cuz 1. it has already been tested at #168 and 2. the test would be basically a re-implementation of the method which losts the purpose of testing.

Things to add and decide.

- I would add some info in the exported Lightning checkpoints for loading them conveniently:
  - model_name
  - num_classes
  and also make the keys in the `state_dict` match so that we use `strict=True`
- Only export on rank zero in Fabric, otherwise there would be issues.
- If needed, the ONNX input and output names can also be part of the `ExportONNXConfig`

## How has it been tested?

N/A

- [ ] Yes

- [ ] Yes


tests/_commands/test_export_onnx.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+def test_onnx_model_consistency_check(onnx_model_path: Path) -> None:
+    """Test that the exported ONNX model passes ONNX validation."""
+    onnx_model: onnx.ModelProto = onnx.load(str(onnx_model_path))
+
+    try:
+        onnx.checker.check_model(onnx_model, full_check=True)
+    except Exception as e:
+        pytest.fail(f"ONNX model validation failed: {e}")
```
guarin: Lets move that code into the main `test_export_onnx` test.

tests/_commands/test_export_onnx.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+def test_shape_inference_runs(onnx_model_path: Path) -> None:
+    """Test that shape inference runs without errors."""
+    model: onnx.ModelProto = onnx.load(str(onnx_model_path))
+
```
guarin: Would drop this test

tests/_commands/test_export_onnx.py
```diff
@@ -0,0 +1,157 @@
...
+    # Verify input has shape information
+    input_shape = inferred.graph.input[0].type.tensor_type.shape
+    assert input_shape.dim, "Input should have shape dimensions"
+
+
+@pytest.mark.parametrize("batch_size", [1, 4, 8])
+def test_export_dynamic_batch(onnx_model_path: Path, batch_size: int) -> None:
+    """Test that the exported model works with different batch sizes."""
+    # Create input with the specified batch size
```
guarin: Would drop this test. Instead the main `test_export_onnx` test should test the resulting model with a dummy input.

src/lightly_train/_commands/export_onnx.py
```diff
@@ -0,0 +1,87 @@
...
+from lightly_train._configs.config import PydanticConfig
+from lightly_train._task_models.dinov2_semantic_segmentation.dinov2_semantic_segmentation import (
+    DINOv2SemanticSegmentation,
+)
+from lightly_train.types import PathLike
+
+logger = logging.getLogger(__name__)
+
+
+def export_onnx(
```
guarin: Let's call this `export_task` for now to allow different formats.

```suggestion
def export_task(
```

src/lightly_train/_commands/export_onnx.py
```diff
@@ -0,0 +1,87 @@
...
+    out_path = common_helpers.get_out_path(out=config.out, overwrite=config.overwrite)
+    checkpoint_path = common_helpers.get_checkpoint_path(checkpoint=config.checkpoint)
+
+    # Load the model
+    torch.use_deterministic_algorithms(True)
+    checkpoint = torch.load(checkpoint_path)
+    model = DINOv2SemanticSegmentation(
+        model_name="vits14",  # checkpoint["model_name"],
+        num_classes=2,  # checkpoint["num_classes"],
+    )
```
guarin: Agree with you that we should store model information in checkpoint.

src/lightly_train/_commands/export_onnx.py
guarin: I think later on we'll need the following arguments:
```
    *,
    out: PathLike,
    checkpoint: PathLike,
    format: Literal["onnx"],
    batch_size: int = 1,
    num_channels: int = 3,
    height: int = 224,
    width: int = 224,
    device: str = "auto",
    overwrite: bool = False,
    format_args: dict[str, Any] | None = None     # Format specific arguments. Eg. "dynamic" for onnx and int8 precision for tensorrt. 
```

src/lightly_train/_commands/export_onnx.py
```diff
@@ -0,0 +1,87 @@
...
+            Path where the exported model will be saved.
+        overwrite:
+            Overwrite the output file if it already exists.
+    """
+    config = ExportONNXConfig(**locals())
+    export_onnx_from_config(config=config)
+
+
+def export_onnx_from_config(config: ExportONNXConfig) -> None:
+    # Only export on rank 0.
```
guarin:
```suggestion
    # Only export on rank 0.
    if distributed.is_initialized() and distributed.get_rank() > 0:
        return
```

tests/_commands/test_export_onnx.py
yutong-xiang-97: I would instead rename it as `test_inference_with_onnxruntime` so that we can also add tests for other inference engines like TensorRT in the future.

src/lightly_train/_commands/export_onnx.py
yutong-xiang-97: I don't think we need the `device` argument but the rest is necessary.

PR #182, opened 2025-07-17
Allow training on mixed number of channels

guarin:
## What has changed and why?

* Allow training on mixed number of channels

This allows for example training on 3- and 4-channel images at the same time. The ChannelDrop augmentation now ignores images that already have the correct number of channels. 

## How has it been tested?

* Unit tests
* Manually tested on mixed dataset with 3- and 4-channel images. 

- [x] Yes

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,12 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Allow loading pretrained DINOv2 teacher weights for distillation methods with an extra
+  `teacher_weights` argument in `method_args`.
+- Support images with different number of channels in the channel drop transform.
 - Documented support for the [RT-DETRv2 models](https://docs.lightly.ai/train/stable/models/rtdetr.html).
 
 ### Changed
 
 - The callback `DeviceStatsMonitor` is now disabled by default.
-- Allow loading pretrained DINOv2 teacher weights for distillation methods with an extra `teacher_weights` argument in `method_args`.
 
 ### Deprecated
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -36,14 +38,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
-### Changed
-
-### Deprecated
-
-### Removed
-
-### Fixed
-
 ### Security
```
yutong-xiang-97, line 41: This was the artifact created by me during the last release. Thanks for fixing!

The "Security" section should be removed, too but I will do it during the next release.

src/lightly_train/_transforms/channel_drop.py
```diff
@@ -70,15 +70,20 @@ def apply(
...
         if len(self.weight_drop) != num_channels:
             raise RuntimeError(
                 f"Length of weight_drop ({len(self.weight_drop)}) must match "
                 f"number of image channels ({num_channels})"
             )
 
-        if self.num_channels_keep >= num_channels:
-            return img
-
         channels_to_drop = np.random.choice(
```
yutong-xiang-97, line 87: Not relevant to the main point of the PR, but do we have to calculate `channels_to_drop` first? Imho. simply
```python
prob_keep = 1.0 - self._prob_drop
prob_keep /= prob_keep.sum()
channels_to_keep = np.sort(
    np.random.choice(
        num_channels,
        size=self.num_channels_keep,
        replace=False,
        p=prob_keep,
    )
)
```
would do the work.

yutong-xiang-97 (approved):
Approved with some comments.

src/lightly_train/_transforms/channel_drop.py
guarin: This doesn't work because `np.random.choice` with `p=prob_keep` could remove channels that have `prob_drop=0`. I believe I first tried to write it this way but ran into issues.

PR #185, opened 2025-07-17
Fix Unit Tests

yutong-xiang-97:
## What has changed and why?

Fix unit test:

- remove `torch.use_deterministic_algorithms(True)` in `export_task` to fix failures
- add `optim.step()` in `test_dinov2::test_layerwise_decay_optimizer` to fix the warning: 

```
UserWarning: Detected call of `lr_scheduler.step()` before `optimizer.step()`. In PyTorch 1.1.0 and later, you should call them in the opposite order: `optimizer.step()` before `lr_scheduler.step()`.  Failure to do this will result in PyTorch skipping the first value of the learning rate schedule. 
```

## How has it been tested?

Successful checks.

- [ ] Yes

- [ ] Yes


guarin (approved):
LGTM ✅

PR #189, opened 2025-07-22
Thomas trn 625 allow arbitrary input size

stegmuel:
## What has changed and why?

Update the DINOv2 ViT to support square images with resolution that is not a multiple of the patch size.
Update the plotting of augmentations for odd size image resolutions

## How has it been tested?

Dummy runs with various resolutions (square only).

- [x] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_models/dinov2_vit/dinov2_vit_src/models/vision_transformer.py
+++ b/src/lightly_train/_models/dinov2_vit/dinov2_vit_src/models/vision_transformer.py
@@ -258,8 +258,9 @@ class DinoVisionTransformer(nn.Module):
         )
 
     def prepare_tokens_with_masks(self, x, masks=None):
-        B, nc, w, h = x.shape
-        x = self.patch_embed(x)
+        # TODO(Thomas, 07/25): Fix swapped h and w.
+        B, nc, _, _ = x.shape
+        x, w, h = self.patch_embed(x)
         if masks is not None:
             x = torch.where(
                 masks.unsqueeze(-1), self.mask_token.to(x.dtype).unsqueeze(0), x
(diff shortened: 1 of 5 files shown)


src/lightly_train/_models/dinov2_vit/dinov2_vit_src/models/vision_transformer.py
```diff
@@ -258,8 +258,9 @@ def interpolate_pos_encoding(self, x, w, h):
         )
 
     def prepare_tokens_with_masks(self, x, masks=None):
-        B, nc, w, h = x.shape
-        x = self.patch_embed(x)
+        # TODO(Thomas, 07/25): Fix swapped h and w.
+        B, nc, _, _ = x.shape
+        x, w, h = self.patch_embed(x)
```
guarin, line 263: Why is this swapped? Patch embed returns `x, new_H, new_W`

guarin (commented):
Left one question otherwise LGMT

src/lightly_train/_models/dinov2_vit/dinov2_vit_src/models/vision_transformer.py
stegmuel: The subsequent code used swapped h and w, so I swapped for consistency.
This is inherited from DINOv2's [code ](https://github.com/facebookresearch/dinov2/issues/504).

guarin: Ok, best to add this context directly in the comments in the code. Otherwise it is a bit confusing why we would use swapped h/w :)

PR #193, opened 2025-07-23
Add ignore_classes arg to the dateset args

stegmuel:
## What has changed and why?

Add `ignore_classes` field to the data args.

## How has it been tested?

-

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_data/mask_semantic_segmentation_dataset.py
+++ b/src/lightly_train/_data/mask_semantic_segmentation_dataset.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 
 from collections.abc import Iterable, Sequence
 from pathlib import Path
+from typing import Optional
 
 from torch.utils.data import Dataset
 
@@ -57,6 +58,7 @@ class MaskSemanticSegmentationDatasetArgs(PydanticConfig):
     image_dir: Path
     mask_dir: Path
     classes: dict[int, str] | None = None
+    ignore_classes: set[int] | None = None
 
     # NOTE(Guarin, 07/25): The interface with below methods is experimental. Not yet
     # sure if it makes sense to have this in dataset args.
@@ -81,6 +83,7 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
     train: SplitArgs
     val: SplitArgs
     classes: dict[int, str]
+    ignore_classes: Optional[set[int]] = None
 
     # NOTE(Guarin, 07/25): The interface with below methods is experimental. Not yet
     # sure if this makes sense to have in data args.
@@ -89,6 +92,7 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
             image_dir=Path(self.train.images),
             mask_dir=Path(self.train.masks),
             classes=self.classes,
+            ignore_classes=self.ignore_classes,
         )
 
     def get_val_args(self) -> MaskSemanticSegmentationDatasetArgs:
@@ -96,4 +100,5 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
             image_dir=Path(self.val.images),
             mask_dir=Path(self.val.masks),
             classes=self.classes,
+            ignore_classes=self.ignore_classes,
         )


src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -197,6 +202,16 @@ def get_dataset(
     )
 
 
+def collate_fn(batch: list[dict[str, Any]]) -> dict[str, Any]:
+    # TODO(Thomas, 07/25): make this task/split dependant.
+    return {
```
guarin, line 207: Out of curiosity, why can we not use the default collate function?

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -22,6 +28,8 @@
     PathLike,
 )
 
+logger = logging.getLogger(__name__)
```
guarin: Please don't add logging in dataset code (or anything that runs in a dataloader). The problem with logging in dataset code is that it leads to deadlocks. The reason this happens is that dataloader processes are forked from the main process and the logging module is not multiprocessing safe. If you are unlucky the logging module has acquired a lock at the same time that a new dataloader process is started. If this happens then the lock is copied in locked state to the new process and is never released again... Hard learning we made a while ago :)

The alternative to logging is to print the message. What is even better is to log the information from outside the dataset if possible.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -33,30 +41,153 @@ def __init__(
...
+
+        # Uniform masks are discarded.
+        if len(unique_values) == 1:
+            return False
+
+        # Check if at least one value in the mask is in the valid classes.
+        return bool(np.isin(unique_values, self.valid_classes).any())
+
+    def filter_empty_targets(self) -> None:
+        # Instantiate new list of file names
```
guarin: Add todo to move this somewhere else as it won't work with memory mapped filenames. Typically we would want to run this with multiple processes as it loads every mask in the dataset.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -33,30 +41,153 @@ def __init__(
...
+        for filename in self.image_filenames:
+            filepath = (self.args.mask_dir / filename).with_suffix(".png")
+
+            mask = file_helpers.open_image(image_path=filepath, mode="MASK")
+            if self.is_mask_valid(mask):
+                new_image_filenames.append(filename)
+
+        # Display the number of filtered files.
+        n_filtered_files = len(self) - len(new_image_filenames)
+        logger.info(f"Filtered {n_filtered_files} invalid masks out of {len(self)}.")
```
guarin: Remove `logger.info` see comment above

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -33,30 +41,153 @@ def __init__(
...
+            "mask": transformed_mask,
+            "target": target,
         }
 
 
 class MaskSemanticSegmentationDatasetArgs(PydanticConfig):
     image_dir: Path
     mask_dir: Path
     classes: dict[int, str] | None = None
+    ignore_classes: set[int] = set()
```
guarin:
```suggestion
    ignore_classes: set[int] | None  = None
```
Don't use empty sets/dicts/lists as default values, as they could be accidentally shared between multiple instances of the object. Pydantic is actually special and handles this behind the scenes (it creates a new set for every instance) but I would prefer to keep `None` for consistency. It also allows us to differentiate between the default case (`None`) and the case where a user passed a set (even if it is empty). Sometimes this is very useful, especially for error messages.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -33,30 +41,153 @@ def __init__(
...
+        targets = {
+            "masks": torch.stack(img_masks),
+            "labels": mask.new_tensor(img_labels, dtype=torch.long),
+        }
+        return targets
+
+    def remap_mask(self, mask: torch.Tensor) -> torch.Tensor:
+        # Create a lookup table initialized with ignore_index
+        max_class = int(mask.max().item())
+        lut = torch.full((max_class + 1,), self.ignore_index, dtype=torch.long)
```
guarin:
```suggestion
        lut = mask.new_full((max_class + 1,), self.ignore_index, dtype=torch.long)
```

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -33,30 +41,153 @@ def __init__(
...
+        # Verify that the mask and the image have the same shape.
+        assert image.shape[:2] == mask.shape, (
+            f"Shape mismatch: image shape is {image.shape[:2]} while mask shape is {mask.shape}."
+        )
+
+        # Re-do the augmentation until the mask is valid.
+        mask_is_valid = False
+        while not mask_is_valid:
+            transformed = self.transform({"image": image, "mask": mask})
+            mask_is_valid = self.is_mask_valid(transformed["mask"].numpy())
```
guarin: Let's iterate a limited number of times to avoid infinitely blocking the dataloader.
```suggestion
        mask_is_valid = False
       for _ in range(20):
            transformed = self.transform({"image": image, "mask": mask})
            mask_is_valid = self.is_mask_valid(transformed["mask"].numpy())
            if mask_is_valid:
                break
       if not mask_is_valid:
           raise RuntimeError("<descriptive message>")
```

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -33,30 +41,153 @@ def __init__(
...
+
+        # Get binary masks.
+        target = self.get_binary_masks(transformed["mask"])
+
+        # Mark pixels to ignore in the masks.
+        transformed_mask = transformed["mask"]
+        transformed_mask = self.remap_mask(transformed_mask)
```
guarin: Add TODO to only do the binary transformation if needed. Also only do the mask remapping if needed. I can do that when I add the linear semseg model.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -81,19 +212,52 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
     train: SplitArgs
     val: SplitArgs
     classes: dict[int, str]
+    ignore_classes: set[int] = set()
+    check_empty_targets: bool = True
+
+    # Ignore index and number of classes can't be set by the user.
+    ignore_index: int = Field(default=-100, exclude=True)
+    num_classes: int = Field(default=-1, exclude=True)
```
guarin: Best to transform attributes that can be inferred from other attributes into properties. Otherwise there is always a risk that they get out of sync.

```suggestion
    @property
    def num_included_classes(self) -> int:
        return len(self.classes.keys() - self.ignore_classes)
```

Would also rename to `num_included_classes` to make it clear that this considers the ignored classes.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -81,19 +212,52 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
...
+
+    @model_validator(mode="after")
+    def compute_internal_fields(self) -> MaskSemanticSegmentationDataArgs:
+        """Automatically set ignore_index and num_classes."""
```
guarin: Can be removed if you use properties above

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -81,19 +212,52 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
     train: SplitArgs
     val: SplitArgs
     classes: dict[int, str]
+    ignore_classes: set[int] = set()
+    check_empty_targets: bool = True
+
+    # Ignore index and number of classes can't be set by the user.
+    ignore_index: int = Field(default=-100, exclude=True)
```
guarin: What about this? Then in most cases it is -100 except if for some reason the dataset has classes that go below -100.

```suggestion
    @property
    def ignore_index(self) -> int:
        return min(-100, min(self.classes) - 1)
```

src/lightly_train/_task_models/dinov2_semantic_segmentation/dinov2_semantic_segmentation_train.py
```diff
@@ -95,7 +92,7 @@ def __init__(
             # We probably don't want to instantiate the model here. Either we pass it
             # from the outside or we use a setup function (might be useful for FSDP).
             model_name=model_name,
-            num_classes=len(data_args.classes),
+            num_classes=data_args.num_classes + 1,
```
guarin: This looks very suspicious! Why is it +1?

src/lightly_train/_task_models/dinov2_semantic_segmentation/dinov2_semantic_segmentation_train.py
```diff
@@ -111,7 +108,7 @@ def __init__(
             mask_coefficient=task_args.loss_mask_coefficient,
             dice_coefficient=task_args.loss_dice_coefficient,
             class_coefficient=task_args.loss_class_coefficient,
-            num_labels=len(data_args.classes),
+            num_labels=data_args.num_classes + 1,
             no_object_coefficient=task_args.loss_no_object_coefficient,
```
guarin: Same as above

src/lightly_train/_task_models/dinov2_semantic_segmentation/dinov2_semantic_segmentation_train.py
```diff
@@ -120,19 +117,19 @@ def __init__(
         # TODO(Guarin, 07/25): Make params configurable.
         self.train_miou = JaccardIndex(
             task="multiclass",  # type: ignore[arg-type]
-            num_classes=max(data_args.classes) + 1,
-            ignore_index=task_args.ignore_index,
+            num_classes=data_args.num_classes + 1,
+            ignore_index=data_args.ignore_index,
         )
```
guarin: same as above

src/lightly_train/_task_models/dinov2_semantic_segmentation/dinov2_semantic_segmentation_train.py
```diff
@@ -120,19 +117,19 @@ def __init__(
...
         self.val_miou = self.train_miou.clone()
 
         # Classwise MeanIoU for each joint block. Based on EoMT implementation.
         self.train_metrics = ModuleList(
             [
                 MulticlassJaccardIndex(
-                    num_classes=max(data_args.classes) + 1,
+                    num_classes=data_args.num_classes + 1,
```
guarin: Same as above

src/lightly_train/_task_models/dinov2_semantic_segmentation/dinov2_semantic_segmentation_train.py
```diff
@@ -141,10 +138,10 @@ def __init__(
         self.val_metrics = ModuleList(
             [
                 MulticlassJaccardIndex(
-                    num_classes=max(data_args.classes) + 1,
+                    num_classes=data_args.num_classes + 1,
```
guarin: Same as above

src/lightly_train/_task_models/dinov2_semantic_segmentation/dinov2_semantic_segmentation_transforms.py
```diff
@@ -80,6 +80,7 @@ class DINOv2SemanticSegmentationTrainTransformArgs(SemanticSegmentationTransform
     Defines default transform arguments for semantic segmentation training with DINOv2.
     """
 
+    ignore_index: int = -100
```
guarin: Let's remove defaults here to make 100% sure that we pass the value from the outside.

src/lightly_train/_task_models/dinov2_semantic_segmentation/dinov2_semantic_segmentation_transforms.py
```diff
@@ -102,6 +103,7 @@ class DINOv2SemanticSegmentationValTransformArgs(SemanticSegmentationTransformAr
     Defines default transform arguments for semantic segmentation validation with DINOv2.
     """
 
+    ignore_index: int = -100
```
guarin: Same as above.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
stegmuel: Learned something. 
Opted for the print option as we are going to move the function outside of the dataset in the future.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
stegmuel: In this context I don't think it matters as the device is always cpu in the dataset, but I agree that it is a good habit.

src/lightly_train/_commands/train_task_helpers.py
stegmuel: Because the shape of  `target["labels"]` and `target["masks"]` depends on the number of instances in the images. Hence adding an extra dimension at position 0 and stacking does not work.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
stegmuel: Ah yes properties is much cleaner.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
stegmuel: In the end, I changed it to always be -100. I was not able to find a scenario where this would not work. Even if the user was to set one of the class to be -100 it would work as the class are first re-mapped to be between 0 and `num_included_classes -1 `.

PR #194, opened 2025-07-23
Fix train task batch size handling

guarin:
## What has changed and why?

* Divide global batch size by world size for dataloader

## How has it been tested?

* Manual

- [ ] Yes

- [ ] Yes


src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -256,12 +256,6 @@ def get_steps(steps: int | Literal["auto"]) -> int:
     return steps
 
 
-def get_batch_size(batch_size: int | Literal["auto"]) -> int:
```
guarin, line 259: Unused code

PR #197, opened 2025-07-24
Save segmentation model config in checkpoint

guarin:
## What has changed and why?

* Save segmentation model info in checkpoint
* Add option to load segmentation model from checkpoint
* Load model from checkpoint for export

The overall design is that we save the model classpath and the arguments that were passed to the model on initialization. From this we can then reconstruct the model. This way we don't have to save the full model instance in the checkpoint. It is also more robust than saving for example the model name and then mapping that back to a model class and arguments.

## How has it been tested?

* Unit test
* Manual

- [ ] Yes

- [ ] Yes


tests/_commands/test_export_task.py
guarin, line 1: Had to remove this file because it was hard to update to the new checkpoint format. I moved the relevant tests directly to `test_train_task.py`. I believe having a single integration test is easier for now than unit tests for individual functions.

PR #204, opened 2025-07-29
Simplify dinov2 model naming

guarin:
## What has changed and why?

* Change DINOv2 naming from `dinov2_vit/vits14_pretrain` to `dinov2/vits14-pretrained`
  * This removes the redundant `vit` part in the package name
  * This makes it more evident that the model is already pretrained
  * This follows model naming convention with dashes instead of underscores

Old names continue working for now. We'll remove them eventually.

## How has it been tested?

* CI
* Manual tests

- [x] Yes

- [x] Yes


src/lightly_train/_models/dinov2_vit/dinov2_vit_src/configs/__init__.py
```diff
@@ -14,9 +14,9 @@
 from omegaconf import OmegaConf
 
 MODELS = {
-    "_vit_test14": {
+    "_vittest14": {
         "url": "",
-        "config": "train/_vit_test14",
+        "config": "train/_vittest14",
```
guarin, line 19: Renamed from `vit_test` to `vittest` to follow naming schema for other model versions (`vits`, `vitl` etc) where don't have a space between the `vit` part and the model size.

PR #205, opened 2025-07-29
expose model loading

liopeer:
## What has changed and why?
- creates the user-facing function `load_task_model_from_checkpoint` which allocates the model on the passed device
- adds the `.device` property to the `TaskModel` which I use for logging in above code and is most likely convenient anyway

## How has it been tested?
 - ran it on macOS with `device={"mps", "cpu"}`

- [ ] Yes
- [x] Not needed for now, add it once we document it

- [ ] Yes
- [x] Not needed for now, make a separate PR to expose it

### DIFF
--- a/src/lightly_train/_commands/load_task_model_from_checkpoint.py
+++ b/src/lightly_train/_commands/load_task_model_from_checkpoint.py
@@ -1,0 +1,82 @@
+#
+# Copyright (c) Lightly AG and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+#
+from __future__ import annotations
+
+import logging
+from typing import Literal
+
+import torch
+from torch import distributed
+
+from lightly_train._task_models.task_model import TaskModel
+import lightly_train._distributed as train_distributed
+from lightly_train import _logging
+from lightly_train._commands import _warnings, common_helpers
+from lightly_train._configs.config import PydanticConfig
+from lightly_train._task_models import task_model_helpers
+from lightly_train.types import PathLike
+
+logger = logging.getLogger(__name__)
+
+def load_task_model_from_checkpoint(
+        *,
+        checkpoint: PathLike,
+        device: Literal["cpu", "cuda", "mps"] | None = None,
+) -> TaskModel:
+    """Load a task model from a checkpoint file.
+
+    Args:
+        checkpoint:
+            Path to the LightlyTrain checkpoint that contains the model.
+        device:
+            Device to load the model on. If None, the model will be loaded to the device
+            it was saved on.
+
+    Returns:
+        The loaded model.
+    """
+    config = LoadTaskModelConfig(**locals())
+    return load_task_model_from_config(config=config)
+
+def load_task_model_from_config(config: LoadTaskModelConfig) -> TaskModel:
+    # Set up logging.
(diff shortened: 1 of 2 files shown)


src/lightly_train/_commands/load_task_model_from_checkpoint.py
```diff
@@ -0,0 +1,82 @@
...
+        device: Literal["cpu", "cuda", "mps"] | None = None,
+) -> TaskModel:
+    """Load a task model from a checkpoint file.
+
+    Args:
+        checkpoint:
+            Path to the LightlyTrain checkpoint that contains the model.
+        device:
+            Device to load the model on. If None, the model will be loaded to the device
+            it was saved on.
```
guarin: I think loading to the device the model was saved on as default is not an ideal as models oftentimes move between devices from training and inference. IMO the default should use a single GPU CUDA if this is available, then MPS if available, then CPU.

src/lightly_train/_commands/load_task_model_from_checkpoint.py
```diff
@@ -0,0 +1,82 @@
...
+def _resolve_device(device: str | None) -> torch.device | None:
+    """Resolve the device to load the model on."""
+    if device is None:
+        return None
+    elif device == "cpu":
+        return torch.device("cpu")
+    elif device == "cuda":
+        return torch.device(train_distributed.get_local_rank() or "cuda")
+    elif device == "mps":
+        return torch.device("mps")
```
guarin: Can we not pass the user provided `device` directly to the `task_model.to(device)` function? IMO we should only select a device if it is `None` and otherwise respect the user provided value.

src/lightly_train/_commands/load_task_model_from_checkpoint.py
```diff
@@ -0,0 +1,82 @@
...
+    else:
+        raise TypeError(f"Expected str or None, got {type(device)}")
```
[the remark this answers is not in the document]
liopeer: Addressed!

src/lightly_train/_commands/load_task_model_from_checkpoint.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class LoadTaskModelConfig(PydanticConfig):
```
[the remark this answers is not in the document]
liopeer: Not public facing, no priority.

src/lightly_train/_commands/load_task_model_from_checkpoint.py
liopeer: Agree, addressed this way!

src/lightly_train/_commands/load_task_model_from_checkpoint.py
liopeer: Done.

src/lightly_train/_task_models/task_model.py
```diff
@@ -58,6 +59,11 @@ def class_path(self) -> str:
         """
         return f"{self.__module__}.{self.__class__.__name__}"
 
+    @property
+    def device(self) -> torch.device:
+        """Returns the device the model is currently on."""
+        return next(self.parameters()).device
+
```
guarin: Not needed anymore

liopeer: True, removing it!

PR #207, opened 2025-07-29
Ignore initial backbone weights for checkpoint saving

guarin:
## What has changed and why?

* Skips saving of the backbone weights in `init_args`

This way the path to the backbone weights are not saved in the checkpoint which makes sense as they are only needed at the beginning when the model is loaded. On subsequent model loads we no longer want to load these weights. 

## How has it been tested?

* Manual

- [ ] Yes

- [ ] Yes


liopeer (approved):
Thanks, that makes a lot of sense!

PR #212, opened 2025-07-31
Thomas trn 1368 fix mask annealing steps

stegmuel:
## What has changed and why?

The mask annealing steps are automatically inferred such that it works for datasets other than ADE20K.

## How has it been tested?

Verified that the inferred steps are close to the original values for ADE20K.

- [ ] Yes

- [ ] Yes


guarin (approved):
LGTM

PR #215, opened 2025-07-31
Change default temporary directory and Refactorings

liopeer:
## What has changed and why?
 - change the default of the temporary directory
 - add additional cache directories, specifically for data and and model weights

## How has it been tested?
 - changed unit tests acoordingly

- [x] Yes

- [ ] Yes
- [x] Not needed (we currently do not document these env variables & I wouldn't for now)


src/lightly_train/_env.py
```diff
@@ -19,13 +19,18 @@
 @dataclass(frozen=True)
 class EnvVar(Generic[T]):
     name: str
-    default: T
+    default_: Callable[[], T] | T
```
liopeer: I can only propagate defaults from one environment variable to another by allowing the default to also be a `callable`.

src/lightly_train/_commands/common_helpers.py
```diff
@@ -137,17 +136,11 @@ def get_out_dir(out: PathLike, resume_interrupted: bool, overwrite: bool) -> Pat
 
 def get_tmp_dir() -> Path:
     """Get the temporary directory for Lightly Train."""
-    tmp_dir = Env.LIGHTLY_TRAIN_TMP_DIR.value
-    if tmp_dir is None:
```
liopeer, line 141: The `None` is not needed anymore, since this is handled inside the `Env` class.

src/lightly_train/_commands/common_helpers.py
```diff
@@ -137,17 +136,11 @@ def get_out_dir(out: PathLike, resume_interrupted: bool, overwrite: bool) -> Pat
...
-        tmp_dir = Path(tempfile.gettempdir()) / "lightly-train"
+    tmp_dir = Env.LIGHTLY_TRAIN_TMP_DIR.value.expanduser().resolve()
+    tmp_dir.mkdir(parents=True, exist_ok=True)
     return tmp_dir
 
 
-def get_data_tmp_dir() -> Path:
```
liopeer, line 146: Moved to `cache.py`.

src/lightly_train/_data/cache.py
```diff
@@ -10,10 +10,21 @@
 from lightly_train._env import Env
 
 
-def get_cache_dir() -> Path:
```
liopeer, line 13: We can reimplement this once we actually need it.

tests/_commands/test_train_task.py
```diff
@@ -67,6 +67,9 @@ def test_train_semantic_segmentation(tmp_path: Path) -> None:
         model_args={
             "num_joint_blocks": 1,  # Reduce joint blocks for _vittest14
         },
+        # The operator 'aten::upsample_bicubic2d.out' raises a NotImplementedError
+        # on macOS with MPS backend.
+        accelerator="auto" if not sys.platform.startswith("darwin") else "cpu",
```
liopeer, line 72: Unrelated!

src/lightly_train/_env.py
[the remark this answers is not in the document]
guarin: Nit: Would rename to `_default` to hide it as the defaults are anyways only used from within the class and shouldn't be accessed externally. And rename `type_` to `_type` with it. Then it is consistent.

src/lightly_train/_env.py
```diff
@@ -56,86 +61,101 @@ def raw_value(self) -> str | None:
...
         name="LIGHTLY_TRAIN_TMP_DIR",
-        default=None,
+        default_=lambda: Env.LIGHTLY_TRAIN_CACHE_DIR.value / "data",
+        type_=Path,
+    )
+    # Path to directory where data is cached. These are mainly the memory-mapped files.
+    # TODO(Lionel, 08/25): Change the default to LIGHTLY_TRAIN_CACHE_DIR.value / "data".
+    LIGHTLY_TRAIN_DATA_CACHE_DIR: EnvVar[Path] = EnvVar(
+        name="LIGHTLY_TRAIN_DATA_CACHE_DIR",
+        default_=lambda: Env.LIGHTLY_TRAIN_TMP_DIR.value,
```
guarin: Why can we not yet change it to `LIGHTLY_TRAIN_CACHE_DIR.value / "data"`?

guarin (approved):
LGTM

src/lightly_train/_env.py
liopeer: We could, but then `LIGHTLY_TRAIN_TMP_DIR` would not behave the same way as it does right now. Currently our users expect `LIGHTLY_TRAIN_TMP_DIR` to define where the memory-mapped files land. With the changes I introduced, this will still be the case (unless of course the override `LIGHTLY_TRAIN_DATA_CACHE_DIR` as well, but this is not yet shared/documented anywhere).

If it's okay to break this behaviour, we can use `LIGHTLY_TRAIN_CACHE_DIR.value / "data"` right now.

PR #218, opened 2025-08-01
Add semantic segmentation documentation

guarin:
## What has changed and why?

* Add documentation for semantic segmentation

## How has it been tested?

[Semantic Segmentation - LightlyTrain documentation - 3.pdf](https://github.com/user-attachments/files/21578518/Semantic.Segmentation.-.LightlyTrain.documentation.-.3.pdf)

[LightlyTrain documentation - semantic segmentation.pdf](https://github.com/user-attachments/files/21544247/LightlyTrain.documentation.-.semantic.segmentation.pdf)

- [x] Yes

- [x] Yes

### DIFF
--- a/docs/source/semantic_segmentation.md
+++ b/docs/source/semantic_segmentation.md
@@ -10,7 +10,9 @@ Kerssies et al. and reaches 58.4% mIoU on the ADE20K dataset.
 ```
 
 Training a semantic segmentation model with LightlyTrain is straightforward and
-only requires a few lines of code:
+only requires a few lines of code. The dataset must follow the [ADE20K format](https://ade20k.csail.mit.edu/)
+with RGB images and integer masks in PNG format. See [data](#semantic-segmentation-data)
+for more details.
 
 ````{tab} Python
 ```python
@@ -19,7 +21,7 @@ import lightly_train
 if __name__ == "__main__":
     lightly_train.train_semantic_segmentation(
         out="out/my_experiment",
-        model="dinov2/vits14-eomt",
+        model="dinov2/vitl14-eomt",
         data={
             "train": {
                 "images": "my_data_dir/train/images",   # Path to training images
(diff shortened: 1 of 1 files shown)


docs/source/semantic_segmentation.md
```diff
@@ -0,0 +1,317 @@
+(semantic-segmentation)=
+
+# Semantic Segmentation
```
IgorSusmelj, line 3: There is information missing about the input format. How do I need to provide the masks? How are they encoded? png images? If yes, what RGB values do they need to have? Do the masks need to have the same folder structure as the images? Same filenames? We should assume here that the reason has no clue and just quickly wants to train a model. Can we point the reader to a reference dataset or so with example images with correct masks?

Other general questions we should clarify:
- how can I change the input/output resolution of the model? Does it support non square images?

IgorSusmelj: The tutorial is quite long. Some people might only read the first part.

An easy to clarify these open questions would be to add two bullet points:
- the input format follows ADE20k which uses RGB masks saved as .png files [link here to an example dataset]
- the model supports arbitrary input/output resolutions

liopeer: > There is information missing about the input format. How do I need to provide the masks? How are they encoded? png images? If yes, what RGB values do they need to have? Do the masks need to have the same folder structure as the images? Same filenames? We should assume here that the reason has no clue and just quickly wants to train a model.

This is in the `Data` section, which is imo good and in harmony with the documentation of the `lightly_train.train` command. But the `Data` section should come directly after the the initial code samples. I would therefore move the `Optional: Pretrain and Fine-tune a Semantic Segmentation Model` to the end of the page and just reference it at the current location through a `seealso`. That would also make it shorter and partially address:
> The tutorial is quite long. Some people might only read the first part.

> Can we point the reader to a reference dataset or so with example images with correct masks?

Fully agree that we need this. To keep consistency with the rest of the documentation however, I would prefer seeing that in the `Quickstart`. Will this make the Quickstart too bloated?

docs/source/semantic_segmentation.md
```diff
@@ -0,0 +1,317 @@
...
+`ignore_classes` argument. The trained model will then not predict these classes.
+
+(semantic-segmentation-model)=
+
+## Model
+
+The `model` argument defines the model used for semantic segmentation training. The
+following models are available:
+
+- `dinov2/vits14-eomt`
```
liopeer, line 188: We should mention our own checkpoints (that we will upload) here as well. Can you also load the official EoMT weights? If yes, we should maybe add these as well?

docs/source/semantic_segmentation.md
```diff
@@ -0,0 +1,317 @@
...
+
+````{tab} Python
+```python
+logger_args={"tensorboard": None}
+```
+````
+
+### Performance Optimizations
+
+For performance optimizations, e.g. using accelerators, multi-GPU, multi-node, and
```
liopeer: Do we log things like `data_time`, `batch_time` etc. as mentioned in the *Performance* docs? Is this automatically logged in Lightning Fabric as well? If not, I would probably just leave the performance section away for now.

docs/source/semantic_segmentation.md
guarin: Will do that once they are ready

docs/source/semantic_segmentation.md
guarin: Removed

PR #219, opened 2025-08-01
Use DINOv2 pretrained models by default

guarin:
## What has changed and why?

* Change `dinov2/vits14` to load a pretrained model by default
  * Previously, pretrained models had to be specified with `dinov2/vits14-pretrained`. However there are very few scenarios where a non-pretrained model is required. Therefore we change the default to load a pretrained model.

Not pretrained models can be requested with `dinov2/vits14-notpretrained` if needed.

## How has it been tested?

* Unit tests
* Manual tests

- [x] Yes

- [x] Yes

### DIFF
--- a/src/lightly_train/_models/log_usage_example.py
+++ b/src/lightly_train/_models/log_usage_example.py
@@ -6,6 +6,7 @@
 # LICENSE file in the root directory of this source tree.
 #
 from __future__ import annotations
+
 from os import linesep
 
 


src/lightly_train/_models/dinov2_vit/dinov2_vit_src/configs/__init__.py
```diff
@@ -7,80 +7,142 @@
...
 
 from omegaconf import OmegaConf
 
 MODELS = {
+    # Test model for development purposes only.
     "_vittest14": {
         "url": "",
         "config": "train/_vittest14",
-    },  # This is a test model for development purposes only.
+        "list": False,
```
guarin: Added the `"list"` key to indicate whether a model should be shown in `list_model_names` or not. We only show the pretrained models by default.

src/lightly_train/_models/package_helpers.py
```diff
@@ -139,25 +138,3 @@ def _parse_model_name(model: str) -> tuple[str, str]:
     if package_name == "dinov2_vit":  # For backwards compatibility.
         package_name = "dinov2"
     return package_name, model_name
-
-
-def format_log_msg_model_usage_example(log_message_code_block: list[str]) -> str:
```
guarin, line 144: Moved this out of `package_helpers.py` because it resulted in cyclic imports. `package_helpers.py` should only be accessed from outside of the `_models` directory. It is basically the entrypoint to packages from any code that lives outside `_models`. The issue with `format_log_msg_model_usage_example` is that it is used in all the individual package implementations within `_models` and must therefore be in a different file, otherwise `package_helpers.py` is accessed from inside and outside the `_models` module at the same time.

src/lightly_train/_models/dinov2_vit/dinov2_vit_src/configs/__init__.py
```diff
@@ -7,81 +7,155 @@
...
+        config="train/vitl14",
+        list=False,
+    ),
+    "vitg14-noreg-notpretrained": _Model(
+        url="",
+        config="train/vitg14",
+        list=False,
+    ),
+    # Models with `-pretrained` suffix for backwards compatibility.
+    "vits14-pretrained": _Model(
```
stegmuel, line 119: Is this needed? Did we release after changing to "-pretrained"?

tests/_models/dinov2_vit/test_dinov2_vit_package.py
```diff
@@ -22,33 +22,39 @@
 
 class TestDINOv2ViTPackage:
     @pytest.mark.parametrize(
-        "model_name, supported",
+        "model_name, listed",
         [
+            ("dinov2/vits14", True),
```
stegmuel, line 27: Is this supposed to be exhaustive?

src/lightly_train/_models/dinov2_vit/dinov2_vit_src/configs/__init__.py
guarin: The code now automatically changes `_pretrain` to `-pretrained`. This was introduced here: https://github.com/lightly-ai/lightly-train/pull/204

tests/_models/dinov2_vit/test_dinov2_vit_package.py
guarin: No :)

PR #221, opened 2025-08-01
Allow uniform masks for semantic segmentation

guarin:
## What has changed and why?

* Allow uniform masks for semantic segmentation

This avoids issues with validation datasets that have uniform masks. This shouldn't be a big issue for training as the model should be able to handle such masks. 

## How has it been tested?

* CI

- [ ] Yes

- [ ] Yes


stegmuel (approved):
LGTM.

PR #230, opened 2025-08-07
Add Autoresolve for Number of Epochs

liopeer:
## What has changed and why?
 - introduces "auto" for the epochs and automatically sets it, depending on whether DINOv2 is used or not

## How has it been tested?
 - added unit tests

- [x] Yes

- [x] Yes

### DIFF
--- a/src/lightly_train/_commands/train_helpers.py
+++ b/src/lightly_train/_commands/train_helpers.py
@@ -359,11 +359,11 @@ def get_epochs(
     elif method_args_cls.default_steps is not None:
         logger.debug(f"Using default steps {method_args_cls.default_steps}.")
         # Calculate epochs from steps.
-        epochs = math.ceil(method_args_cls.default_steps * batch_size / dataset_size)
+        _epochs = math.ceil(method_args_cls.default_steps * batch_size / dataset_size)
         logger.debug(
             f"Calculated epochs {epochs} from steps {method_args_cls.default_steps}."
         )
-        return epochs
+        return _epochs
     else:
         raise ValueError(
             f"An unexpected error occurred while determining the number of epochs for method '{method_args_cls.__name__}'. "


src/lightly_train/_commands/train_helpers.py
```diff
@@ -327,6 +335,47 @@ def get_method(
...
+        logger.debug(
+            f"Using default epochs {method_args_cls.default_epochs} for method '{method_args_cls.__name__}'."
+        )
+        return method_args_cls.default_epochs
+    elif method_args_cls.default_steps is not None:
+        logger.debug(
+            f"Using default steps {method_args_cls.default_steps} for method '{method_args_cls.__name__}'."
+        )
+        # Calculate epochs from steps.
+        epochs = method_args_cls.default_steps * batch_size // dataset_size
```
guarin:
```suggestion
        epochs = math.ceil(method_args_cls.default_steps * batch_size / dataset_size)
```

src/lightly_train/_commands/train_helpers.py
```diff
@@ -327,6 +335,47 @@ def get_method(
...
+
+    if method_args_cls.default_epochs is not None:
+        logger.debug(
+            f"Using default epochs {method_args_cls.default_epochs} for method '{method_args_cls.__name__}'."
+        )
+        return method_args_cls.default_epochs
+    elif method_args_cls.default_steps is not None:
+        logger.debug(
+            f"Using default steps {method_args_cls.default_steps} for method '{method_args_cls.__name__}'."
+        )
```
guarin: Nit: No need to log the method cls name IMO.

docs/source/methods/dinov2.md
```diff
@@ -32,7 +32,7 @@ if __name__ == "__main__":
         data="my_data_dir",
         model="dinov2/vitb14",
         method="dinov2",
-        epochs=100, # We recommend epochs = 125000 * batch size // dataset size
+        epochs=100,
```
guarin: We now infer it automatically 😉 
```suggestion
```

src/lightly_train/_commands/train_helpers.py
```diff
@@ -327,6 +335,47 @@ def get_method(
...
+) -> int:
+    method_args_cls = method_helpers.get_method_cls(method).method_args_cls()
+
+    assert not (
+        method_args_cls.default_epochs is None and method_args_cls.default_steps is None
+    )
+    assert not (
+        isinstance(method_args_cls.default_epochs, int)
+        and isinstance(method_args_cls.default_steps, int)
+    )
```
guarin, line 350:
```suggestion
    assert not (
        method_args_cls.default_epochs is not None
        and method_args_cls.default_steps is not None
    )
```

src/lightly_train/_commands/train_helpers.py
```diff
@@ -327,6 +335,47 @@ def get_method(
     )
 
 
+def get_epochs(
+    method: str, epochs: int | Literal["auto"], dataset_size: int, batch_size: int
+) -> int:
+    method_args_cls = method_helpers.get_method_cls(method).method_args_cls()
```
guarin, line 342: This looks to me like we should initialize `method_args_cls` once inside `def train_from_config()` and then reuse it through the whole function. Doesn't need to be addressed in this PR, just an observation.

guarin (changes requested):

docs/source/methods/dinov2.md
guarin: Please also update it in `semantic_segmentation.md`

liopeer: Whoops. I searched for this in VSCode and thought that I had changed it everywhere.

src/lightly_train/_commands/train_helpers.py
liopeer: That's indeed better.

src/lightly_train/_commands/train_helpers.py
liopeer: True, anyway appears in the logs.

docs/source/methods/dinov2.md
liopeer: Ah, now I see. You obviously mean not only the comment, but the whole argument. 💡

src/lightly_train/_commands/train_helpers.py
liopeer: Would leave it as it is. Otherwise `mypy` complains.

src/lightly_train/_commands/train_helpers.py
liopeer: Could potentially make it a bit cleaner, but would not change right now. It's quite okay imo.

guarin (approved):

PR #231, opened 2025-08-07
Infer num_joint_blocks from model name

stegmuel:
## What has changed and why?

The number joint blocks is automatically inferred in EoMT to accommodate other models than the ViT-L

## How has it been tested?

-

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/train_model.py
+++ b/src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/train_model.py
@@ -78,11 +78,7 @@ class DINOv2EoMTSemanticSegmentationTrainArgs(TrainModelArgs):
     metric_log_classwise: bool = True
     metric_log_debug: bool = False
 
-    def resolve_auto(self, total_steps: int, **kwargs: Any) -> None:
-        # Get the model name.
-        model_name = kwargs["model_name"]
-        assert model_name is not None
-
+    def resolve_auto(self, total_steps: int, model_name: str) -> None:
         if self.num_joint_blocks == "auto":
             match = re.match(r"(dinov2(?:_vit)?)/(vit[slbg]).*", model_name)
             assert match is not None, (
--- a/src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/train_model.py
+++ b/src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/train_model.py
@@ -85,7 +85,9 @@ class DINOv2EoMTSemanticSegmentationTrainArgs(TrainModelArgs):
 
         if self.num_joint_blocks == "auto":
             match = re.match(r"(dinov2(?:_vit)?)/(vit[slbg]).*", model_name)
-            assert match is not None, f"Could not parse model size from model_name='{model_name}'"
+            assert match is not None, (
+                f"Could not parse model size from model_name='{model_name}'"
+            )
             model_size = match.group(1)
             self.num_joint_blocks = {
                 "vits": 3,
--- a/src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/train_model.py
+++ b/src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/train_model.py
@@ -7,6 +7,7 @@
 #
 from __future__ import annotations
 
+import re
 from typing import Any, ClassVar, Literal
 
 import torch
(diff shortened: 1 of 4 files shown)


src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/train_model.py
```diff
@@ -77,7 +78,24 @@ class DINOv2EoMTSemanticSegmentationTrainArgs(TrainModelArgs):
     metric_log_classwise: bool = True
     metric_log_debug: bool = False
 
-    def resolve_auto(self, total_steps: int) -> None:
+    def resolve_auto(self, total_steps: int, **kwargs: Any) -> None:
```
guarin: Let's make this explicit everywhere otherwise it will become quickly very difficult to understand which arguments are passed where
```suggestion
    def resolve_auto(self, total_steps: int, model_name: str) -> None:
```

src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/train_model.py
```diff
@@ -77,7 +78,24 @@ class DINOv2EoMTSemanticSegmentationTrainArgs(TrainModelArgs):
...
+        # Get the model name.
+        model_name = kwargs["model_name"]
+        assert model_name is not None
+
+        if self.num_joint_blocks == "auto":
+            match = re.match(r"(dinov2(?:_vit)?)/(vit[slbg]).*", model_name)
+            assert match is not None, (
+                f"Could not parse model size from model_name='{model_name}'"
+            )
```
guarin: This is the first place in the code where we validate the model name. We have to make this error more actionable for the user otherwise it is very confusing what they can do to solve the error:
```suggestion
            raise ValueError("Unknown model name '{model_name}', see https://docs.lightly.ai/train/stable/semantic_segmentation.html#model for all supported models.")
```

src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/train_model.py
```diff
@@ -118,6 +136,7 @@ def __init__(
         )
 
         self.model_args = model_args
+        assert isinstance(model_args.num_joint_blocks, int)
```
guarin: Not needed, the `no_auto` function already asserts this.

src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/train_model.py
```diff
@@ -203,6 +222,7 @@ def get_task_model(self) -> DINOv2EoMTSemanticSegmentation:
     def training_step(
         self, fabric: Fabric, batch: MaskSemanticSegmentationBatch, step: int
     ) -> TaskStepResult:
+        assert isinstance(self.model_args.num_joint_blocks, int)
```
guarin: Same as above

src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/train_model.py
```diff
@@ -292,6 +312,7 @@ def training_step(
     def validation_step(
         self, fabric: Fabric, batch: MaskSemanticSegmentationBatch
     ) -> TaskStepResult:
+        assert isinstance(self.model_args.num_joint_blocks, int)
```
guarin: Same as above

src/lightly_train/_task_models/train_model.py
```diff
@@ -27,7 +27,7 @@ class TrainModelArgs(PydanticConfig):
     default_batch_size: ClassVar[int]
     default_steps: ClassVar[int]
 
-    def resolve_auto(self, total_steps: int) -> None:
+    def resolve_auto(self, total_steps: int, **kwargs: Any) -> None:
```
guarin: No kwargs :)

src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/train_model.py
stegmuel: The reason why I used kwargs that if I use model_name here, I also need to use it in the mother class and I  wasn't sure we wanted to always use model_name?

src/lightly_train/_task_models/train_model.py
stegmuel: If I just remove kwargs and use model_name in the inherited class mypy complains.

src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/train_model.py
stegmuel: Currently I only use `no_auto`  [here](https://github.com/lightly-ai/lightly-train/blob/47f426bf18e45524451ed8105c704f86928d1b7c/src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/train_model.py#L151) so my understanding is that it only helps with usage of `num_joint_blocks` within `DINOv2EoMTSemanticSegmentation`.
So do you agree that I would have to add some `no_auto` calls ?

src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/train_model.py
guarin: Better to add it to the parent class and pass everywhere.

src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/train_model.py
guarin: Yes

PR #232, opened 2025-08-07
Add normalization to inference model

liopeer:
## What has changed and why?

## How has it been tested?

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -32,6 +32,9 @@ from lightly_train._loggers.mlflow import MLFlowLogger
 from lightly_train._loggers.task_logger_args import TaskLoggerArgs
 from lightly_train._loggers.tensorboard import TensorBoardLogger
 from lightly_train._task_checkpoint import TaskSaveCheckpointArgs
+from lightly_train._task_models.dinov2_eomt_semantic_segmentation.task_model import (
+    DINOv2EoMTSemanticSegmentation,
+)
 from lightly_train._task_models.dinov2_eomt_semantic_segmentation.train_model import (
     DINOv2EoMTSemanticSegmentationTrain,
     DINOv2EoMTSemanticSegmentationTrainArgs,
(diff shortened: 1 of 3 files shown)


src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/task_model.py
```diff
@@ -34,13 +34,27 @@
...
+
 class DINOv2EoMTSemanticSegmentation(TaskModel):
     def __init__(
         self,
         *,
         model_name: str,
         classes: dict[int, str],
         class_ignore_index: int | None,
+        image_size: tuple[int, int],
+        image_normalize: dict[str, float],
```
liopeer, line 45:
```suggestion
        image_size: tuple[int, int] | None = None,
        image_normalize: dict[str, float] | None = None,
```
Wdyt? If `None` it would just use the validation normalization from the training. But the user could overwrite it.

guarin: I wouldn't allow `None` for now. For now there is no reason that users create `DINOv2EoMTSemanticSegmentation` instances manually and we always want to have normalize/image_size defined.

liopeer: Okay. Then it makes sense. In the doc string of `TaskModel` (the super-class of `DINOv2EoMTSemanticSegmentation`) it says the following:
```python
"""Base class for task-specific models that the user interacts with.

    Must implement the forward method for inference. Must be pure PyTorch and not rely
    on Fabric or Lightning modules.
    """
```
So I guess I thought that this was already public. Should have checked the docs, where only the `lightly_train.load_model_from_checkpoint` is documented.

guarin: Yes it might be a bit confusing. Right now users use with the class but do not directly instantiate it.

src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/task_model.py
```diff
@@ -34,13 +34,27 @@
 logger = logging.getLogger(__name__)
 
 
+class DinoV2EoMTSemanticSegmentationArgs:
```
liopeer: We don't want this for now.

src/lightly_train/_task_models/train_model.py
```diff
@@ -27,6 +28,9 @@ class TrainModelArgs(PydanticConfig):
     default_batch_size: ClassVar[int]
     default_steps: ClassVar[int]
 
+    train_transform_cls: ClassVar[type[TaskTransform]]
```
liopeer: These should go into the respective `TrainModel` subclasses. And then we need a helper function that goes over all the `TrainModel` subclasses and figures out the correct one.

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -278,15 +311,24 @@ def get_steps(steps: int | Literal["auto"], default_steps: int) -> int:
...
 def get_train_model_args_cls(
     model_name: str, model_args: dict[str, Any] | TrainModelArgs | None
 ) -> type[TrainModelArgs]:
     if isinstance(model_args, TrainModelArgs):
         return model_args.__class__
 
     # TODO(Guarin, 08/25): Properly handle model name and args linking.
-    if model_name.endswith("-eomt"):
-        return DINOv2EoMTSemanticSegmentationTrainArgs
+    for train_model_cls in TASK_TRAIN_MODEL_CLASSES:
```
guarin: The `get_train_model_args_cls ` function is not needed anymore

src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/train_model.py
```diff
@@ -43,6 +49,13 @@ class DINOv2EoMTSemanticSegmentationTrainArgs(TrainModelArgs):
     # 20210 images / batch size 16 * 31 epochs ~= 40k steps.
     default_steps: ClassVar[int] = 40_000
 
+    train_transform_cls: ClassVar[type[TaskTransform]] = (
+        DINOv2SemanticSegmentationTrainTransform
+    )
+    val_transform_cls: ClassVar[type[TaskTransform]] = (
+        DINOv2SemanticSegmentationValTransform
+    )
+
```
guarin: Remove, we already have it in the `TrainModel` class and don't need it in the args anymore

guarin (changes requested):

src/lightly_train/_commands/train_task_helpers.py
liopeer: Good catch, thanks!

src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/train_model.py
liopeer: Another good catch, thanks!

PR #234, opened 2025-08-08
Add Object Detection Dataset

liopeer:
## What has changed and why?

## How has it been tested?

- [ ] Yes

- [ ] Yes


src/lightly_train/_task_models/train_model.py
```diff
@@ -58,10 +58,6 @@ class TrainModel(Module):
     # - predict_step
     # See: https://github.com/Lightning-AI/pytorch-lightning/blob/95f16c12fe23664ffa5198a43266f715717c6f45/src/lightning/fabric/wrappers.py#L47-L48
 
-    @classmethod
```
liopeer, line 61: We do this now in `TaskModel`.

src/lightly_train/_data/yolo_object_detection_dataset.py
```diff
@@ -0,0 +1,130 @@
...
+
+        print(
+            type(transformed["image"]),
+            type(transformed["bboxes"]),
+            type(transformed["class_labels"]),
+        )
+
+        image = transformed["image"]
+        bboxes = torch.from_numpy(transformed["bboxes"]).float()
+        class_labels = torch.from_numpy(transformed["class_labels"]).long()
```
liopeer, line 68: For some reason `ToTensorv2` from `albumentations` seems to only act on the image, but not the bounding boxes/classes. Need to investigate if this is intended behaviour.

src/lightly_train/_data/file_helpers.py
```diff
@@ -159,3 +165,17 @@ def open_image_numpy(
...
+
+def open_yolo_label_numpy(label_path: Path) -> tuple[NDArrayBBoxes, NDArrayClasses]:
+    """Open a YOLO label file and return the bounding boxes and classes as numpy arrays."""
+    bboxes = []
+    classes = []
+    with open(label_path, "r") as f:
+        for line in f.readlines():
+            class_id, x_center, y_center, width, height = (
+                float(x) for x in line.strip().split()
+            )
```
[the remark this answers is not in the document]
guarin: Would handle the empty lines.

src/lightly_train/_data/file_helpers.py
```diff
@@ -159,3 +165,17 @@ def open_image_numpy(
...
+            bboxes.append([x_center, y_center, width, height])
+            classes.append(int(class_id))
+    return np.array(bboxes), np.array(classes)
```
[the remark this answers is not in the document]
guarin: ^agree

src/lightly_train/_data/yolo_object_detection_dataset.py
```diff
@@ -0,0 +1,123 @@
...
+    path: Path
+    train: Path
+    val: Path
+    test: Path | None = None
+    names: dict[int, str]
+
+    @pydantic.field_validator("train", "val", mode="after")
+    def validate_paths(cls, v: Path) -> Path:
+        if "images" not in str(v):
+            raise ValueError(f"Expected path to include 'images', got {v}.")
```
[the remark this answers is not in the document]
guarin: Is this really needed? AFAIK the only assumption is that there is a `labels` directory in the parent directory of the `train`, `val`, `test` paths. I am not aware that the images directory must be named "images"

src/lightly_train/_data/yolo_object_detection_dataset.py
```diff
@@ -0,0 +1,130 @@
...
+import torch
+from torch.utils.data import Dataset
+
+from lightly_train._configs.config import PydanticConfig
+from lightly_train._data import file_helpers
+from lightly_train._transforms.task_transform import TaskTransform
+from lightly_train.types import ObjectDetectionDatasetItem
+
+
+class YoloObjectDetectionDatasetArgs(PydanticConfig):
```
guarin:
```suggestion
class YOLOObjectDetectionDatasetArgs(PydanticConfig):
```

guarin: Please add comment which yolo versions are supported. AFAIK at least ultralytics changed the format between versions.

src/lightly_train/_data/yolo_object_detection_dataset.py
```diff
@@ -0,0 +1,130 @@
...
+        test_label_dir = self._args.path / test_label_path if test_label_path else None
+
+        if self.mode == "train":
+            return train_img_dir, train_label_dir
+        elif self.mode == "val":
+            return val_img_dir, val_label_dir
+        elif self.mode == "test":
+            return test_img_dir, test_label_dir
+        else:
+            raise ValueError(f"Unknown mode: {self.mode}")
```
guarin: Would move this logic to the args class and follow the example from `MaskSemanticSegmentationDataset` where we distinguish between:
1. `MaskSemanticSegmentationDataArgs`: What the user passes as `data` argument
2. `MaskSemanticSegmentationDatasetArgs`: What is passed to the dataset as `args`

With this approach the logic of the different `DataArgs` classes lives in the same place as the class definition. This will make it easier to adjust for other object detection dataset structures.

src/lightly_train/_data/yolo_object_detection_dataset.py
guarin: Good find!

src/lightly_train/_transforms/task_transform.py
```diff
@@ -7,20 +7,42 @@
...
-    pass
+    # We use the YOLO format internally for now.
+    bbox_params: BboxParams = Field(
+        default_factory=lambda: BboxParams(
+            format="yolo", label_fields=["class_labels"]
+        ),
+    )
+
+    class Config:
+        arbitrary_types_allowed = True
```
guarin:
```suggestion
    model_config = ConfigDict(arbitrary_types_allowed=True)
```

For consistency with the rest of the codebase, see e.g. `TrainConfig`

src/lightly_train/_transforms/task_transform.py
guarin, line 1: Please move the logic you added to this file to a new file called `object_detection_transform.py` in the same directory. This follows the example from `semantic_segmentation_transform.py`

src/lightly_train/_data/file_helpers.py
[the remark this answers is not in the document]
liopeer: Agree. But I don't handle malformed lines. This imo should throw an error (at least for now).

src/lightly_train/_data/file_helpers.py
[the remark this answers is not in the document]
liopeer: Done.

src/lightly_train/_data/yolo_object_detection_dataset.py
liopeer: Done.

liopeer: According to labelformat it never changed: https://labelformat.com/formats/object-detection/yolov11/

On all pages it says: 
> Despite these improvements, YOLOv11 retains the same object detection format as its predecessors, utilizing normalized coordinates in text files. This consistency ensures seamless integration and compatibility with existing workflows while benefiting from YOLOv11's enhanced performance.

Also ChatGPT said that the format never changed.

liopeer: But you might be right that `"images"` is not strictly required (from a different comment).

src/lightly_train/_data/yolo_object_detection_dataset.py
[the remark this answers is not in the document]
liopeer: Honestly, this is a bit of a mess on their side. I don't know what kind of weird resolving they do in order to support directories that are different from `"images"`. But here (which is used in thei `YOLODataset` class): 

https://github.com/ultralytics/ultralytics/blob/edd84f880ba6b751ca8fc453c10419053e4c1356/ultralytics/data/dataset.py#L167

They definitely expect `"images"` and replace it with `"labels"`.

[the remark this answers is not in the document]
liopeer: Their documentation also states that they want `"images` and `"labels"`. https://docs.ultralytics.com/datasets/detect/

But I know that I have seen datasets in the past that were different, but still worked.

src/lightly_train/_data/yolo_object_detection_dataset.py
```diff
@@ -0,0 +1,123 @@
...
+    train: Path
+    val: Path
+    test: Path | None = None
+    names: dict[int, str]
+
+    @pydantic.field_validator("train", "val", mode="after")
+    def validate_paths(cls, v: Path) -> Path:
+        if "images" not in str(v):
+            raise ValueError(f"Expected path to include 'images', got {v}.")
+        if len(str(v).split("/")) != 2:
```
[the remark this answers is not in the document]
liopeer: Fixed.

src/lightly_train/_data/yolo_object_detection_dataset.py
liopeer: Done.

src/lightly_train/_transforms/task_transform.py
liopeer: Fixed.

src/lightly_train/_transforms/task_transform.py
liopeer: So, would you remove the `TaskTransformInput` and `TaskTransformOutput` classes and instead use `ObjectDetectionTransformInput` and `ObjectDetectionTransformOutput`? And then the same for semantic segmentation?
If yes, should I add those as well?

src/lightly_train/_data/yolo_object_detection_dataset.py
liopeer: Checked the documentation/code. Seems like they only handle images and masks https://albumentations.ai/docs/api-reference/albumentations/pytorch/transforms/#ToTensorV2 🤷‍♂️

src/lightly_train/_data/file_helpers.py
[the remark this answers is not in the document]
guarin: Agree with throwing errors for malformed lines

src/lightly_train/_data/yolo_object_detection_dataset.py
```diff
@@ -0,0 +1,165 @@
...
+            dataset_args.image_dir,
+            dataset_args.label_dir,
+        )
+
+    def __len__(self) -> int:
+        return len(self.image_filenames)
+
+    def __getitem__(self, index: int) -> ObjectDetectionDatasetItem:
+        assert self._image_dir is not None
+        assert self._label_dir is not None
```
guarin: Nit: This can't ever be `None` because `YOLOObjectDetectionDatasetArgs .image_dir` and `.label_dir` are typed as `Path`.  I would also recommend to use `self.args.image_dir` and `self.args.label_dir` directly instead of storing the same information twice in the class.

src/lightly_train/_data/yolo_object_detection_dataset.py
```diff
@@ -0,0 +1,165 @@
...
+    val: PathLike
+    test: PathLike | None = None
+    names: dict[int, str]
+
+    @pydantic.field_validator("train", "val", mode="after")
+    def validate_paths(cls, v: Path) -> Path:
+        if "images" not in str(v):
+            raise ValueError(f"Expected path to include 'images', got {v}.")
+        if len(v.parts) != 2:
+            raise ValueError(f"Expected subdirectories of depth 2 from root, got {v}.")
```
guarin: It cannot be deeper than 2?

src/lightly_train/_data/yolo_object_detection_dataset.py
```diff
@@ -0,0 +1,165 @@
...
+    path: PathLike
+    train: PathLike
+    val: PathLike
+    test: PathLike | None = None
+    names: dict[int, str]
+
+    @pydantic.field_validator("train", "val", mode="after")
+    def validate_paths(cls, v: Path) -> Path:
+        if "images" not in str(v):
+            raise ValueError(f"Expected path to include 'images', got {v}.")
```
guarin:
```suggestion
        if "images" not in v.parts:
            raise ValueError(f"Expected path to include 'images' directory, got {v}.")
```

src/lightly_train/_data/yolo_object_detection_dataset.py
```diff
@@ -0,0 +1,165 @@
...
+        test: Path | None,
+        mode: Literal["train", "val", "test"],
+    ) -> tuple[Path | None, Path | None]:
+        train_img_dir = path / train
+        val_img_dir = path / val
+        test_img_dir = path / test if test else None
+
+        train_label_path = Path(str(train).replace("images", "labels"))
+        val_label_path = Path(str(val).replace("images", "labels"))
+        test_label_path = Path(str(test).replace("images", "labels")) if test else None
```
guarin: This looks a bit dangerous. Did you consider replacing the first directory that is named "images" with "labels" instead of all occurrences in any part of the string?

src/lightly_train/_transforms/task_transform.py
```diff
@@ -7,20 +7,41 @@
...
+    mask: NotRequired[NDArray[np.uint8]]
+    bboxes: NotRequired[NDArray[np.float64]]
+    class_labels: NotRequired[NDArray[np.int64]]
+
+
+class TaskTransformOutput(TypedDict):
+    image: Tensor
+    mask: NotRequired[Tensor]
+    bboxes: NotRequired[Tensor]
+    class_labels: NotRequired[Tensor]
```
guarin: I think we have to split this into:

```
class TaskTransformInput(TypedDict):
    pass

class TaskTransformOutput(TypedDict):
   pass
```

And in `semantic_segmentation_transform.py`:
```
class SemanticSegmentationTransformInput(TaskTransformInput):
   image: NDArray[np.uint8]
   mask: NotRequired[NDArray[np.uint8]]
```

etc. same for object detection

src/lightly_train/_data/yolo_object_detection_dataset.py
liopeer: Done.

src/lightly_train/_data/yolo_object_detection_dataset.py
liopeer: Not stated specifically anwhere, but in every example it's exactly 2 directories deep either in the `images/train` or `train/images` format. I generally prefer keeping it more restrictive otherwise it becomes really annoying to check for all edge cases.

src/lightly_train/_data/yolo_object_detection_dataset.py
liopeer: I think if we allow only directories of depth 2 it's not dangerous. Otherwise I agree.

src/lightly_train/_transforms/task_transform.py
liopeer: We can do that, but it would make the `TaskTransform` interface a bit useless (or at least its `__call__` method). There we define the input/output as `TaskTransformInput`. I actually don't quite see why we **have** to split it.

src/lightly_train/_data/yolo_object_detection_dataset.py
liopeer: Removed.

src/lightly_train/_data/yolo_object_detection_dataset.py
liopeer: Removed the check.

src/lightly_train/_transforms/task_transform.py
```diff
@@ -7,22 +7,35 @@
...
+    pass
 
 
 class TaskTransformArgs(PydanticConfig):
+    # We use the YOLO format internally for now.
+    bbox_params: BboxParams = Field(
+        default_factory=lambda: BboxParams(
+            format="yolo", label_fields=["class_labels"]
+        ),
+    )
```
guarin: Shouldn't this be in `ObjectDetectionTransformArgs`?

src/lightly_train/_transforms/task_transform.py
```diff
@@ -35,5 +48,5 @@ def __init__(self, transform_args: TaskTransformArgs):
             )
         self.transform_args = transform_args
 
-    def __call__(self, input: TaskTransformInput) -> TaskTransformOutput:
+    def __call__(self, input: Any) -> Any:
```
guarin, line 43: Does mypy complain about the following? Functions are usually covariant in the output type so this shouldn't be a problem.

```suggestion
    def __call__(self, input: Any) -> TaskTransformOutput:
```

guarin (approved):

src/lightly_train/_transforms/task_transform.py
liopeer: This works in theory, yes. However other issues arise:
- now mypy complains that `TaskTransformOutput` does not have the keys requested in the sem seg/obj det datasets
- if I instead pass `SemanticSegmentationTransform` to the sem seg dataset, I run into issue with the `get_dataset` function in `train_task_helpers`. I would leave it for now.

src/lightly_train/_transforms/task_transform.py
liopeer: Yep. Missed that.

PR #250, opened 2025-08-18
Fix dinov2 segmentation task onnx export.

simonschoelly:
## What has changed and why?

This PR fixes the ONNX export for models trained on the DinoV2 eomt segmentation task.

There were two issues:
1. Dynamic inputs (batch size and image size) can currently not be propagated correctly trough our code. Therefore we have disabled dynamic inputs.
2. Using interpolation with antialiasing enabled cannot be exported with TorchScript (works with Dynamo though). Furthermore if "bicubic" mode is used for the interpolation with antialising enabled the corresponding `Resize` ONNX operator will have very different results than the PyTorch model. Therefore we calculate the the scaled positional encodings now during the export and store them within the model. This works only if the image size is not dynamic for the ONNX model - but as we disabled dynamic inputs anyway for ONNX this won't matter.
 

## How has it been tested?

- I added additional integration tests.
- [x] TODO will try to test with a previously trained checkpoint on our Jetson Orin. 

Yes

It does look as if we don't have the ONNX exporter in our documentation (yet).

### DIFF
--- a/src/lightly_train/_models/dinov2_vit/dinov2_vit_package.py
+++ b/src/lightly_train/_models/dinov2_vit/dinov2_vit_package.py
@@ -111,6 +111,7 @@ class DINOv2ViTPackage(Package):
             ffn_bias=cfg.student.ffn_bias,
             num_register_tokens=cfg.student.num_register_tokens,
             interpolate_offset=cfg.student.interpolate_offset,
+            interpolate_antialias=cfg.student.interpolate_antialias,
             drop_path_rate=cfg.student.drop_path_rate,
             drop_path_uniform=cfg.student.drop_path_uniform,
         )
--- a/src/lightly_train/_models/dinov2_vit/dinov2_vit_src/layers/patch_embed.py
+++ b/src/lightly_train/_models/dinov2_vit/dinov2_vit_src/layers/patch_embed.py
@@ -80,6 +80,7 @@ class PatchEmbed(nn.Module):
         new_W = math.ceil(W / patch_W) * patch_W
 
         if new_H != H or new_W != W:
+            # Resize image to nearest valid resolution
             x = F.interpolate(
                 x, size=(new_H, new_W), mode="bicubic", align_corners=False
             )


src/lightly_train/_models/dinov2_vit/dinov2_vit_src/layers/patch_embed.py
```diff
@@ -81,8 +81,10 @@ def forward(self, x: Tensor) -> Tuple[Tensor, int, int]:
 
         if new_H != H or new_W != W:
             # Resize image to nearest valid resolution
+            # NOTE: Due to a bug, image sizes that are not a multiple of the patch size do not work so this
```
stegmuel: Can you document the bug?

tests/_commands/test_train_task.py
```diff
@@ -112,3 +112,108 @@ def test_train_semantic_segmentation(tmp_path: Path) -> None:
             onnx_masks, onnx_logits = ort_session.run(["masks", "logits"], ort_inputs)
             assert onnx_masks.shape == (1, 224, 224)
             assert onnx_logits.shape == (1, 2, 224, 224)
+
+
+@pytest.fixture(scope="module")
+def dinov2_vits14_eomt_checkpoint(tmp_path_factory: pytest.TempPathFactory) -> Path:
```
yutong-xiang-97: Could this be consolidated with the prev test which also calls `lightly_train.train_semantic_segmentation` since the checkpoint path is the only thing that matters here? This way we only do the call once.

Also, if possible, I would prefer to have all the tests for onnx in a separate `test_export_task.py` so that we don't have the tests clutter within one test file.

simonschoelly: Yes, I agree with that. I did not want to mess too much with the existing file hierarchy but I will now move the onnx tests to a separate file.

src/lightly_train/_models/dinov2_vit/dinov2_vit_src/layers/patch_embed.py
simonschoelly: Done

src/lightly_train/_models/dinov2_vit/dinov2_vit_src/layers/patch_embed.py
```diff
@@ -80,7 +80,6 @@ def forward(self, x: Tensor) -> Tuple[Tensor, int, int]:
         new_W = math.ceil(W / patch_W) * patch_W
 
         if new_H != H or new_W != W:
-            # Resize image to nearest valid resolution
             x = F.interpolate(
```
stegmuel: Why is the comment not relevant anymore?

src/lightly_train/_models/dinov2_vit/dinov2_vit_src/models/vision_transformer.py
```diff
@@ -222,40 +224,60 @@ def init_weights(self):
...
-        M = int(math.sqrt(N))  # Recover the number of patches in each dimension
-        assert N == M * M
-        kwargs = {}
-        if self.interpolate_offset:
-            # Historical kludge: add a small number to avoid floating point error in the interpolation, see https://github.com/facebookresearch/dino/issues/8
-            # Note: still needed for backward-compatibility, the underlying operators are using both output size and scale factors
-            sx = float(w0 + self.interpolate_offset) / M
-            sy = float(h0 + self.interpolate_offset) / M
-            kwargs["scale_factor"] = (sx, sy)
+            result = self.pos_embed.to(previous_dtype)
```
stegmuel, line 247: Can you explain why you need the cast to `previous_dtype`?

src/lightly_train/_models/dinov2_vit/dinov2_vit_package.py
```diff
@@ -111,7 +111,6 @@ def get_model(
             ffn_bias=cfg.student.ffn_bias,
             num_register_tokens=cfg.student.num_register_tokens,
             interpolate_offset=cfg.student.interpolate_offset,
-            interpolate_antialias=cfg.student.interpolate_antialias,
             drop_path_rate=cfg.student.drop_path_rate,
```
stegmuel: Can't we still pass the value from the config and only do things directly at exportation. This way we are sure we don't affect the behaviour.
It seems to me you only do the interpolation in plain torch before doing the exportation. So I don't why you need to change the antialias value.

stegmuel (commented):
Left some comments/questions.
I think the description of the PR is outdated. Can you update it, please?

src/lightly_train/_models/dinov2_vit/dinov2_vit_package.py
simonschoelly: I thought I had reverted that in all places but somehow this slipped by. Will fix it.

src/lightly_train/_models/dinov2_vit/dinov2_vit_src/layers/patch_embed.py
simonschoelly: Same here - I thought I had reverted that too.

src/lightly_train/_models/dinov2_vit/dinov2_vit_src/models/vision_transformer.py
simonschoelly: For consistency - In case we have to scale positional encodings we  also cast the result to the dtype of `x`.

stegmuel (approved):
LGTM.

PR #251, opened 2025-08-19
Add linear semantic segmentation model

guarin:
## What has changed and why?

* Add linear semantic segmentation model

Didn't add documentation page yet as this is not one of the big features. We'll see whether it makes sense to add a full page for them. The models are useful to evaluate the quality of a pretrained DINOv2 model as the linear segmentation uses a frozen backbone by default.

## How has it been tested?

* Manual

I got 81.2% mIoU on PVOC12 with default settings.

- [x] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/task_model.py
+++ b/src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/task_model.py
@@ -49,7 +49,7 @@ class DINOv2LinearSemanticSegmentation(TaskModel):
         """
         Args:
             model_name:
-                The model name. For example "vits14-linear".
+                The model name. For example "dinov2/vits14-linear".
             classes:
                 A dict mapping the class ID to the class name. The dict must only
                 contain the classes that the model should predict. It must NOT contain
--- a/src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/task_model.py
+++ b/src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/task_model.py
@@ -141,7 +141,7 @@ class DINOv2LinearSemanticSegmentation(TaskModel):
                 "more information: https://docs.lightly.ai/train/stable/semantic_segmentation.html"
             )
 
-        if model_name.endswith(f"-{cls.model_suffix}"):
+        if not model_name.endswith(f"-{cls.model_suffix}"):
             raise_invalid_name()
 
         backbone_name = model_name[: -len(f"-{cls.model_suffix}")]
--- a/src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/task_model.py
+++ b/src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/task_model.py
@@ -145,8 +145,7 @@ class DINOv2LinearSemanticSegmentation(TaskModel):
         def raise_invalid_name() -> None:
             raise ValueError(
                 f"Model name '{model_name}' is not supported. Available "
-                f"models are: {cls.list_model_names()}. See the documentation for "
-                "more information: https://docs.lightly.ai/train/stable/semantic_segmentation.html"
+                f"models are: {cls.list_model_names()}."
             )
 
         if not model_name.endswith(f"-{cls.model_suffix}"):
--- a/src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/task_model.py
+++ b/src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/task_model.py
@@ -1,0 +1,381 @@
+#
+# Copyright (c) Lightly AG and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+#
+from __future__ import annotations
+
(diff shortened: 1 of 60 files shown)


src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/task_model.py
```diff
@@ -0,0 +1,387 @@
...
+from lightly_train._models.dinov2_vit.dinov2_vit_src.models.vision_transformer import (
+    DinoVisionTransformer,
+)
+from lightly_train._task_models.task_model import TaskModel
+from lightly_train.types import PathLike
+
+logger = logging.getLogger(__name__)
+
+
+class DINOv2LinearSemanticSegmentation(TaskModel):
```
stegmuel, line 34: The naming suggests that if we were to support linear seg. on top of DINOv3 we would re-create a whole new class, is this intended? If not should we change the name or add a TODO to change it?

src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/task_model.py
```diff
@@ -0,0 +1,387 @@
...
+
+        # Disable drop path by default.
+        args = {
+            "drop_path_rate": 0.0,
+        }
+        if backbone_args is not None:
+            args.update(backbone_args)
+
+        # Get the backbone.
+        self.backbone: DinoVisionTransformer = DINOV2_VIT_PACKAGE.get_model(
```
stegmuel, line 105: Depending on your answer to the above comment, add TODO to suportt DINOv3 backbones.

src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/task_model.py
```diff
@@ -0,0 +1,387 @@
...
+
+        # Up-sample to match original image/mask resolution.
+        # (B, K|K+1, H, W)
+        logits = F.interpolate(
+            logits, size=(H, W), mode="bilinear", align_corners=False
+        )
+
+        return logits
+
+    def tile(
```
stegmuel, line 253: If the tile/untile methods are identical for all segmentation methods, should we consider having a SegmentationTaskModel class and move these there.
If so can you add a TODO?

src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/task_model.py
```diff
@@ -0,0 +1,387 @@
...
+        x = transforms_functional.normalize(
+            x, mean=self.image_normalize["mean"], std=self.image_normalize["std"]
+        )
+        # Resize to configured image size
+        x = transforms_functional.resize(
+            x, size=list(self.image_size)
+        )  # (C, H, W) -> (C, H', W')
+        x = x.unsqueeze(0)  # (1, C, H', W')
+
+        logits = self._forward_logits(x)  # (1, K|K+1, H', W'), K=num_classes
```
stegmuel, line 215: In this scenario, is it possible to have K+1 ?

src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/train_model.py
```diff
@@ -0,0 +1,229 @@
...
+            "val_loss": loss.detach(),
+            "val_metric/miou": self.val_miou,
+        }
+        if self.model_args.metric_log_debug or self.model_args.metric_log_classwise:
+            log_dict["val_metric_classwise/miou"] = self.val_classwise_iou
+
+        return TaskStepResult(loss=loss, log_dict=log_dict)
+
+    def get_optimizer(self, total_steps: int) -> tuple[Optimizer, LRScheduler]:
+        params, params_no_weight_decay = get_weight_decay_parameters([self])
```
stegmuel: Will the optimizer also store the weigths of the backbone?
If so, and we intend to always use this class with a frozen backbone then it will lead to a large memory overhead, no?

src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/task_model.py
guarin: For now yes. We would have to refactor DINOv2/v3 a bit to make this possible.

src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/task_model.py
guarin: We can move them as functions into a shared helpers file. They don't have to be methods as they don't depend on the class.

src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/task_model.py
guarin: Yes as `self.head` depends on whether `class_ignore_index` is set or not.

src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/train_model.py
guarin: Updated

PR #254, opened 2025-08-20
Memory Map Semantic Segmentation Dataset

yutong-xiang-97:
## What has changed and why?

Add mmap to semantic segmentation dataset which handles the following:

- User wants to reuse the mmap file by setting up an env var -> reuse the existing `mmap_filepath`
- If not, the rank 0 process creates the mmap file, other processes wait until the file is available
- If the processes are not using a shared file system in the multi-node setting, creates the mmap file with the local rank zero process on each node.

The current implementation is strongly based on the assumption that the mask images have identical names to their corresponding images. We will need to revisit the implementation once we relax the assumption.

I also wrote some unit tests that pass but decided to not add them here as the interface may be subjected to change.

## How has it been tested?

N/A. See the comments about the tests above.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -283,14 +283,12 @@ def get_dataset_temp_mmap_path(
     try:
         # Delete the file if it already exists from a previous run.
         if not reuse_file and (fabric.local_rank == 0):
-            _unlink_and_ignore(mmap_filepath)
+            _unlink_and_ignore(mmap_filepath_broadcasted)
 
-        yield mmap_filepath
+        yield mmap_filepath_broadcasted
     finally:
         if not reuse_file and (fabric.local_rank == 0):
-            _unlink_and_ignore(mmap_filepath)
-
-    yield mmap_filepath_broadcasted
+            _unlink_and_ignore(mmap_filepath_broadcasted)
 
 
 def get_dataset_mmap_filenames(


src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -222,14 +224,70 @@ def get_val_transform(
...
+        logger.warning(f"Reusing existing memory-mapped file '{mmap_filepath}'.")
+        return memory_mapped_sequence.memory_mapped_sequence_from_file(
+            mmap_filepath=mmap_filepath
+        )
+
+    with fabric.rank_zero_first():
+        memory_mapped_sequence.write_filenames_to_file(
+            filenames=filenames,
+            mmap_filepath=mmap_filepath,
+        )
```
guarin:
```suggestion
        if (fabric.global_rank == 0) or (is_shared_filesystem and fabric.local_rank == 0):
          memory_mapped_sequence.write_filenames_to_file(
              filenames=filenames,
              mmap_filepath=mmap_filepath,
          )
```

You also have to move the `is_shared_filesystem` check above the `write_filenames_to_file` section

src/lightly_train/_commands/train_task.py
```diff
@@ -234,195 +234,213 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
...
-        transform=val_transform,
-    )
-    logger.info(f"Train images: {len(train_dataset)}, Val images: {len(val_dataset)}")
-
-    train_model_args_cls = train_model_cls.train_model_args_cls
+    with common_helpers.get_dataset_temp_mmap_path(
+        data=config.data.train.images
+    ) as train_mmap_filepath, common_helpers.get_dataset_temp_mmap_path(
+        data=config.data.val.images
+    ) as val_mmap_filepath:
```
guarin: Don't use `common_helpers`, write a new function in `train_task_helpers` that follows the logic of `train_task_helpers.py:get_out_dir`

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -222,14 +224,70 @@ def get_val_transform(
...
+        is_shared_filesystem = False
+
+    # If the filesystem is not shared we have to create the output directory on every
+    # node individually.
+    if not is_shared_filesystem:
+        with fabric.rank_zero_first(local=True):
+            if fabric.local_rank == 0 and fabric.global_rank != 0:
+                memory_mapped_sequence.write_filenames_to_file(
+                    filenames=filenames,
+                    mmap_filepath=mmap_filepath,
```
guarin: Can be skipped if you add changes from comment above

yutong-xiang-97: The actual mmap file is not created until we call `memory_mapped_sequence.write_filenames_to_file`. However, if the parent dir is created with `get_dataset_temp_mmap_path` we can use the parent directory to check `is_shared_filesystem`

src/lightly_train/_commands/train_task_helpers.py
yutong-xiang-97: Should be `not is_shared_filesystem`

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -222,14 +227,109 @@ def get_val_transform(
...
+        is_shared_filesystem = False
+
+    # If the filesystem is not shared we have to create the mmap file on every
+    # node individually.
+    if not is_shared_filesystem:
+        with fabric.rank_zero_first(local=True):
+            if fabric.local_rank == 0 and fabric.global_rank != 0:
+                mmap_dirpath_broadcasted.mkdir(parents=True, exist_ok=True)
+
+    yield mmap_filepath_broadcasted
```
guarin: Needs automatic cleanup handling:

```
    reuse_file = Env.LIGHTLY_TRAIN_MMAP_REUSE_FILE.value
    try:
        # Delete the file if it already exists from a previous run.
        if not reuse_file and distributed_helpers.is_local_rank_zero():
            _unlink_and_ignore(mmap_filepath)

        yield mmap_filepath
    finally:
        if not reuse_file and distributed_helpers.is_local_rank_zero():
            _unlink_and_ignore(mmap_filepath)
```

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -222,14 +227,109 @@ def get_val_transform(
...
+        is_shared_filesystem = fabric_utilities.is_shared_filesystem(
+            strategy=fabric.strategy, path=mmap_filepath.parent
+        )
+    except FileNotFoundError:
+        # Clearly not a shared filesystem because we just created the parent directory.
+        is_shared_filesystem = False
+
+    # If the filesystem is not shared we have to create the mmap file on every
+    # node individually.
+    with fabric.rank_zero_first():
```
guarin:
```suggestion
    with fabric.rank_zero_first(local=True):
```

guarin (changes requested):

guarin (approved):

PR #257, opened 2025-08-22
Fix race condition when training multiple models

yutong-xiang-97:
## What has changed and why?

Use the file lock mechanism to fix 

- the race conditions of using mmap files when launching multiple training runs on the same data and with the same node. 
- the race condition of using tmp file in `get_dataset_mmap_filenames`?

Also added corresponding tests.

## How has it been tested?

Manually launching multiple MLflow runs and unit tests.

- [ ] Yes

- [ ] Yes


src/lightly_train/_commands/common_helpers.py
```diff
@@ -391,18 +392,82 @@ def get_dataset_temp_mmap_path(
...
     mmap_filepath.parent.mkdir(parents=True, exist_ok=True)
 
-    reuse_file = Env.LIGHTLY_TRAIN_MMAP_REUSE_FILE.value
     try:
-        # Delete the file if it already exists from a previous run.
-        if not reuse_file and distributed_helpers.is_local_rank_zero():
-            _unlink_and_ignore(mmap_filepath)
+        # Increment reference count atomically
+        if distributed_helpers.is_local_rank_zero():
+            _increment_ref_count(ref_count_filepath)
```
guarin: Should all processes increase the refcount and not only rank=0 as all processes need to access the file?

src/lightly_train/_commands/common_helpers.py
```diff
@@ -391,18 +392,82 @@ def get_dataset_temp_mmap_path(
...
 
         yield mmap_filepath
     finally:
-        if not reuse_file and distributed_helpers.is_local_rank_zero():
-            _unlink_and_ignore(mmap_filepath)
+        # Decrement reference count and cleanup if zero
+        if distributed_helpers.is_local_rank_zero():
+            _decrement_and_cleanup_if_zero(mmap_filepath, ref_count_filepath)
```
guarin: Same as above

src/lightly_train/_commands/common_helpers.py
```diff
@@ -391,18 +392,82 @@ def get_dataset_temp_mmap_path(
...
+
+
+def _increment_ref_count(ref_file: Path) -> None:
+    lock_file = ref_file.with_suffix(".lock")
+
+    # Retry logic for Python 3.8 robustness
+    max_retries = 10
+    for attempt in range(max_retries):
+        try:
+            with FileLock(lock_file):
```
guarin: Always pass a timeout when acquiring a lock.
```suggestion
            with FileLock(lock_file).acquire(timeout=300):
```
Then you should also be able to remove the retry logic.

src/lightly_train/_commands/common_helpers.py
```diff
@@ -391,18 +392,82 @@ def get_dataset_temp_mmap_path(
...
+            try:
+                with FileLock(lock_file):
+                    with open(ref_file, "r+") as f:
+                        count = int(f.read() or "1") - 1
+                        if count <= 0:
+                            should_cleanup = True
+                        else:
+                            f.seek(0)
+                            f.write(str(count))
+                            f.truncate()
```
guarin:
```suggestion
                with FileLock(lock_file):
                    with open(ref_file, "r+") as f:
                        count = int(f.read() or "1") - 1
                        f.seek(0)
                        f.write(str(count))
                        f.truncate()
                        if count <= 0:
                            should_cleanup = True
                    if not Env.LIGHTLY_TRAIN_MMAP_REUSE_FILE.value:
                        # Remove mmap file only if we are not reusing it
                        _unlink_and_ignore(mmap_file)  
```
You have to unlink and ignore the files within the same `with FileLock(...)` statement. Once you release the lock the `should_cleanup` condition might not be true anymore as another process might already change something. I would also not remove the ref count file as it uses zero space on the disk.

src/lightly_train/_commands/common_helpers.py
```diff
@@ -391,18 +392,82 @@ def get_dataset_temp_mmap_path(
...
+            _unlink_and_ignore(ref_file)
+            if not Env.LIGHTLY_TRAIN_MMAP_REUSE_FILE.value:
+                # Remove mmap file only if we are not reusing it
+                _unlink_and_ignore(mmap_file)
+    except (FileNotFoundError, OSError):
+        pass  # Another process already cleaned up
+    finally:
+        # Clean up lock file if ref_file no longer exists (cleanup completed)
+        if not ref_file.exists():
+            _unlink_and_ignore(lock_file)
```
guarin:
```suggestion
```
If you use locks then you you should only modify files within the lock context as anything can happen between the `if not ref_file.exists()` and `unlink_and_ignore(lock_file)` statements. I would actually not cleanup the `.lock` and `.ref_counter` files as they don't use any space on the disk.

src/lightly_train/_commands/common_helpers.py
```diff
@@ -391,18 +392,82 @@ def get_dataset_temp_mmap_path(
...
     finally:
-        if not reuse_file and distributed_helpers.is_local_rank_zero():
-            _unlink_and_ignore(mmap_filepath)
+        # Decrement reference count and cleanup if zero
+        if distributed_helpers.is_local_rank_zero():
+            _decrement_and_cleanup_if_zero(mmap_filepath, ref_count_filepath)
+
+
+def _increment_ref_count(ref_file: Path) -> None:
+    lock_file = ref_file.with_suffix(".lock")
```
guarin, line 410: Should we pass the lock file to the function? Then it is clear that the same lock is used in the increment and decrement functions.

src/lightly_train/_commands/common_helpers.py
yutong-xiang-97: The race condition is between the multiple training runs that on the same node and (possibly) use the same GPU, not different processes within one training run. Then it makes sense that the main process for every run handle the ref count.

src/lightly_train/_commands/common_helpers.py
yutong-xiang-97: I don't think we need to. The lock file path is the same and we only do locking when we need to increment / decrement the reference counts. No need to hold the locks all the time.

src/lightly_train/_commands/common_helpers.py
guarin: Yes but all processes access the mmap file. I think it would be great to have a clear link between the refcount and whether the mmap file is needed. --> If ref count > 0 then some process needs it.

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -312,16 +314,50 @@ def get_dataset_temp_mmap_path(
...
                 mmap_dirpath_broadcasted.mkdir(parents=True, exist_ok=True)
 
-    reuse_file = Env.LIGHTLY_TRAIN_MMAP_REUSE_FILE.value
     try:
-        # Delete the file if it already exists from a previous run.
-        if not reuse_file and (fabric.local_rank == 0):
-            _unlink_and_ignore(mmap_filepath_broadcasted)
+        # Increment reference count atomically
+        if fabric.local_rank == 0:
+            _increment_ref_count(ref_count_filepath_broadcasted)
```
guarin: Same as https://github.com/lightly-ai/lightly-train/pull/257/files#r2300427426

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -312,16 +314,50 @@ def get_dataset_temp_mmap_path(
...
 
         yield mmap_filepath_broadcasted
     finally:
-        if not reuse_file and (fabric.local_rank == 0):
-            _unlink_and_ignore(mmap_filepath_broadcasted)
+        # Decrement reference count and cleanup if zero
+        if fabric.local_rank == 0:
+            _decrement_and_cleanup_if_zero(
+                mmap_filepath_broadcasted, ref_count_filepath_broadcasted
+            )
```
guarin: Same as https://github.com/lightly-ai/lightly-train/pull/257/files#r2300427426

src/lightly_train/_commands/common_helpers.py
```diff
@@ -391,21 +392,56 @@ def get_dataset_temp_mmap_path(
...
+            f.truncate()
+
+
+def _decrement_and_cleanup_if_zero(mmap_file: Path, ref_file: Path) -> None:
+    try:
+        lock_file = ref_file.with_suffix(".lock")
+
+        with FileLock(lock_file, timeout=300):
+            with open(ref_file, "r+") as f:
+                count = int(f.read() or "1") - 1
```
guarin: Count must never go below 0. A negative state would be an impossible state and must be made impossible.

```suggestion
                count = max(0, int(f.read() or "1") - 1)
```

tests/_commands/test_common_helpers.py
```diff
@@ -540,11 +572,143 @@ def test_get_num_workers__slurm(
...
+    # Verify final count is correct
+    assert ref_file.read_text() == str(num_increments)
+
+
+@pytest.mark.parametrize(
+    "initial_count,num_decrements,should_cleanup",
+    [
+        (10, 3, False),  # 10 - 3 = 7, no cleanup
+        (5, 5, True),  # 5 - 5 = 0, cleanup
+        (3, 8, True),  # 3 - 8 < 0, cleanup
```
guarin: Expected count must be 0 for `(3, 8, True)`

tests/_commands/test_common_helpers.py
```diff
@@ -563,10 +727,35 @@ def test_get_dataset_temp_mmap_path__rank(
...
+    data_path = tmp_path / "data"
+
+    # Run 5 concurrent context managers (processes)
+    with concurrent.futures.ProcessPoolExecutor(max_workers=5) as executor:
+        futures = [executor.submit(_ctx_mmap_worker, str(data_path)) for _ in range(5)]
+        # Collect results and re-raise any exceptions
+        mmap_paths = [Path(future.result()) for future in futures]
+
+    # Verify all threads got the same mmap path
+    assert len(set(mmap_paths)) == 1, "All threads should get the same mmap path"
```
guarin:
```suggestion
    # Verify all processes got the same mmap path
    assert len(set(mmap_paths)) == 1, "All processes should get the same mmap path"
```

tests/_commands/test_common_helpers.py
```diff
@@ -563,10 +727,35 @@ def test_get_dataset_temp_mmap_path__rank(
...
+
+    data_hash = common_helpers.get_sha256(f"{data_path}-0")
+    mmap_path = (cache.get_data_cache_dir() / data_hash).with_suffix(".mmap")
+
+    # After all context managers exit, the mmap file should be cleaned up
+    assert not mmap_path.exists()
```
guarin:
```suggestion
    assert not mmap_paths[0].exists()
```

PR #258, opened 2025-08-22
Add fine-tune transform arg customization

guarin:
## What has changed and why?

* Add fine-tune transform arg customization
* Cleanup transform code
* Remove unused transforms
* Add documentation

## How has it been tested?

Train log from a run:
```
[2025-08-22 13:38:38,965][DEBUG] Resolved train transform args {
    "color_jitter": {
        "brightness": 0.12549019607843137,
        "contrast": 0.5,
        "hue": 0.05,
        "prob": 0.5,
        "saturation": 0.5,
        "strength": 1.0
    },
    "ignore_index": -100,
    "image_size": [
        512,
        512
    ],
    "normalize": {
        "mean": [
            0.0,
            0.0,
            0.0
        ],
        "std": [
            1.0,
            1.0,
            1.0
        ]
    },
    "random_crop": {
        "fill": 0,
        "height": 512,
        "pad_if_needed": true,
        "pad_position": "center",
        "prob": 1.0,
        "width": 512
    },
    "random_flip": {
        "horizontal_prob": 0.5,
        "vertical_prob": 0.0
    },
    "scale_jitter": {
        "max_scale": 2.0,
        "min_scale": 0.5,
        "num_scales": 20,
        "prob": 1.0
    },
    "smallest_max_size": null
}
[2025-08-22 13:38:38,965][DEBUG] Resolved val transform args {
    "color_jitter": null,
    "ignore_index": -100,
    "image_size": [
        512,
        512
    ],
    "normalize": {
        "mean": [
            0.0,
            0.0,
            0.0
        ],
        "std": [
            1.0,
            1.0,
            1.0
        ]
    },
    "random_crop": null,
    "random_flip": null,
    "scale_jitter": null,
    "smallest_max_size": {
        "max_size": 512,
        "prob": 1.0
    }
}
```

Docs:

[Semantic Segmentation - LightlyTrain documentation transform args.pdf](https://github.com/user-attachments/files/21938627/Semantic.Segmentation.-.LightlyTrain.documentation.transform.args.pdf)

- [x] Yes

- [x] Yes

### DIFF
--- a/docs/prebuild.py
+++ b/docs/prebuild.py
@@ -6,12 +6,17 @@ from argparse import ArgumentParser
 from pathlib import Path
 from re import Match
 
-from lightly_train._commands import common_helpers, train_helpers
+from lightly_train._commands import common_helpers, train_helpers, train_task_helpers
+from lightly_train._commands.train_task_helpers import TASK_TRAIN_MODEL_CLASSES
 from lightly_train._methods import method_helpers
+from lightly_train._task_models.dinov2_linear_semantic_segmentation.train_model import (
+    DINOv2LinearSemanticSegmentationTrain,
+)
 
 THIS_DIR = Path(__file__).parent.resolve()
 DOCS_DIR = THIS_DIR / "source"
 PROJECT_ROOT = THIS_DIR.parent
+SOURCE_AUTO_DIR = DOCS_DIR / "_auto"
 METHODS_AUTO_ARGS_DIR = DOCS_DIR / "methods" / "_auto"
 
 
@@ -81,6 +86,23 @@ def dump_transform_args_for_methods(dest_dir: Path) -> None:
             f.write("```\n")
 
 
+def dump_transform_args_for_tasks(dest_dir: Path) -> None:
+    dest_dir.mkdir(parents=True, exist_ok=True)
+    for train_model_cls in TASK_TRAIN_MODEL_CLASSES:
+        if train_model_cls in {DINOv2LinearSemanticSegmentationTrain}:
+            continue
+        train_transform_args = train_model_cls.train_transform_cls.transform_args_cls(
+            ignore_index=-100
+        )
+        args = train_task_helpers.pretty_format_args(train_transform_args.model_dump())
+        name = train_model_cls.__name__.lower()
+        # write to file
+        with open(dest_dir / f"{name}_train_transform_args.md", "w") as f:
+            f.write("```json\n")
+            f.write(args + "\n")
+            f.write("```\n")
+
+
 def dump_method_args(dest_dir: Path) -> None:
     dest_dir.mkdir(parents=True, exist_ok=True)
     # dump transform args for all methods
@@ -98,8 +120,11 @@ def dump_method_args(dest_dir: Path) -> None:
 
 def main(source_dir: Path) -> None:
     build_changelog_html(source_dir=source_dir)
+    # Methods
     dump_transform_args_for_methods(dest_dir=METHODS_AUTO_ARGS_DIR)
     dump_method_args(dest_dir=METHODS_AUTO_ARGS_DIR)
+    # Tasks
+    dump_transform_args_for_tasks(dest_dir=SOURCE_AUTO_DIR)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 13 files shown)


docs/prebuild.py
```diff
@@ -81,6 +86,23 @@ def dump_transform_args_for_methods(dest_dir: Path) -> None:
             f.write("```\n")
 
 
+def dump_transform_args_for_tasks(dest_dir: Path) -> None:
+    dest_dir.mkdir(parents=True, exist_ok=True)
+    for train_model_cls in TASK_TRAIN_MODEL_CLASSES:
+        if train_model_cls in {DINOv2LinearSemanticSegmentationTrain}:
+            continue
+        train_transform_args = train_model_cls.train_transform_cls.transform_args_cls(
+            ignore_index=-100
```
liopeer: If `-100` is supposed to be the default, we should not hardcode it here, but actually put it as. the default in `SemanticSegmentationTransformArgs`. Otherwise these 2 will for sure diverge.

docs/source/_auto/dinov2eomtsemanticsegmentationtrain_train_transform_args.md
```diff
@@ -0,0 +1,48 @@
+```json
```
liopeer: I would not commit anything in the `docs/source/_auto` directory. We also don't do it for `docs/source/methods/_auto`. Same for the next file.

docs/source/semantic_segmentation.md
```diff
@@ -367,3 +367,59 @@ if __name__ == "__main__":
...
+            "classes": {                                # Classes in the dataset                    
+                0: "background",
+                1: "car",
+                2: "bicycle",
+                # ...
+            },
+            # Optional, classes that are in the dataset but should be ignored during
+            # training.
+            "ignore_classes": [0], 
+        },
```
liopeer, line 406:
```suggestion
        data=data_args
```
Would prefer not repeating everything from further above. But if the linter complains we can leave as is.

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -192,25 +195,42 @@ def pretty_format_args_dict(args: dict[str, Any]) -> dict[str, Any]:
...
+    train_transform_args = validate.pydantic_model_validate(
+        train_transform_args_cls, transform_args
+    )
+    train_transform_args.resolve_auto()
 
-    return train_transform_args_cls(ignore_index=ignore_index), val_transform_args_cls(  # type: ignore[call-arg]
-        ignore_index=ignore_index
+    val_transform_args = validate.pydantic_model_validate(
+        val_transform_args_cls,
+        train_transform_args.model_dump(
```
liopeer, line 221: I would maybe put this into a separate helper function `get_val_transform_args_from_train_transform_args` or something like that. Doesn't necessarily have to be in this PR though. But we will definitely need this in other places as well.

src/lightly_train/_transforms/semantic_segmentation_transform.py
```diff
@@ -11,11 +11,9 @@
 import numpy as np
 from albumentations import (
     BasicTransform,
-    CenterCrop,
     ColorJitter,
     Compose,
     HorizontalFlip,
-    LongestMaxSize,
```
liopeer, line 18: Would be great to have some context on why we can remove these.

docs/source/semantic_segmentation.md
guarin: We always use full examples to make the code runnable without having to read the full page first or copy pasting snippets from different places together.

src/lightly_train/_commands/train_task_helpers.py
guarin: We can add a `get_val_transform_args` function once it is needed 👍🏼

src/lightly_train/_transforms/semantic_segmentation_transform.py
guarin: They are never used. Sorry forgot to add a comment about this. You can check that we had everywhere:

```
class XYZTransformArgs:
    center_crop: CenterCropArgs | None = None
```

Same for longest max size.

liopeer (approved):
LGTM! Thanks for the changes.

PR #262, opened 2025-08-25
Add verify option to validate exported onnx model

simonschoelly:
## What has changed and why?

Added a verify option to the onnx export task to validate the exported model. This option does two things:
- Use onnx.checker.check_model to verify the exported models.
- Run a random tensor trough both the exported model and the torch model. 

The second part could probably be made arbitrary complicated - but for not I just test with random data and use `torch.testing.assert_close` as that function seems to be able to also handle `NaN`, `+Inf` and `-Inf`.

## How has it been tested?

As the verify option is enabled by default it is covered by the tests anyway.

- [] Is this change big enough that we want in CHANGELOG.md?

### DIFF
--- a/src/lightly_train/_commands/export_task.py
+++ b/src/lightly_train/_commands/export_task.py
@@ -167,7 +167,7 @@ def _export_task_from_config(config: ExportTaskConfig) -> None:
 
             x = torch.rand_like(dummy_input)
             session = ort.InferenceSession(out_path)
-            input_feed = {input_name: x.numpy()}
+            input_feed = {input_name: x.cpu().numpy()}
             outputs_onnx = session.run(output_names=output_names, input_feed=input_feed)
             outputs_onnx = tuple(torch.from_numpy(y) for y in outputs_onnx)
 


src/lightly_train/_commands/export_task.py
```diff
@@ -140,17 +144,58 @@ def _export_task_from_config(config: ExportTaskConfig) -> None:
...
+        if config.verify:
+            logger.info("Verifying ONNX model")
+            import onnx
+            import onnxruntime as ort
+
+            onnx.checker.check_model(out_path, full_check=True)
+
+            x = torch.rand_like(dummy_input)
+            session = ort.InferenceSession(out_path)
+            input_feed = {input_name: x.numpy()}
```
guarin:
```suggestion
            input_feed = {input_name: x.cpu().numpy()}
```

As `x` might be on GPU/MPS/... because it is created on the same device as `dummy_input`.

simonschoelly: Thanks. I changed it for now - there will be some changes to that line later when I add the logic to specify the resolution.

PR #263, opened 2025-08-26
Add multi-channel image loading and channel drop transform for fine tuning

guarin:
## What has changed and why?

* Add multi-channel image loading for fine-tuning
* Add channel drop augmentation for fine-tuning

Usage (assuming 4-channel images):
```python
lightly_train.train_semantic_segmentation(
    out=out,
    data={
        "train": {
            "images": data / "train" / "images",
            "masks": data / "train" / "masks",
        },
        "val": {
            "images": data / "val" / "images",
            "masks": data / "val" / "masks",
        },
        "classes": {
            0: "background",
            1: "Solar-panel",
            2: "diode",
            3: "hotspot",
        },
        "ignore_classes": [0],
    },
    model="dinov2/vits16-eomt",
    transform_args={
        "channel_drop": {
            "num_channels_keep": 3,
            "weight_drop": [1.0, 1.0, 0.0, 0.0], # Convert NRGB to NGB or RGB randomly.
        },
        "val": {
            "channel_drop": {
                "num_channels_keep": 3,
                "weight_drop": [1.0, 0.0, 0.0, 0.0], # Convert NRGB to RGB.
            }
        },
    },
)
``` 

Then run with `LIGHTLY_TRAIN_IMAGE_MODE="UNCHANGED" python train.py`

## How has it been tested?

* Manually, tested EoMT and linear models

[Uploading Semantic Segmentation - LightlyTrain documentation multichannel.pdf…]()

- [x] Yes

- [x] Yes

### DIFF
--- a/docs/source/semantic_segmentation.md
+++ b/docs/source/semantic_segmentation.md
@@ -414,6 +414,28 @@ if __name__ == "__main__":
     )
 ```
 
+`````{dropdown} EoMT DINOv2 Default Transform Arguments
+````{dropdown} Train
+```{include} _auto/dinov2eomtsemanticsegmentationtrain_train_transform_args.md
+```
+````
+````{dropdown} Val
+```{include} _auto/dinov2eomtsemanticsegmentationtrain_val_transform_args.md
+```
+````
+`````
+
+`````{dropdown} EoMT DINOv3 Default Transform Arguments
+````{dropdown} Train
+```{include} _auto/dinov3eomtsemanticsegmentationtrain_train_transform_args.md
+```
+````
+````{dropdown} Val
+```{include} _auto/dinov3eomtsemanticsegmentationtrain_val_transform_args.md
+```
+````
+`````
+
 In case you need different parameters for training and validation, you can pass an
 optional `val` dictionary to `transform_args` to override the validation parameters:
 
@@ -429,13 +451,3 @@ transform_args={
     }
 }
 ```
-
-````{dropdown} EoMT DINOv2 Default Train Transform Arguments
-```{include} _auto/dinov2eomtsemanticsegmentationtrain_train_transform_args.md
-```
-````
-
-````{dropdown} EoMT DINOv3 Default Train Transform Arguments
-```{include} _auto/dinov3eomtsemanticsegmentationtrain_train_transform_args.md
-```
-````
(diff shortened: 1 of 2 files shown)


docs/source/semantic_segmentation.md
```diff
@@ -414,6 +414,22 @@ if __name__ == "__main__":
     )
 ```
 
+In case you need different parameters for training and validation, you can pass an
```
liopeer, line 439: This is a bit unclear, since training/validation parameters **will/should always** be different. I would actually move things around a bit on this page:
```markdown

## Default Image Transform Arguments
The following are the default train transform arguments for EoMT. The validation
arguments are automatically inferred from the train arguments. Specifically the
image size and normalization are shared between train and validation.

````{dropdown} EoMT DINOv2 Default Train Transform Arguments
```{include} _auto/dinov2eomtsemanticsegmentationtrain_train_transform_args.md
```
````
````{dropdown} EoMT DINOv3 Default Train Transform Arguments
```{include} _auto/dinov3eomtsemanticsegmentationtrain_train_transform_args.md
```
````
\# here it would be nice to also have the auto-generated defaults for the val transforms (sorry that I did not see earlier that those are missing)
... and then we can show how to modify ...
```
Wdyt? We can also leave it like this for now and change later. It's good enough for me.

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -216,11 +224,14 @@ def get_transform_args(
     )
     train_transform_args.resolve_auto()
 
+    # Take defaults from train transform.
+    val_args_dict = train_transform_args.model_dump(
+        include={"image_size": True, "normalize": True, "ignore_index": True}
```
liopeer, line 229: I somehow dislike that `"ignore_index"` is in the transform_args. Does not need to be addressed right now, but I should try to refactor this at some point.

liopeer (approved):
Approved. You can decide yourself whether you want to address my nit-picking :)

docs/source/semantic_segmentation.md
guarin: Updated 👍🏼

src/lightly_train/_commands/train_task_helpers.py
guarin: It is required for the random crop transform.

PR #265, opened 2025-08-26
Load images with torch if possible

guarin:
## What has changed and why?

* Fix bug where images were always loaded with PIL
* Remove "L" image mode from `open_image_numpy` as it is not needed anymore

Torch image loading is much faster. The issue was that we listed the supported extensions as `{"jpg", "jpeg", "png"}` instead of `{".jpg", ".jpeg", ".png"}`

## How has it been tested?

* Added unit tests

- [ ] Yes

- [ ] Yes

### DIFF
--- a/tests/_data/test_image_dataset.py
+++ b/tests/_data/test_image_dataset.py
@@ -149,7 +149,9 @@ class TestImageDataset:
     def test___getitem____mode(self, tmp_path: Path, mode: str, extension: str) -> None:
         filenames = [ImageFilename(f"image1.{extension}")]
         image_dir = tmp_path / "images"
-        _create_images(base_path=image_dir, filenames=filenames, mode=mode)
+        _create_images(
+            base_path=image_dir, filenames=filenames, mode=None, convert_mode=mode
+        )
         dataset = ImageDataset(
             image_dir=image_dir,
             image_filenames=filenames,
@@ -186,7 +188,9 @@ class TestImageDataset:
         image_dir = tmp_path / "images"
         mask_dir = tmp_path / "masks"
         _create_images(base_path=image_dir, filenames=img_filenames)
-        _create_images(base_path=mask_dir, filenames=mask_filenames, mode="L")
+        _create_images(
+            base_path=mask_dir, filenames=mask_filenames, mode="L", num_channels=0
+        )
         dataset = ImageDataset(
             image_dir=image_dir,
             image_filenames=img_filenames,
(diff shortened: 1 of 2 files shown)


tests/_data/test_mask_semantic_segmentation_dataset.py
```diff
@@ -71,7 +71,7 @@ def test__getitem__(
         dataset_args = MaskSemanticSegmentationDatasetArgs(
             image_dir=image_dir,
             mask_dir=mask_dir,
-            classes={0: "background", 1: "object"},
+            classes={i: f"class_{i}" for i in range(num_classes)},
```
guarin: Had to fix this because tests kept failing. The issue was that the actual masks contain values in [0, num_classes) but only classes [0, 1] were valid. This resulted in the `is_valid(mask)` check failing because there was a high probability that no [0, 1] values were in the randomly generated mask.

src/lightly_train/_data/image_dataset.py
```diff
@@ -54,7 +54,7 @@ def __getitem__(self, idx: int) -> DatasetItem:
 
         if self.mask_dir:
             maskname = Path(filename).with_suffix(".png")
-            mask = file_helpers.open_image_numpy(self.mask_dir / maskname, mode="L")
+            mask = file_helpers.open_image_numpy(self.mask_dir / maskname, mode="MASK")
```
guarin: Changed to MASK as L is not used anymore

src/lightly_train/_data/file_helpers.py
```diff
@@ -93,9 +93,7 @@ def _get_image_filepaths(image_dir: Path) -> Iterable[Path]:
                 yield fpath
 
 
-def _torchvision_supported_image_extensions() -> set[str]:
-    # See https://pytorch.org/vision/0.18/generated/torchvision.io.read_image.html
-    return {"jpg", "jpeg", "png"}
+_TORCHVISION_SUPPORTED_IMAGE_EXTENSIONS = {".jpg", ".jpeg", ".png"}
```
guarin, line 109: This is the actual fix. Everything else are fixes/refactorings to make the torch loading work correctly.

src/lightly_train/_data/file_helpers.py
```diff
@@ -128,31 +126,18 @@ def open_image_tensor(image_path: Path) -> Tensor:
 
 def open_image_numpy(
     image_path: Path,
-    mode: Literal["RGB", "L", "UNCHANGED", "MASK"] = "RGB",
+    mode: Literal["RGB", "UNCHANGED", "MASK"] = "RGB",
```
simonschoelly: Have we thought about doing these things with an Enum? Something like
```python
class ImageMode(str, Enum):
    RGB = "RGB"
    UNCHANGED = "UNCHANGED"
    MASK = "MASK"
```
From Python 3.11 on there is also a `StrEnum` class.

tests/_data/test_file_helpers.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def test_open_image_numpy__mask(tmp_path: Path) -> None:
+    image_path = tmp_path / "image.png"
+    helpers.create_image(path=image_path, height=32, width=32, mode="L")
+
+    result = file_helpers.open_image_numpy(image_path=image_path, mode="MASK")
+    assert isinstance(result, np.ndarray)
+    assert result.shape == (32, 32)
+    assert result.dtype == np.uint8
```
simonschoelly: Should we perhaps cover the test case where the `dtype` is not `np.unit8`? [io.decode_image](https://docs.pytorch.org/vision/main/generated/torchvision.io.decode_image.html#torchvision.io.decode_image) which is called by `io.read_image` says we could trigger that with a 16-bit png.

guarin: Added, 16-bit png are a bit a pain to work with, not sure how many people actually use them.

simonschoelly: I am also not sure - but could perhaps occur if someone creates a training set by just taking all the image material they have.

simonschoelly (approved):
LGTM

PR #266, opened 2025-08-26
Add semantic segmentation class mapping support

yutong-xiang-97:
## What has changed and why?

Allow mapping multiple classes to a single class for training when using `train_semantic_segmentation` command. More specifically, we now allow the following input format for `classes` in the `data` argument:

```python
classes={
  255: {"name": "background", "values": [0, 1, 2]},
  3: {"name": "class 3", "values": [3]},
}
```

Use the `"name"` to specify the name of the new label, and `"values"` to specify all the class labels to be mapped. 

Note that one can still pass a single string as the label name in the `classes` dict, i.e. the following works:

```python
classes: dict[int, ClassInfo] = {
  255: {"name": "background", "values": [0, 1, 2]},
  3: "class 3", 
}
```

Internally, all the inputs will be mapped as such:

```python
classes: dict[int, ClassInfo] = {
  255: {"name": "background", "values": [0, 1, 2]},
  3: {"name": "class 3", "values": [3]},
}
```

and then converted to a `class_mapping` with continuous natural numbers as the new labels, without the ignored labels.

Final note for ignore behavior:

- if you ignore a label that appear in keys, then all labels mapped to this label are ignored; 
- if you ignore a label that is to be mapped (which shouldn't appear in the keys, see case 2 below), only the label itself will be ignored.

Several cases:

1. Self-mapping is allowed. Note that we see the label `255` as a new label anyway.

```python
classes: dict[int, ClassInfo] = {
  255: {"name": "background", "values": [0, 1, 255]}, # this is allowed
  3: "class 3", 
}
```

2. The classes to be mapped are not allowed to appear in the class keys:

```python
classes: dict[int, ClassInfo] = {
  255: {"name": "background", "values": [0, 1, 2]},
  0: "class 0", # this is not allowed
}

classes: dict[int, ClassInfo] = {
  255: {"name": "background-255", "values": [0, 1, 2]},
  0: {"name": "background-0", "values": [3, 4 ,5]}, # this is not allowed
}
```

3. Multi-mapping is not allowed

```python
classes: dict[int, ClassInfo] = {
  255: {"name": "background-255", "values": [0, 1, 2]},
  254: {"name": "background-254", "values": [0, 1, 2]}, #this is not allowed
}
```

## How has it been tested?

Unit tests & Actual task runs on the compute machine.

- [x] Yes

- [ ] Yes
- [x] Not needed (will be updated once everything for segmentation finishes)

### DIFF
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -177,7 +177,7 @@ class PrettyFormatArgsJSONEncoder(JSONEncoder):
         if isinstance(obj, set):
             return sorted(obj)
         try:
-            return self.default(obj)
+            return super().default(obj)
         except TypeError:
             # Return class name for objects that cannot be serialized
             return obj.__class__.__name__
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -175,7 +175,7 @@ class PrettyFormatArgsJSONEncoder(JSONEncoder):
         if isinstance(obj, Path):
             return str(obj)
         if isinstance(obj, set):
-            return sorted([super().default(o) for o in obj])
+            return sorted([self.default(o) for o in obj])
         try:
             return super().default(obj)
         except TypeError:
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -175,7 +175,15 @@ class PrettyFormatArgsJSONEncoder(JSONEncoder):
         if isinstance(obj, Path):
             return str(obj)
         if isinstance(obj, set):
-            return sorted([self.default(o) for o in obj])
+            result = []
+            for o in obj:
+                # Check if the object is already JSON-serializable
+                try:
+                    json.dumps(o)  # Test if it's directly serializable
+                    result.append(o)  # Use the object directly
+                except (TypeError, ValueError):
+                    result.append(self.default(o))  # Apply custom formatting
+            return sorted(result)
         try:
             return super().default(obj)
         except TypeError:
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -179,7 +179,9 @@ class PrettyFormatArgsJSONEncoder(JSONEncoder):
             for o in obj:
                 # Check if the object is already JSON-serializable
                 try:
-                    json.dumps(o, cls=self.__class__)  # Test with the same encoder class
+                    json.dumps(
+                        o, cls=self.__class__
+                    )  # Test with the same encoder class
                     result.append(o)  # Use the object directly
                 except (TypeError, ValueError):
                     result.append(self.default(o))  # Apply custom formatting
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -179,7 +179,7 @@ class PrettyFormatArgsJSONEncoder(JSONEncoder):
             for o in obj:
                 # Check if the object is already JSON-serializable
                 try:
-                    json.dumps(o)  # Test if it's directly serializable
+                    json.dumps(o, cls=self.__class__)  # Test with the same encoder class
                     result.append(o)  # Use the object directly
                 except (TypeError, ValueError):
                     result.append(self.default(o))  # Apply custom formatting
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -175,17 +175,7 @@ class PrettyFormatArgsJSONEncoder(JSONEncoder):
         if isinstance(obj, Path):
             return str(obj)
         if isinstance(obj, set):
-            result = []
-            for o in obj:
-                # Check if the object is already JSON-serializable
-                try:
-                    json.dumps(
-                        o, cls=self.__class__
-                    )  # Test with the same encoder class
-                    result.append(o)  # Use the object directly
-                except (TypeError, ValueError):
-                    result.append(self.default(o))  # Apply custom formatting
-            return sorted(result)
+            return sorted(list(obj))
         try:
             return super().default(obj)
         except TypeError:
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -175,7 +175,7 @@ class PrettyFormatArgsJSONEncoder(JSONEncoder):
         if isinstance(obj, Path):
             return str(obj)
         if isinstance(obj, set):
-            return sorted(obj)
+            return sorted(list(obj))
         try:
             return super().default(obj)
         except TypeError:
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -174,6 +174,8 @@ class PrettyFormatArgsJSONEncoder(JSONEncoder):
     def default(self, obj: Any) -> Any:
         if isinstance(obj, Path):
             return str(obj)
+        if isinstance(obj, set):
+            return sorted(list(obj))
         try:
             return super().default(obj)
         except TypeError:
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -175,9 +175,9 @@ class PrettyFormatArgsJSONEncoder(JSONEncoder):
         if isinstance(obj, Path):
             return str(obj)
         if isinstance(obj, set):
-            return sorted(list(obj))
+            return sorted(obj)
         try:
-            return super().default(obj)
+            return self.default(obj)
         except TypeError:
             # Return class name for objects that cannot be serialized
             return obj.__class__.__name__
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -175,7 +175,7 @@ class PrettyFormatArgsJSONEncoder(JSONEncoder):
         if isinstance(obj, Path):
             return str(obj)
         if isinstance(obj, set):
-            return sorted(list(obj))
+            return sorted([super().default(o) for o in obj])
         try:
             return super().default(obj)
         except TypeError:
--- a/src/lightly_train/_data/mask_semantic_segmentation_dataset.py
+++ b/src/lightly_train/_data/mask_semantic_segmentation_dataset.py
@@ -14,7 +14,7 @@ from typing import ClassVar, Dict, Literal, Union
 import numpy as np
 import torch
 from numpy.typing import NDArray
-from pydantic import Field, TypeAdapter, field_serializer, field_validator
+from pydantic import Field, TypeAdapter, field_validator
 from torch import Tensor
 from torch.utils.data import Dataset
 
@@ -34,10 +34,6 @@ from lightly_train.types import (
 class ClassInfo(PydanticConfig):
     name: str
     values: set[int] = Field(strict=False)
-
-    @field_serializer("values")
-    def serialize_values(self, values: set[int]) -> list[int]:
-        return sorted(values)
 
 
 class MaskSemanticSegmentationDataset(Dataset[MaskSemanticSegmentationDatasetItem]):
@@ -210,10 +206,6 @@ class MaskSemanticSegmentationDatasetArgs(PydanticConfig):
     check_empty_targets: bool = True
     ignore_index: int
 
-    @field_serializer("ignore_classes")
-    def serialize_ignore_classes(self, values: set[int] | None) -> list[int] | None:
-        return sorted(values) if values is not None else None
-
     # NOTE(Guarin, 07/25): The interface with below methods is experimental. Not yet
     # sure if it makes sense to have this in dataset args.
     def list_image_filenames(self) -> Iterable[ImageFilename]:
@@ -240,10 +232,6 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
     classes: dict[int, ClassInfo]
     ignore_classes: set[int] | None = Field(default=None, strict=False)
     check_empty_targets: bool = True
-
-    @field_serializer("ignore_classes")
-    def serialize_ignore_classes(self, values: set[int] | None) -> list[int] | None:
-        return sorted(values) if values is not None else None
 
     @field_validator("classes", mode="before")
     @classmethod
--- a/src/lightly_train/_data/mask_semantic_segmentation_dataset.py
+++ b/src/lightly_train/_data/mask_semantic_segmentation_dataset.py
@@ -14,7 +14,7 @@ from typing import ClassVar, Literal
 import numpy as np
 import torch
 from numpy.typing import NDArray
-from pydantic import Field
+from pydantic import BaseModel, Field, field_validator
 from torch import Tensor
 from torch.utils.data import Dataset
 
@@ -29,6 +29,11 @@ from lightly_train.types import (
     MaskSemanticSegmentationDatasetItem,
     PathLike,
 )
+
+
+class ClassInfo(BaseModel):
+    name: str
+    values: list[int]
 
 
 class MaskSemanticSegmentationDataset(Dataset[MaskSemanticSegmentationDatasetItem]):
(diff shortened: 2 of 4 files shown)


src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -174,6 +174,8 @@ class PrettyFormatArgsJSONEncoder(JSONEncoder):
     def default(self, obj: Any) -> Any:
         if isinstance(obj, Path):
             return str(obj)
+        if isinstance(obj, set):
+            return sorted(list(obj))
```
guarin, line 178:
```suggestion
        if isinstance(obj, set):
            return sorted([super().default(o) for o in obj])
```
Objects in the set must also be pretty formatted

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -91,12 +96,8 @@ def filter_empty_targets(self) -> None:
...
     def get_class_mapping(self) -> dict[int, int]:
         # Verify the classes are set (for mypy).
-        assert self.args.classes is not None, (
-            "Segmentation dataset classes must be set."
-        )
-
-        # Set original classes.
-        original_classes = self.args.classes.keys()
+        input_classes = self.args.classes
+        assert input_classes is not None, "Segmentation dataset classes must be set."
```
guarin: This looks like an error in `MaskSegmentationDatasetArgs`. I don't think the `classes` attribute can ever be `None`.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -241,15 +255,56 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
...
+                # Let Pydantic validate the structure and types
+                class_info_obj = ClassInfo.model_validate(class_info)
+
+                # Check for overlapping values across different class mappings
+                for value in class_info_obj.values:
+                    if value in all_mapped_values:
+                        raise ValueError(
+                            f"Class value {value} appears in multiple class mappings. "
+                            f"Each class value can only be mapped to one target class."
+                        )
```
guarin: Could you add an example in the error message? Otherwise it might be really difficult for users to understand what they did wrong. For example, they don't know what "target class" means. They might not even know that the error is due to a problem in the `data={"classes": {..}}` dict. I would point that out explicitly as well.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -241,15 +255,56 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
...
+                    # Check if the value conflicts with any class key
+                    if (value in class_keys) and (value != class_id):
+                        raise ValueError(
+                            f"Class value {value} in ClassInfo for class {class_id} conflicts "
+                            f"with class key {value}. Class keys cannot appear as values in ClassInfo instances."
+                        )
```
guarin: Why do we have to check for this? We values and the class ids should be unrelated. You might be able to remove this check if you change the code above to:
```
if isinstance(class_info, str):
                result[class_id] = ClassInfo(name=class_info, values={class_id})
                all_mapped_values.add(class_id)
```


From the PR description:
```
classes: dict[int, ClassInfo] = {
  255: {"name": "background", "values": [0, 1, 2]},
  0: {"name": "background", "values": [3, 4 ,5]}, # this is not allowed
}
```
IMO this should be allowed. Every value should appear at most once in values and every class id should appear at most once in the class ids. But values and class ids remain independent. Whether two classes can be called "background" is another question. I think it might result in errors because we use the class names in the classwise metrics and those might fail if two metrics have the same name. Not 100% sure though.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -105,16 +106,29 @@ def get_class_mapping(self) -> dict[int, int]:
...
+
+        for class_key, class_info in input_classes.items():
+            # Skip if the class key itself is in ignore_classes
+            if class_key in ignore_classes:
+                continue
+
+            # Check if this ClassInfo has any non-ignored values
+            has_included_classes = any(
+                val not in ignore_classes for val in class_info.values
+            )
```
guarin: I don't think this check is needed. `ignore_classes` only includes class keys but not the values. We shouldn't mix the two.

Maybe the mapping can even be written as:
```
class_mapping = {class_id: i for i, class_id in enumerate(self.args.included_classes())}
```

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
yutong-xiang-97: The problem arises when class `0` is not a class to be mapped, i.e., the original class to use:

```python
classes: dict[int, ClassInfo] = {
  255: {"name": "background", "values": [0, 1, 2]},
  0: "class 0", # this is not allowed
}
```

Also, mapping the classes to a new class by re-using an old label does not sound like a good practice to me. It creates potential confusion what `0` actually refers to. In this case, the users can choose to map to a non-existant label in the original dataset and it is clear to everyone what is done here:

```python
classes: dict[int, ClassInfo] = {
  255: {"name": "class-255", "values": [0, 1, 2]},
  254: {"name": "class-254", "values": [3, 4 ,5]},
}
```

Also, for the same class name issue, it was a typo on my side in the PR message, but maybe it is good that we have a check to enforce uniqueness, cuz `get_class_mapping` does not check the names at all. In this case, when you do training there will be two "background" classes, which is not desirable. I will fix it in a new commit.

```python
classes: dict[int, ClassInfo] = {
  255: {"name": "background", "values": [0, 1, 2]},
  254: {"name": "background", "values": [3, 4 ,5]}, # this is NOT allowed
}
```

```python
classes: dict[int, ClassInfo] = {
  255: {"name": "background-255", "values": [0, 1, 2]},
  254: {"name": "background-254", "values": [3, 4 ,5]}, # this is allowed
}
```

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
yutong-xiang-97: The users may do something like this

```python
classes={
  255: {"name": "background", "values": [0, 1, 2]},
  3: "class 3", 
}
ignore_classes={2}
```

If `ignore_classes` is supposed to only include class keys but not the values, then I can do a check, and raise a warning to the user that they should ignore only the keys, ignoring anything not existing in the keys will not be considered. I will make this change into a new commit.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -90,31 +96,14 @@ def filter_empty_targets(self) -> None:
...
 
     def get_class_mapping(self) -> dict[int, int]:
-        # Verify the classes are set (for mypy).
-        assert self.args.classes is not None, (
-            "Segmentation dataset classes must be set."
+        all_new_classes = set(self.args.classes.keys())
+        ignore_classes = self.args.ignore_classes
+        included_classes = (
+            all_new_classes - ignore_classes if ignore_classes else all_new_classes
         )
```
guarin: I am a bit afraid that this might result in different iteration order for `included_classes` as it is a set. This is a problem because we want the mapping to be deterministic across runs.

Alternative proposal:
```
ignore_classes = set() if self.args.ignore_classes is None else self.args.ignore_classes
class_mapping = {class_id: i for i, class_id in enumerate(self.args.classes) if class_id not in ignore_classes}
```

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -241,15 +230,139 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
...
+    @field_validator("classes", mode="before")
+    @classmethod
+    def validate_classes(
+        cls, classes: dict[int, str | dict[str, str | Sequence[int]]]
+    ) -> dict[int, ClassInfo]:
+        new_classes = set(classes.keys())
+        all_class_names: set[str] = set()
+        all_class_values: set[int] = set()
+        result: dict[int, ClassInfo] = {}
+
```
guarin: I looked a bit into this and I think the proper way to do this is to 1) first validate that the input is correctly formatted (it is a dict[int, str | ClassInfo]), then 2) convert `dict[int, str | ClassInfo]` to `dict[int, ClassInfo]`, then 3) run all the extra checks whether classes/values interfere with each other.

For 1) you can use a pydantic `TypeAdapter`:
```
_CLASSES_INPUT_ADAPTER = TypeAdapter(dict[int, str | ClassInfo])

def validate_classes(cls, classes: dict[int, str | ClassInfo]):
    classes = _CLASSES_INPUT_ADAPTER.validate_python(classes)
```

Doing this first has the benefit that the user gets the correct error message from pydantic. If we only validate after converting to `ClassInfo` then the user will get confused if they set `classes={0: "background"}` and they get an error related to `ClassInfo`.

The second benefit is that we know that the classes dict has the correct format.


After this the conversion to ClassInfo is simple:
```
class_infos: dict[int, ClassInfo] = {key: val if isinstance(val, ClassInfo) else ClassInfo(name=val, values={key})}
```

And then you can run the class/values checks. Note that you now don't have to handle the `dict[int, str]` case anymore.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
guarin: IMO you don't have to do anything extra, just remove the check and treat `values` and `ignore_classes` completely independently.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
yutong-xiang-97: If you have to keep the order, then in your implementation we should not use `enumerate()`, cuz the mapped labels may not be continuous if there are indices to ignore.

We can add a `sort()` to the original impl to fix this.

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -174,6 +174,16 @@ class PrettyFormatArgsJSONEncoder(JSONEncoder):
...
             return str(obj)
+        if isinstance(obj, set):
+            result = []
+            for o in obj:
+                # Check if the object is already JSON-serializable
+                try:
+                    json.dumps(o)  # Test if it's directly serializable
+                    result.append(o)  # Use the object directly
+                except (TypeError, ValueError):
+                    result.append(self.default(o))  # Apply custom formatting
```
guarin: Why does `return sorted([self.default(o) for o in obj])` not work? The issue with calling `json.dumps(o)` is that this will not apply the formatting with the custom json encoder class and the settings that we pass in `pretty_format_args`

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -31,6 +31,11 @@
 )
 
 
+class ClassInfo(BaseModel):
```
guarin: We always use `PydanticConfig` defined in `lightly_train/_configs/config.py` to have the same base class for all our pydantic models:
```suggestion
class ClassInfo(PydanticConfig):
```

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
guarin: I would prefer to keep the original class order. We also depend on it elsewhere in the code. For example here: https://github.com/lightly-ai/lightly-train/blob/0f3e02d22f144b13c504a17f979103d12acfa6b3/src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/task_model.py#L97

I know it is not ideal, we have to refactor this eventually.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -241,15 +229,85 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
...
+            Dict[int, Union[str, ClassInfo]]
+        ).validate_python(classes)
+
+        # Convert to ClassInfo objects
+        class_infos: dict[int, ClassInfo] = {}
+        for class_id, class_info in classes_validated.items():
+            if isinstance(class_info, str):
+                class_infos[class_id] = ClassInfo(name=class_info, values={class_id})
+            else:
+                class_infos[class_id] = ClassInfo.model_validate(class_info)
```
guarin: No need for validation anymore as it was already validated by the type adapter.
```suggestion
                class_infos[class_id] = class_info
```

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -241,15 +229,85 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
...
+                        f"classes = {{\n"
+                        f"  255: {{'name': 'background-255', 'values': [0, 1, 2]}},\n"
+                        f"  0: {{'name': 'background-0', 'values': [3, 4, 5]}}  # <- class key 0 conflicts with class 0 above\n"
+                        f"}}"
+                        f"CORRECT (class keys don't appear as values):\n"
+                        f"classes = {{\n"
+                        f"  255: {{'name': 'background', 'values': [0, 1, 2]}},\n"
+                        f"  10: 'class 10'  # <- class key 10 does not conflict with any value\n"
+                        f"}}"
+                    )
```
guarin: This will already be caught by the above check as everything is a ClassInfo now.
```suggestion
```

And `value != class_id` should not be checked as the following is valid:
```
        classes_with_existing_keys = {
            0: {"name": "background", "values": [1, 2, 3]},
            10: {"name": "person", "values": [7, 8, 0]},  # Value 0 conflicts with key 0 <--- This is not a problem
        }
```

src/lightly_train/_commands/train_task_helpers.py
yutong-xiang-97: It will print "int" but not the actual number like shown the following in the training log:

```
            "148": {
                "name": "glass",
                "values": [
                    "int"
                ]
            },
            "149": {
                "name": "clock",
                "values": [
                    "int"
                ]
            },
```

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
yutong-xiang-97: Yes, I've added a sort function already in the updates.

src/lightly_train/_commands/train_task_helpers.py
yutong-xiang-97: If you want to use the custom JSON encoder class then this may be the way to go:

`json.dumps(o, cls=self.__class__)`

Added to a new commit. Also I realized we don't have tests for `PrettyFormatArgsJSONEncoder`. Will create an issue for this.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
guarin: This is still open, here is the problem:

In `MaskSemanticSegmentationDataset` we create the mapping with this code:
```
        all_new_classes = set(self.args.classes.keys())
        ignore_classes = self.args.ignore_classes
        included_classes = sorted(
            all_new_classes - ignore_classes if ignore_classes else all_new_classes
        )
        class_mapping = {class_id: i for i, class_id in enumerate(included_classes)}
```

Assuming we have `classes={2: "cls 2", 4: "cls 4", 3: "cls 3"}` this will return `{2: 0, 3: 1, 4: 2}`

In `DINOv2EoMTSemanticSegmentationModel` (and all the other task models) we have to map the internal classes back to the external ones ( revert the mapping). But that code uses:

```
 internal_class_to_class = list(self.classes.keys()) 
```

With the above class definition this will result in:
```
internal_class_to_class = [2, 4, 3]
```
which is equivalent to:
```
{0: 2, 1: 4, 2: 3}
```

This is clearly not the reverse mapping of the `class_mapping` in `MaskSemanticSegmentationDataset` because class 4 is first mapped to 2 in with `class_mapping` and then the 2 is mapped back to 3 instead of 4 with `internal_class_to_class`.

Either we have to use `classes` everywhere without changing the order or we sort it everywhere. But we cannot have a mix of the two.

src/lightly_train/_commands/train_task_helpers.py
yutong-xiang-97: I realized that the problem is not the JSONEncoder but the serialization of the Pydantic model config. In `.model_dump()` the `set` type is not properly serialized. You can check the `ignore_classes` attribute in the training log before this PR ("set" instead of the class labels). So, whenever we use `set` we need to add a custom field serializer. I've made the change in the new commit.

guarin (approved):

PR #270, opened 2025-08-28
Improve semantic segmentation empty target handling

guarin:
## What has changed and why?

* Remove empty target check
* Remove error if no valid mask is found
* Handle masks with unknown classes

This solves two issues:
1. We don't have to check empty targets at the beginning of the run anymore. This makes dataset loading faster and avoids the mmap to list conversion.
2. No more errors can happen because of invalid masks during training. This avoids crashes late in the training run.

## How has it been tested?

Ran 3 runs with and without the changes and there are no noticeable differences. With DINOv2 EoMT ViT-S I got around 49.8% mIoU on ADE20K on average. 

I also checked whether we can remove the `is_valid_mask` test completely but this resulted in lower mIoU. 

- [x] Yes

- [ ] Yes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,7 @@ requires-python = ">=3.8, <3.13"    # 3.13 is not yet supported by PyTorch, Torc
 dependencies = [
     "albumentations>=1.3.1,!=1.4.18,!=1.4.19,!=1.4.20,!=1.4.21,!=1.4.22", # 1.4.18-1.4.22 will cause mypy issues
     "eval-type-backport>=0.2",  # Required for Pydantic. Allows 3.10 types in old Python versions.
+    "filelock>=3.0.0",     # Cross-platform file locking
     "lightly>=1.5.20",
     "omegaconf>=2.3",
     "psutil>=5.0",
--- a/src/lightly_train/_commands/common_helpers.py
+++ b/src/lightly_train/_commands/common_helpers.py
@@ -19,6 +19,7 @@ from pathlib import Path
 from typing import Any, Generator, Iterable, Literal, Sequence, Sized, TypeVar
 
 import torch
+from filelock import FileLock
 from pytorch_lightning.accelerators.accelerator import Accelerator
 from pytorch_lightning.accelerators.cpu import CPUAccelerator
 from pytorch_lightning.accelerators.cuda import CUDAAccelerator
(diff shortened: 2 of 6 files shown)


src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -119,7 +88,7 @@ def get_binary_masks(self, mask: Tensor) -> BinaryMasksDict:
         # Iterate over the labels present in the mask.
         for class_id in class_ids:
             # Check if the class id is the valid classes.
-            if class_id not in self.valid_classes:
+            if class_id not in self.class_mapping:
```
guarin, line 91: Changed this from `valid_classes` to `class_mapping` because `class_mapping` is a dict and `valid_classes` a list/tensor.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -60,39 +58,10 @@ def __init__(
...
-        # Display the number of filtered files.
-        # TODO(Thomas, 07/25): Change the print to logging once the function is moved
-        # outside of the dataset.
-        n_filtered_files = len(self) - len(new_image_filenames)
-        print(f"Filtered {n_filtered_files} invalid masks out of {len(self)}.")
-
-        # Update the list of valid files.
-        self.image_filenames = new_image_filenames
+        unique_classes: Tensor = mask.unique()
+        return bool(torch.isin(unique_classes, self.valid_classes).any())
```
guarin: Changed from numpy to torch to avoid torch -> numpy conversion

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -129,7 +98,11 @@ def get_binary_masks(self, mask: Tensor) -> BinaryMasksDict:
             img_labels.append(self.class_mapping[class_id])
 
         binary_masks: BinaryMasksDict = {
-            "masks": torch.stack(img_masks),
+            "masks": (
+                torch.stack(img_masks)
+                if img_masks
```
yutong-xiang-97, line 103: `is_mask_valid` has the identical checking conditions with the for loop here. So whenever you get invalid masks, in `get_binary_masks` you get empty `img_masks`, which makes binary masks all zeros. On the other hand, `remap_mask` in this case will return a mask filled with `ignore_index`, so the behavior is inconsistent.

I would propose we use `ignore_index` for all cases, which basically ignores all invalid masks. WDYT?

guarin: The code here doesn't return masks filled with zeros. It returns an empty tensor because the first dimension has size 0. So basically you have:
* if `img_masks` list is empty: empty tensor with shape `(0, H, W)`
* if `img_masks` is not empty: tensor with shape `(num_binary_masks, H, W)`

The special handling is needed because `torch.stack` raises an error if you pass an empty list.

yutong-xiang-97 (approved):
Thanks for pointing that out! Looks good to me.

PR #272, opened 2025-09-01
Add option to export ONNX with float16 precision

simonschoelly:
## What has changed and why?

This PR adds a flag `half` to the ONNX export task. When this flag is set, the model weights and the internal calculations are done with `float16` precision instead of `float32`.

I also had to adjust the tolerance in our tests a bit as we got higher deviations now with `float16` precision.

**To discuss**:

Instead of having a flag `half` should we instead have a field `precision` that allows values such as `f16` and `f32`? And perhaps later allows one to specify mixed precision.

## How has it been tested?

- Changes are covered by tests
- Manually exported ONNX and run them on a Jetson Orin

- [x] Yes

- [ ] Yes

We do not document ONNX export yet.


src/lightly_train/_commands/export_task.py
```diff
@@ -165,13 +174,19 @@ def _export_task_from_config(config: ExportTaskConfig) -> None:
...
-            input_feed = {input_name: x.cpu().numpy()}
+            input_feed = {input_name: x_onnx.numpy()}
             outputs_onnx = session.run(output_names=output_names, input_feed=input_feed)
             outputs_onnx = tuple(torch.from_numpy(y) for y in outputs_onnx)
 
-            outputs_model = task_model(x)
+            task_model = task_model_helpers.load_model_from_checkpoint(
+                checkpoint=checkpoint_path, device="cpu"
+            )
+            outputs_model = task_model(x_model)
```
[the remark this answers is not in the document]
simonschoelly: I am not sure about that. In such a case we would need to make absolutely sure that the copy is a deep copy and there is no aliasing.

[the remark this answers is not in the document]
liopeer: It's fine like it is, we don't export e.g. every epoch, so a bit of overhead doesn't matter.

src/lightly_train/_commands/export_task.py
```diff
@@ -165,13 +174,19 @@ def _export_task_from_config(config: ExportTaskConfig) -> None:
 
             onnx.checker.check_model(out_path, full_check=True)
 
-            x = torch.rand_like(dummy_input)
+            # Always run the reference input in float32 and on cpu for consistency
+            x_model = torch.rand_like(dummy_input, dtype=torch.float32, device="cpu")
```
liopeer, line 200: Does this not raise an error when you input a float32 array into an fp16 model?

tests/_commands/test_export_task.py
```diff
@@ -64,15 +64,15 @@ def dinov2_vits14_eomt_checkpoint(tmp_path_factory: pytest.TempPathFactory) -> P
...
+    (1, 42, 154, False),
+    (1, 42, 154, False),
+    (2, 14, 14, False),
+    (3, 140, 280, True),
+    (4, 266, 28, True),
 ]
 
 
-@pytest.mark.parametrize("batch_size,height,width", onnx_export_testset)
+@pytest.mark.parametrize("batch_size,height,width, half", onnx_export_testset)
```
liopeer:
```suggestion
@pytest.mark.parametrize("batch_size,height,width,half", onnx_export_testset)
```
Nit: for consistency

liopeer (dismissed):
Looks really good overall! About the `half` parameter. I would also prefer `precision` that would be more flexible and allow to extend to other precisions as well. For the naming conventions I would follow the naming already use in the `.train` commands, i.e. the convention that PyTorch Lightning also uses: https://lightning.ai/docs/pytorch/stable/common/trainer.html#trainer-class-api

src/lightly_train/_commands/export_task.py
simonschoelly: Ah, you mean if the model was trained on f16 inputs previously? That is perhaps a good point. But how do we figure out the what the exact input type of the model is? We probably cannot just take the first buffer.

Or perhaps you are referring to the fact that I use `float32` here even if the exported model is `float16`? My idea was that perhaps there could be errors when converting from `float32` to `float16`. So I run the reference inputs always with `float32`. The inputs to the ONNX model are still `float16` as specified in the line below.

src/lightly_train/_commands/export_task.py
```diff
@@ -63,6 +75,7 @@ def export_onnx(
     num_channels: int = 3,
     height: int = 224,
     width: int = 224,
+    precision: OnnxPrecision = OnnxPrecision.F32_TRUE,
```
guarin: We should not force the user to pass enums in public functions as that makes using the function more complicated and users have to learn about the enum class and where to import it from. Instead we use literals. See for example `train.py` or `train_task.py`.

Without enums:
```
export_onnx(..., precision="16-true")
```

With enums:
```
from lightly_train._commands.export_task import OnnxPrecision

export_onnx(..., precision=OnnxPrecision.F32_TRUE)
```

Internally we can use the enum. Pydantic should be able to convert the string value automatically to it.

src/lightly_train/_commands/export_task.py
```diff
@@ -165,13 +185,19 @@ def _export_task_from_config(config: ExportTaskConfig) -> None:
 
             onnx.checker.check_model(out_path, full_check=True)
 
-            x = torch.rand_like(dummy_input)
+            # Always run the reference input in float32 and on cpu for consistency
+            x_model = torch.rand_like(dummy_input, dtype=torch.float32, device="cpu")
+            x_onnx = x_model.to(onnx_dtype)
```
guarin, line 201: Nit:
```suggestion
            # Always run the reference input on cpu for consistency
            x_model = torch.rand_like(dummy_input, dtype=onnx_dtype, device="cpu")
```

src/lightly_train/_commands/export_task.py
```diff
@@ -165,13 +185,19 @@ def _export_task_from_config(config: ExportTaskConfig) -> None:
...
-            input_feed = {input_name: x.cpu().numpy()}
+            input_feed = {input_name: x_onnx.numpy()}
             outputs_onnx = session.run(output_names=output_names, input_feed=input_feed)
             outputs_onnx = tuple(torch.from_numpy(y) for y in outputs_onnx)
 
-            outputs_model = task_model(x)
+            task_model = task_model_helpers.load_model_from_checkpoint(
+                checkpoint=checkpoint_path, device="cpu"
+            )
+            outputs_model = task_model(x_model)
```
guarin, line 211: Why not `task_model.to("cpu")` instead of loading again from the checkpoint?

src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/task_model.py
```diff
@@ -415,13 +415,23 @@ def untile(
...
         # Initialize the tensors containing the final predictions.
         for size in image_sizes:
             logit_sums.append(
-                torch.zeros((crop_logits.shape[1], *size), device=crop_logits.device)
+                torch.zeros(
+                    (crop_logits.shape[1], *size),
+                    device=crop_logits.device,
+                    dtype=dtype,
+                )
             )
```
guarin: Nit (doesn't need to be addressed): The "modern" torch way to do this is:
```
crop_logits.new_zeros((crop_logits.shape[1], *size))
```
Then the new tensor is automatically on the correct device with the correct dtype.

src/lightly_train/_commands/export_task.py
simonschoelly: My idea is here that there might be some degradation in quality when we call ` task_model.to(onnx_dtype)` above. Therefore I want to rather compare the output of the model that has not been converted to another dtype the onnx model that has been converted. That is also why call `x_model = torch.rand_like(dummy_input, dtype=torch.float32, device="cpu")` above.
`

src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/task_model.py
simonschoelly: No, that's a good point, I did not know `new_zeros`. Changed it - and also I changed the dtype for counts to `int32` as we only use this for counting and dividing and do not return it.

src/lightly_train/_commands/export_task.py
simonschoelly: Done

src/lightly_train/_commands/export_task.py
simonschoelly: See bellow

src/lightly_train/_commands/export_task.py
guarin: Makes sense 👍🏼

PR #273, opened 2025-09-01
Raise ValueError during ONNX export if image size not divisible by patch size

simonschoelly:
## What has changed and why?

This PR makes it so that when we export an ONNX model and the iamge size is not divisible by the patch size, we raise a ValueError that explains what is going on. Previously the user would get a confusing RuntimeError.

## How has it been tested?

Case is covered by two new tests.

- [ ] Yes

We get an exception anyways if the image size is not correct, this just gives a gentler error message.

- [ ] Yes

No documentation yet

### DIFF
--- a/tests/_commands/test_export_task.py
+++ b/tests/_commands/test_export_task.py
@@ -127,7 +127,6 @@ def test_onnx_export(
         torch.testing.assert_close(ort_y, expected_y, rtol=rtol, atol=atol)
 
 
-@pytest.mark.parametrize("batch_size,height,width", onnx_export_testset)
 @pytest.mark.skipif(
     sys.version_info < (3, 9),
     reason="Requires Python 3.9 or higher for image preprocessing.",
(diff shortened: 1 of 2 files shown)


tests/_commands/test_export_task.py
```diff
@@ -125,3 +125,77 @@ def test_onnx_export(
...
+
+@pytest.mark.skipif(
+    sys.version_info < (3, 9),
+    reason="Requires Python 3.9 or higher for image preprocessing.",
+)
+@pytest.mark.skipif(not RequirementCache("onnx"), reason="onnx not installed")
+@pytest.mark.skipif(
+    not RequirementCache("onnxruntime"), reason="onnxruntime not installed"
+)
+def test_onnx_export__height_not_patch_size_multiple_fails(
```
yutong-xiang-97, line 138: Instead of using two tests, you could parameterize the height and width and construct different types of failure cases there.

simonschoelly: I quickly looked into this, but it feels like the code is getting messier if I do that. So if its ok with you then we leave it as it is.

yutong-xiang-97 (approved):
Approved

PR #275, opened 2025-09-02
Support custom mask names

yutong-xiang-97:
## Summary
• Add memory-mapped sequence task class for efficient large dataset handling
• Improve mask semantic segmentation dataset to support custom mask naming patterns
• Move helper functions to common_helpers module for better code organization

## Test plan
- [x] Add comprehensive test coverage for memory-mapped sequence task class
- [x] Add tests for custom mask name pattern support in semantic segmentation dataset
- [x] Add tests for file helper functions
- [x] Update existing tests to cover new functionality
- [x] All tests pass with `make test-ci`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Supported using custom mask names for the inputs in semantic segmentation.
+
 ### Deprecated
 
 ### Removed
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -25,7 +25,7 @@ from torch import Tensor
 from torch.utils.data import DataLoader, Dataset
 
 from lightly_train._configs import validate
-from lightly_train._data import cache
+from lightly_train._data import cache, file_helpers
 from lightly_train._data._serialize import memory_mapped_sequence_task
 from lightly_train._data._serialize.memory_mapped_sequence_task import (
     MemoryMappedSequenceTask,
@@ -360,6 +360,24 @@ def _decrement_and_cleanup_if_zero(mmap_file: Path, ref_file: Path) -> None:
         pass  # Another process already cleaned up
 
 
+def list_image_and_mask_filepaths(
+    image_dir: Path, mask_file: str
+) -> Iterable[tuple[str, str]]:
+    for image_filepath in file_helpers.list_image_files(imgs_and_dirs=[image_dir]):
+        mask_filepath = Path(
+            mask_file.format(
+                image_path=image_filepath,
+            )
+        )
+
+        if mask_filepath.exists():
+            yield image_filepath.as_posix(), mask_filepath.as_posix()
+        else:
+            logger.warning(
+                f"Mask file '{mask_filepath}' for image '{image_filepath}' does not exist."
+            )
+
+
 def get_dataset_mmap_filenames(
     fabric: Fabric,
     filepaths: Iterable[tuple[str, ...]],
@@ -412,7 +430,11 @@ def get_dataset(
     transform: TaskTransform,
     mmap_filepath: Path,
 ) -> MaskSemanticSegmentationDataset:
-    image_and_mask_filepaths = list(dataset_args.list_image_and_mask_filepaths())
+    image_and_mask_filepaths = list(
+        list_image_and_mask_filepaths(
+            image_dir=dataset_args.image_dir, mask_file=dataset_args.mask_dir_or_file
+        )
+    )
     dataset_cls = dataset_args.get_dataset_cls()
     return dataset_cls(
         dataset_args=dataset_args,
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -360,7 +360,7 @@ def _decrement_and_cleanup_if_zero(mmap_file: Path, ref_file: Path) -> None:
 
 def get_dataset_mmap_filenames(
     fabric: Fabric,
-    filenames: Iterable[str],
+    filenames: Iterable[tuple[Path, Path]],
     mmap_filepath: Path,
 ) -> MemoryMappedSequence[str]:
     """Returns memory-mapped filenames shared across all ranks.
@@ -391,7 +391,7 @@ def get_dataset_mmap_filenames(
             not is_shared_filesystem and fabric.local_rank == 0
         ):
             memory_mapped_sequence.write_filenames_to_file(
-                filenames=filenames,
+                filenames=filenames,  # type: ignore[arg-type]
                 mmap_filepath=mmap_filepath,
             )
 
@@ -407,13 +407,13 @@ def get_dataset(
     transform: TaskTransform,
     mmap_filepath: Path,
 ) -> MaskSemanticSegmentationDataset:
-    filenames = list(dataset_args.list_image_filenames())
+    image_and_mask_filepaths = list(dataset_args.list_image_and_mask_filepaths())
     dataset_cls = dataset_args.get_dataset_cls()
     return dataset_cls(
         dataset_args=dataset_args,
-        image_filenames=get_dataset_mmap_filenames(
+        image_and_mask_filepaths=get_dataset_mmap_filenames(  # type: ignore[arg-type]
             fabric=fabric,
-            filenames=filenames,
+            filenames=image_and_mask_filepaths,
             mmap_filepath=mmap_filepath,
         ),
         transform=transform,
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -26,8 +26,10 @@ from torch.utils.data import DataLoader, Dataset
 
 from lightly_train._configs import validate
 from lightly_train._data import cache
-from lightly_train._data._serialize import memory_mapped_sequence
-from lightly_train._data._serialize.memory_mapped_sequence import MemoryMappedSequence
+from lightly_train._data._serialize import memory_mapped_sequence_task
+from lightly_train._data._serialize.memory_mapped_sequence_task import (
+    MemoryMappedSequenceTask,
+)
 from lightly_train._data.mask_semantic_segmentation_dataset import (
     MaskSemanticSegmentationDataset,
     MaskSemanticSegmentationDatasetArgs,
@@ -360,9 +362,9 @@ def _decrement_and_cleanup_if_zero(mmap_file: Path, ref_file: Path) -> None:
 
 def get_dataset_mmap_filenames(
     fabric: Fabric,
-    filenames: Iterable[tuple[Path, Path]],
+    filepaths: Iterable[tuple[str, ...]],
     mmap_filepath: Path,
-) -> MemoryMappedSequence[str]:
+) -> MemoryMappedSequenceTask[str]:
     """Returns memory-mapped filenames shared across all ranks.
 
     Filenames are written to mmap_filepath by rank zero and read by all ranks.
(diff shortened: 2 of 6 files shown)


src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -407,13 +430,17 @@ def get_dataset(
     transform: TaskTransform,
     mmap_filepath: Path,
 ) -> MaskSemanticSegmentationDataset:
-    filenames = list(dataset_args.list_image_filenames())
+    image_and_mask_filepaths = list(
+        list_image_and_mask_filepaths(
+            image_dir=dataset_args.image_dir, mask_file=dataset_args.mask_dir_or_file
+        )
+    )
```
guarin: Two things here:
1. Remove the conversion to `list()` otherwise we have to load all file/mask paths into memory at once.
2. I would refactor this to `image_infos = dataset_args.liste_image_infos()`

The second change is needed because the `get_dataset` transform will also have to work with object detection and potentially other dataset types. `image_infos` is generic enough that it can be used for these dataset types. By implementing the method in `dataset_args` instead we also keep the code generic enough that segmentation and detection will call different `list_image_infos` methods automatically.

src/lightly_train/_data/_serialize/memory_mapped_sequence_task.py
guarin, line 1: Can we replace our existing memory mapped sequence with this one? Then we have a single implementation that is more general. I left some comments below how I think we could achieve this.

src/lightly_train/_data/_serialize/memory_mapped_sequence_task.py
```diff
@@ -0,0 +1,177 @@
...
+
+T = TypeVar("T")
+
+
+def write_filepaths_to_file(
+    filepaths: Iterable[tuple[str, ...]],
+    mmap_filepath: Path,
+    column_names: list[str],
+    chunk_size: int = 10_000,
+) -> None:
```
guarin:
```suggestion
def write_items_to_file(
    items: Iterable[tuple[bool | int | float | str, ...]],
    mmap_filepath: Path,
    column_names: list[str],
    chunk_size: int = 10_000,
) -> None:
```

src/lightly_train/_data/_serialize/memory_mapped_sequence_task.py
```diff
@@ -0,0 +1,177 @@
...
+
+    @overload
+    def __getitem__(self, index: int) -> tuple[T, ...]: ...
+
+    @overload
+    def __getitem__(self, index: slice) -> Sequence[tuple[T, ...]]: ...
+
+    def __getitem__(
+        self, index: int | slice
+    ) -> tuple[T, ...] | Sequence[tuple[T, ...]]:
```
guarin:
```suggestion
    @overload
    def __getitem__(self, index: int) -> T: ...

    @overload
    def __getitem__(self, index: slice) -> Sequence[T]: ...

    def __getitem__(
        self, index: int | slice
    ) -> T | Sequence[T]:
```

src/lightly_train/_data/_serialize/memory_mapped_sequence_task.py
```diff
@@ -0,0 +1,177 @@
...
+    mmap_filepath: Path, column_names: list[str]
+) -> MemoryMappedSequenceTask[str]:
+    table = _mmap_table_from_file(mmap_filepath=mmap_filepath)
+    logger.debug(
+        f"Creating memory mapped sequence with {table.num_rows} '{column_names}'."
+    )
+    return MemoryMappedSequenceTask(path=mmap_filepath, columns=column_names)
+
+
+class MemoryMappedSequenceTask(Sequence[Tuple[T, ...]], Generic[T]):
```
guarin: Change definition to:
```suggestion
K = TypeVar("K")
V = TypeVar("V")
T = dict[K, V]

class MemoryMappedSequence(Sequence[T], Generic[K, V]):
```

This way we can always return a dict with the column names and values.

src/lightly_train/_data/_serialize/memory_mapped_sequence_task.py
```diff
@@ -0,0 +1,177 @@
...
+        MemoryMappedSequenceTask.__init__(self, path=path, columns=columns)
+
+
+def _stream_write_table_to_file(
+    items: Iterable[tuple[str, ...]],
+    mmap_filepath: Path,
+    column_names: list[str],
+    chunk_size: int = 10_000,
+) -> None:
+    schema = pa.schema([(name, pa.string()) for name in column_names])
```
guarin: Instead of assuming `pa.string()` you can map the actual item type to the pyarrow type with something like:
```
ArrowTypes = {
    int: pa.int64(),
    str: pa.string(),
    # etc.
}
```

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -39,11 +42,14 @@ class MaskSemanticSegmentationDataset(Dataset[MaskSemanticSegmentationDatasetIte
     def __init__(
         self,
         dataset_args: MaskSemanticSegmentationDatasetArgs,
-        image_filenames: Sequence[ImageFilename],
+        image_and_mask_filepaths: Sequence[tuple[str, str]]
+        | MemoryMappedSequenceTask[str],
         transform: TaskTransform,
```
guarin: With the `MemoryMappedSequence` refactoring you should be able to change this to:
```
image_and_mask_filepaths: Sequence[dict[str, str]],
```

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -172,20 +180,19 @@ def __getitem__(self, index: int) -> MaskSemanticSegmentationDatasetItem:
...
-        ):
-            if (self.mask_dir / image_filename).with_suffix(".png").exists():
-                yield image_filename
+    @field_validator("mask_dir_or_file", mode="before")
+    @classmethod
+    def validate_mask_dir_or_file(cls, v: str) -> str:
+        if not bool(re.search(r"\{[^}]+\}", v)):  # mask_dir_or_file is a directory
+            return f"{v}/{{image_path.stem}}.png"
+        else:  # mask_dir_or_file is a format string
+            return v
```
guarin: This doesn't work for nested directories. Imagine the following scenario:
```
images/
    train/
        image1.jpg
masks/
    train/
        image1.png
```

with:
```
image_dir="images"
mask_dir="masks"
```

Then we would get:
```
f"{mask_dir}/{image_path.stem}.png"
# Which is:
f"masks/image1.png"
```

You would have to use `f"{v}/{{image_path.relative_to(image_dir).with_suffix('.png')}}"` but that doesn't work with `.format()`. If you want to use an f-string, I believe the only thing you can do is pass more arguments to `.format()`:

```
f"{v}/{{image_relpath_nosuffix}}.png"
```

And then compute `image_relpath_nosuffix = image_path.relative_to(image_dir).with_suffix("")` in the `list_image_and_mask_filepaths` code.

However, what I would propose instead is to move all the logic directly to `list_image_and_mask_filepaths` and remove `validate_mask_dir_or_file` here. Then you can do the following in `list_image_and_mask_filepaths`:
1. Check if `mask_file` is a directory.
2. If yes, then `mask_filepath = mask_file / image_path.relative_to(image_dir).with_suffix(".png")`
3. If no, then `mask_filepath = mask_file.format(image_path=image_path)`

tests/_data/test_mask_semantic_segmentation_dataset.py
```diff
@@ -32,6 +33,12 @@
 from .. import helpers
 
 
+@pytest.fixture
+def mask_file_pattern() -> str:
+    """Standard mask file pattern for semantic segmentation tests."""
+    return "{image_path.parent.parent}/masks/{image_path.stem}.png"
```
guarin: Don't create fixtures for constants. I would copy paste the constant to the relevant places, it makes the tests easier to read. Alternatively create a constant variable 
`MASK_FILE_PATTTERN="{image_path.parent.parent}/masks/{image_path.stem}.png"` but for tests it is nearly always better to hardcode these things.

src/lightly_train/_commands/train_task_helpers.py
yutong-xiang-97: I would use `list_filepaths` instead because essentially the mmap files contains filepaths for images and targets (masks etc.)

guarin: Yes but it might also contain other information (e.g. bounding boxes) in the near future.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Supported using custom mask names for the inputs in semantic segmentation.
+
```
guarin: Changelog messages are always in present tense

```suggestion
- Add support for using custom mask names for the inputs in semantic segmentation.
```

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -390,14 +409,19 @@ def get_dataset_mmap_filenames(
...
-        mmap_filepath=mmap_filepath
+            try:
+                memory_mapped_sequence_task.write_items_to_file(
+                    items=items,
+                    mmap_filepath=mmap_filepath,
+                )
+            except ValueError as e:
+                # Surface empty-sequence errors at this call site for clearer traces
+                # while preserving the original message.
+                raise ValueError(str(e)) from None
```
guarin: Remove the try/except and don't raise from `None` as this removes the stacktrace.

src/lightly_train/_data/_serialize/memory_mapped_sequence_task.py
```diff
@@ -0,0 +1,211 @@
...
+        raise ValueError(f"Invalid `chunk_size` {chunk_size} must be positive!")
+    logger.debug(f"Writing filepaths to '{mmap_filepath}' (chunk_size={chunk_size})")
+
+    it = iter(items)
+    try:
+        first_item = next(it)
+    except StopIteration:
+        raise ValueError(
+            "The memory-mapped sequence is empty because no valid entries are found. Please check your input data."
+        ) from None
```
guarin: Instead of raising an error I would use a default of `pa.string()`. IMO it is valid to have an empty mmap file. If we want to raise an error because of an empty input dataset we should do that in a top level function and not deep down in the mmap code.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -172,20 +174,32 @@ def __getitem__(self, index: int) -> MaskSemanticSegmentationDatasetItem:
...
-    def list_image_filenames(self) -> Iterable[ImageFilename]:
-        for image_filename in file_helpers.list_image_filenames(
-            image_dir=self.image_dir
+    def list_image_info(self) -> Iterable[dict[str, str]]:
+        for image_filepath in file_helpers.list_image_files(
+            imgs_and_dirs=[self.image_dir]
         ):
-            if (self.mask_dir / image_filename).with_suffix(".png").exists():
-                yield image_filename
+            if Path(self.mask_dir_or_file).is_dir():
```
guarin: Run this check before the for loop. Otherwise we have to convert str to Path and check existence of the mask directory for every file in the dataset.

tests/_data/_serialize/test_memory_mapped_sequence_task.py
```diff
@@ -0,0 +1,228 @@
...
+            ],
+            mmap_filepath=tmp_path / "test.arrow",
+        )
+        sequence = memory_mapped_sequence_task.memory_mapped_sequence_from_file(
+            mmap_filepath=tmp_path / "test.arrow",
+        )
+        assert len(sequence) == 3
+        copy = pickle.loads(pickle.dumps(sequence))
+        assert len(copy) == 3
+        assert sequence[:] == copy[:]
```
guarin, line 153:
```suggestion
        assert sequence == copy
```

tests/_data/_serialize/test_memory_mapped_sequence_task.py
```diff
@@ -0,0 +1,228 @@
...
+            "image_filepaths": str(image_dir / "image3.jpg"),
+            "mask_filepaths": str(mask_dir / "mask3.png"),
+        },
+    ]
+
+
+@pytest.mark.parametrize(
+    "chunk_size",
+    [0, -1],
+)
```
guarin: Nit: IMO not necessary to test.

tests/_data/_serialize/test_memory_mapped_sequence_task.py
yutong-xiang-97: This is because `MemoryMappedSequenceTask` does not implement `__eq__` and comparing the objects would raise an error. Using `list()` may also work but I think slices are indeed simpler (and possibly faster) in this case. It is adapted from the existing tests in `tests/_data/_serialize/test_memory_mapped_sequence.py`

PR #276, opened 2025-09-02
Fix mmap cache

guarin:
## What has changed and why?

* Change mmap cache behavior

This fixes an issue when multiple runs with the same data are scheduled on the same node. Before this PR the mmap files were based on the `data` directory. A mmap file could already exist for a given `data` directory if a previous run created and didn't delete it (e.g. if the run crashed). The new behavior is to use the `data` directory only if `LIGHTLY_TRAIN_MMAP_REUSE_FILE` is true, otherwise the `out` directory determines the mmap location. This is safer as only one run at a time can use a given `out` directory and we can error quickly if multiple runs with the same `out` directory are detected. This is not possible with the `data` directory as it is valid to have multiple runs using the same `data` directory at once.

## How has it been tested?

* Manual

- [x] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/common_helpers.py
+++ b/src/lightly_train/_commands/common_helpers.py
@@ -371,13 +371,14 @@ def _get_package(model: Module) -> BasePackage:
 @contextlib.contextmanager
 def get_dataset_temp_mmap_path(
     data: PathLike | Sequence[PathLike],
+    out: PathLike,
 ) -> Generator[Path, Any, Any]:
-    """Generate file in temporary directory to be used for memory-mapping the dataset.
+    """Generate file in the cache directory to be used for memory-mapping the dataset.
 
-    Creates a unique filename for the memory-mapped file based on the data arg.
-    We use the data arg as a deterministic value that is consistent across all ranks
-    on the same node. Additionally, we can cache the file if required, since the hash
-    directly reflects the used config.
+    Creates a unique filename for the memory-mapped file based on the `out` or `data`
+    arguments. We use those arguments as they are consistent across all ranks on the
+    same node for the same run. Additionally, we can cache the file if required, since
+    the hash directly reflects the used config.
 
     We need a deterministic value from "outside" at this point in the code as the
     code might already be running on multiple processes depending on how it was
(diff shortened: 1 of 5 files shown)


src/lightly_train/_commands/common_helpers.py
```diff
@@ -485,7 +516,8 @@ def get_dataset_mmap_filenames(
                     )
                 time.sleep(0.2)
     finally:
-        _unlink_and_ignore(tmp_path)
+        if distributed_helpers.is_local_rank_zero():
+            _unlink_and_ignore(tmp_path)
```
guarin, line 541: I ran into this issue. If a previous run created a mmap file, rank 0 would still recreate a new one but all other ranks would detect the old one and trigger the `finally` section which removes `tmp_path`. This would then lead to an error on line 498 because `tmp_path` might not exist anymore at that place.

src/lightly_train/_commands/common_helpers.py
```diff
@@ -390,12 +392,62 @@ def get_dataset_temp_mmap_path(
     The filename is different on each node. This is necessary to avoid multiple
     processes writing to the same file in case the nodes use a shared filesystem.
     """
-    out_hash = get_sha256(f"{data}-{distributed_helpers.get_node_rank() or 0}")
-    mmap_filepath = (cache.get_data_cache_dir() / out_hash).with_suffix(".mmap")
+    if Env.LIGHTLY_TRAIN_MMAP_REUSE_FILE.value:
+        # Use data as identifier to share the mmap file across multiple runs.
```
yutong-xiang-97, line 396: I just realized that if there is a previous training run with `LIGHTLY_TRAIN_MMAP_REUSE_FILE` set to `True` somehow crashed and left a **corrupted** mmap file identified by `data`, then a new run with `LIGHTLY_TRAIN_MMAP_REUSE_FILE` set to `True` would simply reuse this corrupted `data` file. AFAIK unfortunately we indeed do not have a cleanup procedure for crashed runs. How could we solve this?

src/lightly_train/_commands/common_helpers.py
yutong-xiang-97: Just for better understanding: it seems that adding `distributed_helpers.is_local_rank_zero()` here is sufficient for solving the problem described here. Using `out_dir` as the hash for concurrent runs is simply to make it safer. Am I correct?

guarin: Exactly

src/lightly_train/_commands/common_helpers.py
guarin: The mmap file only exists if it is uncorrupt. This is because `tmp_path.replace(mmap_filepath.resolve())` is an atomic operation. If it is executed the file is uncorrupt. If it is not executed the run might have crashed but then the mmap file doesn't exist.

However, if the user changes contents of the data directory between two runs then the mmap file with the old data content will be loaded. This is the risk of re-using the file. The only way to solve that would be to check if the underlying data has changed.

yutong-xiang-97 (approved):

PR #278, opened 2025-09-04
Support Multi-channel semantic segmentation masks

yutong-xiang-97:
## What has changed and why?

Add support for using multi-channel masks for semantic segmentation.

1. Introduce `SingleChannelClassInfo` and `MultiChannelClassInfo` classes to handle different mask types. They can be specified under the `values` key in the `classes` dict when calling the training command
  - `SingleChannelClassInfo`: compatible with the previous `ClassInfo` class
  - `MultiChannelClassInfo`: accepts a list of tuples to represent channel values.
2. Convert multi-channel masks to single-channel mask to be compatible with the existing pipeline.
3. Fix: 
 - raise `ValueError` when mask and the image have the same shape instead of assertion cuz it would indicate issues with provided masks (external input).
 - `remap_mask` to handle the negative `ignore_index` introduced when doing transforms which resulted in index errors in the lookup table. The fix assumes the non-ignored classes are all non-negative, though. (TODO: add tests for `remap_mask()` later.)
5. Add corresponding tests.

### Failure Cases Handling
 
1. **Duplicate Mappings**: Prevents mapping the same channel/label value to multiple classes with detailed error messages (same behavior as the existing solution).  
2. **Type Consistency**: Ensures all classes are either channel values or integer labels, no mix.  
6. **Missing Class Definitions**: Requires channel values to be specified in `classes` when multi-channel masks are detected.  

## How Has It Been Tested?

- Unit tests
- MLflow runs on CamVid data on the compute machine

- [x] Yes

- [ ] Yes
- [x] Not needed (will be added later)

### DIFF
--- a/src/lightly_train/_data/mask_semantic_segmentation_dataset.py
+++ b/src/lightly_train/_data/mask_semantic_segmentation_dataset.py
@@ -142,7 +142,7 @@ class MaskSemanticSegmentationDataset(Dataset[MaskSemanticSegmentationDatasetIte
         }
         return binary_masks
 
-    def remap_mask(self, mask: torch.Tensor) -> torch.Tensor:
+    def map_class_id_to_internal_class_id(self, mask: torch.Tensor) -> torch.Tensor:
         # Map only non-ignored pixels through the LUT and keep ignored pixels untouched.
         # NOTE: this relies on the fact that all `old_class`es are non-negative.
         ignore = mask == self.ignore_index
@@ -254,7 +254,7 @@ class MaskSemanticSegmentationDataset(Dataset[MaskSemanticSegmentationDatasetIte
 
         # Mark pixels to ignore in the masks.
         # TODO(Thomas, 07/25): Make this optional.
-        transformed_mask = self.remap_mask(transformed["mask"])
+        transformed_mask = self.map_class_id_to_internal_class_id(transformed["mask"])
 
         return {
             "image_path": str(image_path),  # Str for torch dataloader compatibility.


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Add support for using RGB masks for the inputs in semantic segmentation.
 - Add support for using custom mask names for the inputs in semantic segmentation.
 - Add `precision` flag to ONNX export task to specify if we export with float16 or float32 precision.
```
guarin: These points should actually all go under `### Added`

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -25,13 +27,84 @@
...
+        elif isinstance(v, list):
+            return set(v)
+        # Single label: 0 -> {0}
+        elif isinstance(v, int):
+            return {v}
+        else:
+            raise ValueError(f"Expected int or list of ints, got {type(v)}")
+
+
+class ColorsClassInfo(PydanticConfig):
```
guarin:
```suggestion
class MultiChannelClassInfo(PydanticConfig):
```

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -25,13 +27,84 @@
 from lightly_train.types import (
     BinaryMasksDict,
     MaskSemanticSegmentationDatasetItem,
+    NDArrayImage,
     PathLike,
 )
 
 
-class ClassInfo(PydanticConfig):
+class LabelsClassInfo(PydanticConfig):
```
guarin:
```suggestion
class SingleChannelClassInfo(PydanticConfig):
```

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -142,9 +234,34 @@ def __getitem__(self, index: int) -> MaskSemanticSegmentationDatasetItem:
...
+        if image.shape[:2] != mask.shape[:2]:
+            raise ValueError(
+                f"Shape mismatch: image shape is {image.shape[:2]} while mask shape is {mask.shape}."
+            )
+
+        # Local alias to enable type narrowing with TypeGuard
+        classes = self.args.classes
+
+        # Check that if the mask is RGB, then the class info must be ColorsClassInfo
+        if len(mask.shape) == 3 and mask.shape[2] == 3:
```
guarin:
```suggestion
        if len(mask.shape) == 3:
```

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -25,13 +27,57 @@
...
+
+    @field_validator("labels", mode="before")
+    @classmethod
+    def normalize_labels(cls, v: set[int] | list[int]) -> set[int]:
+        if isinstance(v, set):
+            return v
+        elif isinstance(v, list):
+            return set(v)
+        else:
+            raise ValueError(f"Expected set or list of ints, got {type(v)}")
```
guarin:
```suggestion
    labels: set[int] = Field(alias="values", strict=False)
```

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -25,13 +27,57 @@
...
+    @classmethod
+    def normalize_channels(
+        cls, v: set[tuple[int, ...]] | list[tuple[int, ...]]
+    ) -> set[tuple[int, ...]]:
+        if isinstance(v, set):
+            return v
+        elif isinstance(v, list):
+            return set(v)
+        else:
+            raise ValueError(f"Expected set or list of tuple of ints, got {type(v)}")
```
guarin:
```suggestion
    labels: set[tuple[int, ...]] = Field(alias="values", strict=False)
```

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -115,17 +161,46 @@ def get_binary_masks(self, mask: Tensor) -> BinaryMasksDict:
         return binary_masks
 
     def remap_mask(self, mask: torch.Tensor) -> torch.Tensor:
-        # Create a lookup table initialized with ignore_index
-        max_class = int(mask.max().item())
+        # Map only non-ignored pixels through the LUT and keep ignored pixels untouched.
+        # NOTE: this relies on the fact that all `old_class`es are non-negative.
+        ignore = mask == self.ignore_index
+        valid = ~ignore
```
guarin: Why are the changes in this function needed?

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -115,17 +161,46 @@ def get_binary_masks(self, mask: Tensor) -> BinaryMasksDict:
...
-        return lut[mask.to(torch.long)]
+        mask = mask.to(torch.long)
+        mask[valid] = lut[mask[valid]]
+
+        return mask
+
+    def _map_multi_channel_masks_to_single_channel(
+        self,
+        multi_channel_mask: NDArrayImage,
+        class_infos: dict[int, MultiChannelClassInfo],
```
guarin: No need to pass `class_infos` you can access it with `self.class_infos`

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -142,9 +217,33 @@ def __getitem__(self, index: int) -> MaskSemanticSegmentationDatasetItem:
         )
 
         # Verify that the mask and the image have the same shape.
-        assert image.shape[:2] == mask.shape, (
-            f"Shape mismatch: image shape is {image.shape[:2]} while mask shape is {mask.shape}."
-        )
+        if image.shape[:2] != mask.shape[:2]:
+            raise ValueError(
+                f"Shape mismatch: image shape is {image.shape[:2]} while mask shape is {mask.shape}."
```
guarin:
```suggestion
                f"Shape mismatch: image (height, width) is {image.shape[:2]} while mask (height, width) is {mask.shape[:2]}."
```

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -142,9 +217,33 @@ def __getitem__(self, index: int) -> MaskSemanticSegmentationDatasetItem:
...
+                    "For multi-channel masks, you have to specify the channel value that correspond to each class.\n\n"
+                    "The channel values should be provided as a list of tuples to `values`:\n"
+                    "classes = {\n"
+                    "  0: {'name': 'background', 'values': [(0, 0, 0)]},\n"
+                    "  1: {'name': 'road', 'values': [(128, 128, 128)]}\n"
+                    "}\n\n"
+                    "classes = {\n"
+                    "  0: {'name': 'background', 'values': [(0, 0, 0), (255, 255, 255)]},\n"
+                    "  1: {'name': 'road', 'values': [(128, 128, 128), (64, 64, 64)]}\n"
+                    "}"
```
guarin: Let's use `labels` everywhere as `values` is deprecated. 
```suggestion
                    "The channel values must be provided as a list of tuples to `labels`:\n"
                    "classes = {\n"
                    "  0: {'name': 'background', 'labels': [(0, 0, 0)]},\n"
                    "  1: {'name': 'road', 'labels': [(0, 128, 128)]}\n"
                    "}\n\n"
                    "classes = {\n"
                    "  0: {'name': 'background', 'labels': [(0, 0, 0), (255, 255, 255)]},\n"
                    "  1: {'name': 'road', 'labels': [(0, 128, 128), (64, 64, 64)]}\n"
                    "}"
```

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -142,9 +217,33 @@ def __getitem__(self, index: int) -> MaskSemanticSegmentationDatasetItem:
...
+            )
+
+        # Local alias to enable type narrowing with TypeGuard
+        classes = self.args.classes
+
+        # Check that if the mask is multi-channel, then the class info must be MultiChannelClassInfo
+        if len(mask.shape) == 3:
+            if not _are_multi_channel_classes(classes):
+                raise ValueError(
+                    "Expected channel values specified in `classes` for multi-channel masks but got labels. "
```
guarin:
```suggestion
                    "Expected channel values specified in `classes` for multi-channel masks but got single-channel labels. "
```

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -115,17 +161,46 @@ def get_binary_masks(self, mask: Tensor) -> BinaryMasksDict:
         return binary_masks
 
     def remap_mask(self, mask: torch.Tensor) -> torch.Tensor:
-        # Create a lookup table initialized with ignore_index
-        max_class = int(mask.max().item())
+        # Map only non-ignored pixels through the LUT and keep ignored pixels untouched.
+        # NOTE: this relies on the fact that all `old_class`es are non-negative.
```
guarin, line 147: Can we rename this function to `map_class_id_to_internal_class_id`.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -152,9 +227,33 @@ def __getitem__(self, index: int) -> MaskSemanticSegmentationDatasetItem:
...
+                    "  1: {'name': 'road', 'values': [(128, 128, 128)]}\n"
+                    "}\n\n"
+                    "classes = {\n"
+                    "  0: {'name': 'background', 'values': [(0, 0, 0), (255, 255, 255)]},\n"
+                    "  1: {'name': 'road', 'values': [(128, 128, 128), (64, 64, 64)]}\n"
+                    "}"
+                )
+
+            # Map multi-channel mask to single channel mask using class labels
+            mask = self._map_multi_channel_masks_to_single_channel(mask, classes)
```
guarin: Where do we map single channel labels to class id?

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -232,43 +330,80 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
...
+                    "classes = {\n"
+                    "  0: {'name': 'background', 'values': [0, 1, 2]},\n"
+                    "  1: {'name': 'road', 'values': [3, 4, 5]}\n"
+                    "}\n\n"
+                    "CORRECT (use only channel tuples):\n"
+                    "classes = {\n"
+                    "  0: {'name': 'background', 'values': [(0, 0, 0), (255, 255, 255)]},\n"
+                    "  1: {'name': 'road', 'values': [(128, 128, 128), (64, 64, 64)]}\n"
+                    "}"
+                )
```
guarin: Use labels everywhere. Also adjust in the text and the other error messages further below.

```suggestion
                    "classes = {\n"
                    "  0: {'name': 'background', 'labels': [0, 1, 2]},\n"
                    "  1: {'name': 'road', 'labels': [(0, 0, 0), (128, 128, 128)]}  # <- mixed types\n"
                    "}\n\n"
                    "CORRECT (use only integer labels):\n"
                    "classes = {\n"
                    "  0: {'name': 'background', 'values': [0, 1, 2]},\n"
                    "  1: {'name': 'road', 'values': [3, 4, 5]}\n"
                    "}\n\n"
                    "CORRECT (use only channel tuples):\n"
                    "classes = {\n"
                    "  0: {'name': 'background', 'labels': [(0, 0, 0), (255, 255, 255)]},\n"
                    "  1: {'name': 'road', 'labels': [(128, 128, 128), (64, 64, 64)]}\n"
                    "}"
                )
```

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -232,43 +330,80 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
...
+                            f"INCORRECT (class label {label} is duplicated):\n"
+                            f"classes = {{\n"
+                            f"  255: {{'name': 'background-255', 'values': [0, 1, 2]}},\n"
+                            f"  254: {{'name': 'background-254', 'values': [0, 1, 2]}}  # <- class [0, 1, 2] conflict with class 255\n"
+                            f"}}\n\n"
+                            f"CORRECT (each label value belongs to only one class):\n"
+                            f"classes = {{\n"
+                            f"  255: {{'name': 'background-255', 'values': [0, 1, 2]}},\n"
+                            f"  254: {{'name': 'background-254', 'values': [3, 4, 5]}}  # <- unique values\n"
+                            f"}}"
```
guarin: Is there a reason for the weird class names `255`, `254` and `background-255`, `background-254`?  Could we use the same classes as in the other error messages?

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -232,43 +330,80 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
     @field_validator("classes", mode="before")
     @classmethod
     def validate_classes(
-        cls, classes: dict[int, str | dict[str, str | Sequence[int]]]
+        cls, classes: dict[int, str | dict[str, Any]]
     ) -> dict[int, ClassInfo]:
-        # Let Pydantic validate the structure and types
         classes_validated = TypeAdapter(
-            Dict[int, Union[str, ClassInfo]]
+            Dict[int, Union[str, SingleChannelClassInfo, MultiChannelClassInfo]]
```
guarin, line 320:
```suggestion
            Union[Dict[int, Union[str, SingleChannelClassInfo]], Dict[int, MultiChannelClassInfo]]
```

This way pydantic validates that you don't mix single- and multi-channel class infos. And you won't have to worry about that anymore for the rest of the code.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
yutong-xiang-97: I met errors when training with CamVid. It is because that some transforms (e.g. `RandomCrop`) fill in the padding value with ignore index (-100). This would result in the LUT indexing to fail with index out-of-bounds. Negative index shouldn't be used for the indexing there at all.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
yutong-xiang-97: Nope. I thought this was the default. I can use `background` and `road` instead.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
yutong-xiang-97: This does not work with the TypeAdapter (validation issues), but the above implementation for class infos are already sufficient.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
yutong-xiang-97: This syntax does not work, cuz it only accepts `"values"` in the dict input.

Instead, for our desired behavior, we should go with:

```python
class SingleChannelClassInfo(PydanticConfig):
    name: str
    labels: set[int] = Field(
        validation_alias=AliasChoices("labels", "values"),
        serialization_alias="labels",
        strict=False,
    )


class MultiChannelClassInfo(PydanticConfig):
    name: str
    labels: set[tuple[int, ...]] = Field(
        validation_alias=AliasChoices("labels", "values"),
        serialization_alias="labels",
        strict=False,
    )
```

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
yutong-xiang-97: Same as above

tests/_data/test_mask_semantic_segmentation_dataset.py
```diff
@@ -285,7 +486,152 @@ def test__getitem__(
...
+        image_dir = tmp_path / "images"
+        mask_dir = tmp_path / "masks"
+
+        # Create image with one size
+        image_dir.mkdir(parents=True, exist_ok=True)
+        helpers.create_image(image_dir / "image0.jpg", height=64, width=64)
+
+        # Create mask with different size
+        mask_dir.mkdir(parents=True, exist_ok=True)
+        helpers.create_mask(
```
guarin: No need to call `.mkdir`, the directories and files will be automatically created in `create_image` and `create_mask` respectively.

guarin (approved):

PR #279, opened 2025-09-04
Add multi-channel support

guarin:
## What has changed and why?

* Add multi-channel support with `transform_args={"num_channels": <num channels>}`
* Add DINOv2, DINOv3, and TIMM model multi-channel support
* Add multi-channel support for all pretraining methods
* Add multi-channel support for all fine-tuning methods
* It is no longer necessary to set `LIGHTLY_TRAIN_IMAGE_MODE` when loading images with more than 3 channels

Multi-channel support is currently limited to 4 channel images because we use PIL for image loading. We'll add support for TIFF/DICOM with more than 4 channels in a follow-up PR.

The PR looks huge but is mostly passing `num_channels` variables around. I used `num_channels` when the code is related to the image data (e.g. dataset, transforms, etc.). And `num_input_channels` when the code is in a model context. This is to disambiguate between input and output channels in a model.

## How has it been tested?

* Added integration tests
* Manual tests

- [ ] Yes

- [ ] Yes

### DIFF
--- a/tests/_data/test_mask_semantic_segmentation_dataset.py
+++ b/tests/_data/test_mask_semantic_segmentation_dataset.py
@@ -34,20 +34,21 @@ from .. import helpers
 
 
 def _dummy_transform(num_channels: int = 3) -> SemanticSegmentationTransform:
-    return SemanticSegmentationTransform(
-        SemanticSegmentationTransformArgs(
-            ignore_index=-100,
-            image_size=(32, 32),
-            channel_drop=None,
-            num_channels=num_channels,
-            normalize=NormalizeArgs(),
-            random_flip=None,
-            color_jitter=None,
-            scale_jitter=None,
-            smallest_max_size=SmallestMaxSizeArgs(max_size=32, prob=1.0),
-            random_crop=None,
-        )
+    args = SemanticSegmentationTransformArgs(
+        ignore_index=-100,
+        image_size=(32, 32),
+        channel_drop=None,
+        num_channels=num_channels,
+        normalize=NormalizeArgs(),
+        random_flip=None,
+        color_jitter=None,
+        scale_jitter=None,
+        smallest_max_size=SmallestMaxSizeArgs(max_size=32, prob=1.0),
+        random_crop=None,
     )
+    args.resolve_auto()
+    args.resolve_incompatible()
+    return SemanticSegmentationTransform(args)
 
 
 class TestMaskSemanticSegmentationDataArgs:


src/lightly_train/_commands/train_helpers.py
```diff
@@ -60,9 +60,15 @@ def get_transform_args(
...
-        return transform_args_cls()  # type: ignore[call-arg]
+        transform_args = transform_args_cls()  # type: ignore[call-arg]
+    else:
+        transform_args = validate.pydantic_model_validate(
+            transform_args_cls, transform_args
+        )
 
-    return validate.pydantic_model_validate(transform_args_cls, transform_args)
+    transform_args.resolve_auto()
+    transform_args.resolve_incompatible()
```
guarin, line 68: Had to add `resolve_auto` because `num_channels` is by default `"auto"` in the transforms. This is nice if you set a custom `normalize` or `channel_drop` augmentation because then the number of channels are automatically inferred.

Similarly, I had to add `resolve_incompatible` because there are now some situations where transforms don't work if `num_channels != 3`. Check `semantic_segmentation_transforms.py` for details.

src/lightly_train/_data/image_dataset.py
```diff
@@ -27,20 +27,31 @@ def __init__(
...
+        self.num_channels = num_channels
 
-        try:
-            self.image_mode = ImageMode(Env.LIGHTLY_TRAIN_IMAGE_MODE.value)
-        except ValueError:
+        image_mode = (
+            None
+            if Env.LIGHTLY_TRAIN_IMAGE_MODE.value is None
+            else ImageMode(Env.LIGHTLY_TRAIN_IMAGE_MODE.value)
+        )
```
guarin, line 43: Had to change logic because `LIGHTLY_TRAIN_IMAGE_MODE` is now `None` by default. It is new `None` because we want to infer the image mode from `num_channels` if possible and only overwrite by `LIGHTLY_TRAIN_IMAGE_MODE` if it was explicitly set by the user.

We might also be able to remove this env variable completely in the future.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -51,20 +54,27 @@ def __init__(
...
-        image_mode = Env.LIGHTLY_TRAIN_IMAGE_MODE.value
-        if image_mode not in ("RGB", "UNCHANGED"):
+        transform_args = transform.transform_args
+        assert isinstance(transform_args, SemanticSegmentationTransformArgs)
+
+        image_mode = (
+            None
+            if Env.LIGHTLY_TRAIN_IMAGE_MODE.value is None
+            else ImageMode(Env.LIGHTLY_TRAIN_IMAGE_MODE.value)
+        )
```
guarin, line 63: Same change as for image dataset.

src/lightly_train/_methods/densecl/densecl.py
```diff
@@ -156,12 +156,14 @@ def __init__(
         optimizer_args: OptimizerArgs,
         embedding_model: EmbeddingModel,
         global_batch_size: int,
+        num_input_channels: int,
```
guarin, line 159: Had to add `num_input_channels` to the method initialization because distillation methods rely on it to initialize the teacher models correctly.

src/lightly_train/_models/_model_helpers.py
```diff
@@ -0,0 +1,55 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def patch_embed_adjust_input_channels_hook(
+    module: Module,
+    state_dict: dict[str, Any],
+    prefix: str,
+    *args: Any,
+    **kwargs: Any,
```
guarin, line 24: Used in DINOv2 and DINOv3 vision transformer to automatically adjust the number of input channels of the model.

src/lightly_train/_models/dinov3/dinov3_package.py
```diff
@@ -74,13 +74,18 @@ def parse_model_name(cls, model_name: str) -> str:
...
+        model_name: str,
+        num_input_channels: int = 3,
+        model_args: dict[str, Any] | None = None,
     ) -> DinoVisionTransformer:
         """
         Get a DINOv3 ViT model by name. Here the student version is build.
         """
-        model_args = {} if model_args is None else model_args
-        model = MODEL_NAME_TO_GETTER[model_name](**model_args)
+        args: dict[str, Any] = {"in_chans": num_input_channels}
```
guarin, line 85: It is called `in_chans` inside the DINOv2 VisionTransformer (same for DINOv3 and all TIMM models).

src/lightly_train/_models/dinov3/dinov3_src/hub/backbones.py
```diff
@@ -217,6 +217,7 @@ def dinov3_vits16(
     pretrained: bool = True,
     weights: Union[Weights, str] = Weights.LVD1689M,
     check_hash: bool = False,
+    in_chans: int = 3,
```
guarin, line 220: Had to add default here to be able to overwrite the value.

src/lightly_train/_plot.py
```diff
@@ -60,7 +60,7 @@ def plot_example_augmentations(train_batch: Batch, max_examples: int = 10) -> PI
                 :,
                 x_start:x_end,
                 y_start:y_end,
-            ] = image_tensor.cpu()
+            ] = image_tensor[:3].cpu()  # Take only first 3 channels.
```
guarin, line 63: This is the code that plots the example augmentations in the logs. Here we cannot use more than 3 channels.

src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/task_model.py
```diff
@@ -45,7 +45,7 @@ def __init__(
         classes: dict[int, str],
         class_ignore_index: int | None,
         image_size: tuple[int, int],
-        image_normalize: dict[str, float],
+        image_normalize: dict[str, tuple[float, ...]],
```
guarin, line 48: This type was wrong

src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/transforms.py
```diff
@@ -69,9 +69,10 @@ class DINOv2EoMTSemanticSegmentationTrainTransformArgs(
 
     image_size: tuple[int, int] = (518, 518)
     channel_drop: ChannelDropArgs | None = None
+    num_channels: int | Literal["auto"] = "auto"
     normalize: NormalizeArgs = Field(default_factory=NormalizeArgs)
     random_flip: RandomFlipArgs = Field(default_factory=RandomFlipArgs)
-    color_jitter: DINOv2EoMTSemanticSegmentationColorJitterArgs = Field(
+    color_jitter: DINOv2EoMTSemanticSegmentationColorJitterArgs | None = Field(
```
guarin, line 75: Had to add `None` to be able to disable it because color jitter doesn't support more than 3 channels

src/lightly_train/_transforms/transform.py
```diff
@@ -169,6 +174,58 @@ class MethodTransformArgs(PydanticConfig):
...
+        if self.num_channels == "auto":
+            if self.channel_drop is not None:
+                self.num_channels = self.channel_drop.num_channels_keep
+            else:
+                self.num_channels = len(self.normalize.mean)
+
+    def resolve_incompatible(self) -> None:
+        # Adjust normalization mean and std to match num_channels.
+        if len(self.normalize.mean) != no_auto(self.num_channels):
+            logger.debug(
```
guarin, line 187: Same as for task transforms. The only difference is that we disable more transforms here.

tests/_commands/test_train_task.py
```diff
@@ -36,15 +37,27 @@
...
     ),
 )
-def test_train_semantic_segmentation(tmp_path: Path) -> None:
+@pytest.mark.parametrize(
+    "model_name, model_args",
+    [
+        # Reduce number of joint blocks _vittest14.
+        ("dinov2/_vittest14-eomt", {"num_joint_blocks": 1}),
+        ("dinov2/_vittest14-linear", {}),
+    ],
```
guarin, line 48: Couldn't test dinov3 because we don't have a test model yet.

tests/_data/test_mask_semantic_segmentation_dataset.py
```diff
@@ -22,32 +21,33 @@
...
-    def __init__(self, transform_args: TaskTransformArgs):
-        super().__init__(transform_args=transform_args)
-        self.transform = A.Compose(
-            [
-                A.Resize(32, 32),
-                A.Normalize(mean=(0.5, 0.5, 0.5), std=(0.5, 0.5, 0.5)),
-                A.pytorch.transforms.ToTensorV2(),
-            ]
+def _dummy_transform() -> SemanticSegmentationTransform:
+    return SemanticSegmentationTransform(
```
guarin: Required because I changed the transform type to `SemanticSegmentationTransform` in `MaskSemanticSegmentationDataset(transform: SemanticSegmentationTransform)`. It was easier to return a true SemanticSegmentationTransform than to create a new dummy class.

src/lightly_train/_commands/train_helpers.py
```diff
@@ -51,6 +51,7 @@ def get_transform_args(
     logger.debug(f"Getting transform args for method '{method}'.")
     logger.debug(f"Using additional transform arguments {transform_args}.")
     if isinstance(transform_args, MethodTransformArgs):
+        transform_args.resolve_auto()
```
yutong-xiang-97: Maybe also `transform_args.resolve_incompatible()` here?

src/lightly_train/_models/ultralytics/ultralytics_package.py
```diff
@@ -93,6 +96,11 @@ def get_model(
                 f"Cannot create model '{model_name}' because '{cls.name}' is not "
                 "installed."
             )
+        if num_input_channels != 3:
```
yutong-xiang-97, line 99: Ultralytics actually supports multi-channel input starting from v8.3.112:

- https://github.com/ultralytics/ultralytics/pull/20223
- https://docs.ultralytics.com/datasets/detect/coco8-multispectral/

src/lightly_train/_models/rfdetr/rfdetr_package.py
```diff
@@ -61,6 +64,11 @@ def get_model(
             raise ValueError(
                 f"Cannot create model '{model_name}' because rfdetr is not installed."
             )
+        if num_input_channels != 3:
```
yutong-xiang-97, line 67: RF-DETR is actively working on multi-channel support: https://github.com/roboflow/rf-detr/pull/180

src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/transforms.py
```diff
@@ -69,9 +69,10 @@ class DINOv2LinearSemanticSegmentationTrainTransformArgs(
 
     image_size: tuple[int, int] = (518, 518)
     channel_drop: ChannelDropArgs | None = None
+    num_channels: int | Literal["auto"] = "auto"
     normalize: NormalizeArgs = Field(default_factory=NormalizeArgs)
-    random_flip: RandomFlipArgs = Field(default_factory=RandomFlipArgs)
-    color_jitter: DINOv2LinearSemanticSegmentationColorJitterArgs = Field(
+    random_flip: RandomFlipArgs | None = Field(default_factory=RandomFlipArgs)
```
yutong-xiang-97, line 74: You don't have to disable `random_flip`

src/lightly_train/_task_models/dinov3_eomt_semantic_segmentation/transforms.py
```diff
@@ -70,6 +70,7 @@ class DINOv3EoMTSemanticSegmentationTrainTransformArgs(
     # TODO(Guarin, 08/25): Check if we should change default to 512.
     image_size: tuple[int, int] = (518, 518)
     channel_drop: ChannelDropArgs | None = None
+    num_channels: int | Literal["auto"] = "auto"
     normalize: NormalizeArgs = Field(default_factory=NormalizeArgs)
     random_flip: RandomFlipArgs = Field(default_factory=RandomFlipArgs)
     color_jitter: DINOv3EoMTSemanticSegmentationColorJitterArgs = Field(
```
yutong-xiang-97: Color jitter should be disabled for multi-channel

src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/task_model.py
yutong-xiang-97: The same thing should be applied to DINOv3 EoMT model

src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/task_model.py
```diff
@@ -110,6 +110,7 @@ def __init__(
         # Disable drop path by default.
         backbone_model_args = {
             "drop_path_rate": 0.0,
+            "in_chans": len(self.image_normalize["mean"]),
         }
```
yutong-xiang-97, line 114: The same thing should be applied to DINOv3 EoMT model

tests/_data/test_image_dataset.py
```diff
@@ -156,6 +157,7 @@ def test___getitem____mode(self, tmp_path: Path, mode: str, extension: str) -> N
             image_dir=image_dir,
             image_filenames=filenames,
             transform=DummyMethodTransform(),
+            num_channels=3,
```
yutong-xiang-97, line 169: Here you may also want to test multi-channel modes (also pass the correct `num_channels` to `_create_images()`)

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -50,20 +53,27 @@ def __init__(
...
+        if image_mode is None:
```
yutong-xiang-97, line 64: You may also want to add a corresponding test for image modes and multi-channel images for `__getitem__` to make sure it loads properly.

tests/_models/test_package_helpers.py
```diff
@@ -49,20 +49,24 @@ def test_get_model__rfdetr() -> None:
...
+    )
     assert isinstance(model.get_model(), ResNet)
 
 
 def test_get_model__timm() -> None:
     pytest.importorskip("timm")
     from timm.models.resnet import ResNet
 
-    model = package_helpers.get_wrapped_model("timm/resnet18")
+    model = package_helpers.get_wrapped_model("timm/resnet18", num_input_channels=3)
```
yutong-xiang-97: You may also want to add tests for models that supports multi-channel

src/lightly_train/_transforms/transform.py
```diff
@@ -169,6 +174,58 @@ class MethodTransformArgs(PydanticConfig):
     gaussian_blur: GaussianBlurArgs | None
     solarize: SolarizeArgs | None
 
+    def resolve_auto(self) -> None:
+        if self.num_channels == "auto":
```
yutong-xiang-97, line 178: Here, if the actual input images have > 3 channels but the user specifies `num_channels` as `auto` & does NOT specify normalization (default to 3-chan) & without channel drop, then `self.num_channels` is set to 3. This would not trigger the incompatibility message below, and would affect all future operations based on `transform_args.num_channels`. We should somehow validate this in the data set iteration when we load the actual image.

src/lightly_train/_data/image_dataset.py
```diff
@@ -27,20 +27,31 @@ def __init__(
...
+        if image_mode is None:
+            image_mode = (
```
yutong-xiang-97, line 45: An example for my message concerning `transform_args.num_channels`: if `num_channels` is accidentally missing for a >3-chan input, then here the `num_channels` is 3 which will set the image_mode to RGB!

yutong-xiang-97 (commented):
Mostly good. I left some comments.

The biggest issue is that when `num_channels` is `auto` with a default normalization in `transform_args`. 

The suggestions for adding multi-channel to some of the tests are optional. I believe if the high-level tests for `.train()` or `train_sem_seg()` work then we should be fine.

Btw don't forget the changelog :)

src/lightly_train/_models/ultralytics/ultralytics_package.py
guarin: Good to know! We can enable that in a follow-up.

src/lightly_train/_models/rfdetr/rfdetr_package.py
guarin: Same as for Ultralytics, let's enable once it is supported

src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/transforms.py
guarin: Yes not in this PR but in general it should be possible to be disabled.

src/lightly_train/_transforms/transform.py
guarin: We expect that the user provides information about multi-channel images at least in one place. Either by setting `num_channels=4` which is the recommended way. Or by setting a custom `normalize` or `crop_channel` transform. If they don't set anything then we assume 3-channel RGB images. This is actually useful because then images are automatically converted to RGB if possible. For example, RGBA images can be converted to RGB without problems.

yutong-xiang-97: OK, if using RGB is the fallback behavior, then we can go on like this. I would instead make sure it is super clear in the docs to prevent users from forgetting to specify `num_channel` as they might have intended to.

PR #286, opened 2025-09-09
Use flash attention for dinov2 eomt segmentation task

simonschoelly:
## What has changed and why?

Replaces our custom attention code for the DinoV2 EOMT segmentation task with torch's `scaled_dot_product_attention`.

## How has it been tested?

So far only tested if exported ONNX models look the same

- [ ] Yes

- [ ] Yes


src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/task_model.py
```diff
@@ -527,23 +528,39 @@ def _attn(self, module: Attention, x: Tensor, mask: Tensor | None) -> Tensor:
...
 
+        embedding_dim = C // module.num_heads
         qkv = (
             module.qkv(x)
-            .reshape(B, N, 3, module.num_heads, C // module.num_heads)
+            .reshape(B, N, 3, module.num_heads, embedding_dim)
             .permute(2, 0, 3, 1, 4)
         )
-        q, k, v = qkv[0] * module.scale, qkv[1], qkv[2]
+        q, k, v = qkv[0], qkv[1], qkv[2]
```
stegmuel, line 535: AFAIK when scale is set to None the default value is used, namely: 1. / sqrt(embed_dim).
Similarly, `module.scale = sqrt(embed_dim)`  (see [here](https://github.com/lightly-ai/lightly-train/blob/9a0dbc03c49a74aaca9899fe392d7be2b24e1d13/src/lightly_train/_models/dinov2_vit/dinov2_vit_src/layers/attention.py#L41)) so `factor=1`, no?

Overall, why not:
- never pre-scale the queries or keys  and
- always set scale to None?

guarin: We might also be able to leave the scale at `None` as default. We don't use it in DINOv3 either and it works just fine: https://github.com/lightly-ai/lightly-train/blob/9a0dbc03c49a74aaca9899fe392d7be2b24e1d13/src/lightly_train/_task_models/dinov3_eomt_semantic_segmentation/task_model.py#L554

In my test on https://github.com/lightly-ai/lightly-train/compare/main...dinov3-eomt-test I also set `scale=None` and didn't see any performance difference.

simonschoelly: Yes, I was not sure if we always set `module.scale = sqrt(embed_dim)` - but if that is the case I am very happy to set scale to `None` and remove that workaround.

simonschoelly: Changed

PR #287, opened 2025-09-09
Memory-map file follow-up

yutong-xiang-97:
## What has changed and why?

Use the `MemoryMappedSequence` class for the segmentation task also for pretraining

## How has it been tested?

- Unit tests
- training runs on compute machine

- [ ] Yes

- [ ] Yes


src/lightly_train/_data/file_helpers.py
```diff
@@ -38,10 +38,10 @@ class ImageMode(Enum):
     MASK = "MASK"
 
 
-def list_image_filenames_from_iterable(
+def list_image_filepaths_from_iterable(
     imgs_and_dirs: Iterable[PathLike],
 ) -> Iterable[ImageFilename]:
-    """List image files recursively from the given list of image files and directories.
+    """List image filepaths recursively from the given list of image files and directories.
```
guarin:
```suggestion
    """List image files recursively from the given list of image files and directories.
```

src/lightly_train/_data/file_helpers.py
```diff
@@ -50,7 +50,7 @@ def list_image_filenames_from_iterable(
             directories that should be scanned for images.
 
     Returns:
-        An iterable of image filenames starting from the given paths. The given paths
+        An iterable of image filepaths starting from the given paths. The given paths
```
guarin:
```suggestion
        An iterable of image filenames starting from the given paths. The given paths
```

tests/_data/_serialize/test_memory_mapped_sequence.py
```diff
@@ -8,105 +8,191 @@
...
 
 
 class TestMemoryMappedSequence:
     def test_index(self, tmp_path: Path) -> None:
-        memory_mapped_sequence.write_filenames_to_file(
-            filenames=["image1.jpg", "image2.jpg", "image3.jpg"],
+        image_dir = tmp_path / "images"
+        mask_dir = tmp_path / "masks"
+        image_dir.mkdir()
+        mask_dir.mkdir()
```
guarin: You don't need to create the directory or use `tmp_path` as the images/masks are just strings and the underlying files are never accessed.

src/lightly_train/_data/file_helpers.py
```diff
@@ -38,10 +38,10 @@ class ImageMode(Enum):
     MASK = "MASK"
 
 
-def list_image_filenames_from_iterable(
+def list_image_filepaths_from_iterable(
```
guarin:
```suggestion
def list_image_filenames_from_iterable(
```

The function returns `Iterable[ImageFilename]` and not paths.

yutong-xiang-97: I've checked the mmap file and it contains the paths instead of names. I think based on the logic, if you provide the paths to the files in a list, you will get the parent directories kept in the names; if it is a directory, then you also get paths with `os.path.join(dir_str, image_filename)` on line 71

src/lightly_train/_data/file_helpers.py
yutong-xiang-97: Explained above.

src/lightly_train/_data/file_helpers.py
guarin: Depends whether the directories themselves are absolute paths or not. If `dir_str` is absolute then the returned string is absolute, otherwise it is relative. In any case the returned type is an `ImageFilename` and not a `Path`.

yutong-xiang-97: OK, if by "filename" you mean "ImageFilename" type instead of a name then we can switch back.

guarin (approved):

PR #288, opened 2025-09-09
Use image size everywhere in predict

guarin:
## What has changed and why?

* Use image size everywhere in predict

Noticed that this wasn't passed everywhere.

- [ ] Yes

- [ ] Yes


stegmuel (commented):
Shouldn't the variable be called `window_size` or `crop_size`.
I find it a bit misleading.

stegmuel (approved):
LGTM.

PR #292, opened 2025-09-12
Object Detection Transforms

liopeer:
## What has changed and why?
- adds transforms: RandomOrder, RandomPhotometricDistort and ZoomOut, which follow the respective torchvision transforms
- it also adds the arguments for the object detection transform, but not the actual transform yet, otherwise it's too much to review

## How has it been tested?
- minimal unit tests (shapes and some things that are deterministic)

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_transforms/random_photometric_distort.py
+++ b/src/lightly_train/_transforms/random_photometric_distort.py
@@ -88,6 +88,8 @@ class RandomPhotometricDistort(ImageOnlyTransform):  # type: ignore[misc]
                     hue=self.hue,
                     p=p,
                 ),
+                # TODO: Lionel (09/25): This might be stronger augmentation than in
+                # torchvision. Verify influence on performance.
                 ChannelShuffle(p=p),
             ]
         )


src/lightly_train/_transforms/random_photometric_distort.py
```diff
@@ -0,0 +1,109 @@
...
+        if any(s < 0 for s in saturation):
+            raise ValueError(
+                f"Saturation values must be non-negative, got {saturation}."
+            )
+
+        if any(-0.5 > h or h > 0.5 for h in hue) or hue[0] > hue[1]:
+            raise ValueError(
+                f"Hue values must respect -0.5 <= min <= max <= 0.5, got {hue}."
+            )
+        if not 0 < p <= 1:
```
liopeer, line 79: Can't allow `p=0.0` here since there will be a zero division in older albumentations versions.

src/lightly_train/_transforms/random_zoom_out.py
```diff
@@ -0,0 +1,103 @@
...
+        if self.side_range[0] < 1.0 or self.side_range[1] < self.side_range[0]:
+            raise ValueError(
+                f"side_range must be a sequence of two floats (min >= 1.0, max >= min), got {side_range}."
+            )
+
+    def get_params_dependent_on_data(
+        self, params: dict[str, Any], data: dict[str, Any]
+    ) -> dict[str, Any]:
+        if ALBUMENTATIONS_GEQ_1_4_21:
+            scale = self.py_random.uniform(self.side_range[0], self.side_range[1])
```
liopeer, line 117: This is the recommended way according to the docs, but was only added recently.

tests/_transforms/test_random_photometric_distort.py
```diff
@@ -0,0 +1,144 @@
...
+        )
+        out = transform(image=img)
+        assert isinstance(out, dict)
+        assert "image" in out
+        result = out["image"]
+        assert isinstance(result, np.ndarray)
+        assert result.shape == img.shape
+        assert result.dtype == img.dtype
+
+    def test__call__always_transform_when_p1(self) -> None:
```
liopeer, line 133: Usually I would test here that it is never applied with `p=0.0` but as mentioned above that's not possible.

src/lightly_train/_transforms/object_detection_transform.py
```diff
@@ -35,12 +41,11 @@ class ObjectDetectionTransformOutput(TaskTransformOutput):
...
-    bbox_params: BboxParams = Field(
-        default_factory=lambda: BboxParams(
-            format="yolo", label_fields=["class_labels"]
-        ),
-    )
+    photometric_distort: RandomPhotometricDistortArgs | None
+    random_zoom_out: RandomZoomOutArgs | None
+    random_flip: RandomFlipArgs | None
+    resize: ResizeArgs | None
+    bbox_params: BboxParams
```
stegmuel, line 48: At which point are the bboxes normalized and which format is used?
RT-DETR has custom code to both re-format to cxywh and normalize the bboxes, shouldn't this be added in this PR as well?

src/lightly_train/_transforms/object_detection_transform.py
```diff
@@ -20,10 +19,17 @@
     TaskTransformInput,
     TaskTransformOutput,
 )
+from lightly_train._transforms.transform import (
+    RandomFlipArgs,
+    RandomPhotometricDistortArgs,
+    RandomZoomOutArgs,
+    ResizeArgs,
+)
+from lightly_train.types import NDArrayImage
```
stegmuel, line 29: Shouldn't there be a class for SanitizeBoundingBoxes as well?

src/lightly_train/_transforms/object_detection_transform.py
```diff
@@ -20,10 +19,17 @@
...
 
 
 class ObjectDetectionTransformInput(TaskTransformInput):
-    image: NDArray[np.uint8]
+    image: NDArrayImage
     bboxes: NotRequired[NDArray[np.float64]]
     class_labels: NotRequired[NDArray[np.int64]]
```
stegmuel, line 35: Shouldn't there be a class for RandomIoUCrop?

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+    resize: DINOv2LTDetrObjectDetectionResizeArgs = Field(
+        default_factory=DINOv2LTDetrObjectDetectionResizeArgs
+    )
+    # We use the YOLO format internally for now.
+    bbox_params: BboxParams = Field(
+        default_factory=lambda: BboxParams(
+            format="yolo", label_fields=["class_labels"], min_width=0.0, min_height=0.0
+        ),
+    )
```
stegmuel, line 68: Shouldn't there be a stop_epoch argument to store the epoch at which some of the transforms are not applied anymore?

src/lightly_train/_transforms/random_order.py
```diff
@@ -0,0 +1,28 @@
...
+
+class RandomOrder(SomeOf):  # type: ignore[misc]
+    def __init__(
+        self,
+        transforms: list[BasicTransform | BaseCompose],
+        n: int = 1,
+        replace: bool = False,
+        p: float = 1.0,
+    ):
+        super().__init__(transforms=transforms, n=n, replace=replace, p=p)
```
stegmuel, line 39: Shouldn't `n=len(transforms)` always to match the behaviour of the torchvision implementation?

stegmuel: Similarly, `replace`should always be False, no?

stegmuel (commented):
Left some comments.

src/lightly_train/_transforms/object_detection_transform.py
liopeer: We currently support the YOLO dataset format as input which means the format of the bboxes is already supposed to be (c_x, c_y, w, h) in normalized coordinates. This is reflected in the default for the `Bboxparams` that we use (inside the `src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py` file).

https://github.com/lightly-ai/lightly-train/blob/760b6d05536df3940f090b304bd4507b562c8ab6/src/lightly_train/_data/yolo_object_detection_dataset.py#L25

src/lightly_train/_transforms/object_detection_transform.py
liopeer: This should not be necessary with albumentations. The filtering is done in every augmentation that affects the bounding boxes and is configured through `BboxParams`, in our case in `src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py`:

```python
bbox_params: BboxParams = Field(
    default_factory=lambda: BboxParams(
        format="yolo", label_fields=["class_labels"], min_width=0.0, min_height=0.0
    ),
)
```

src/lightly_train/_transforms/object_detection_transform.py
liopeer: Yes, this will be added in a follow up PR (the 3rd). There is no direct correspondence in albumentations, so it requires still a bit of figuring out how we can reproduce it. The current PR is already quite big as it is.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
liopeer: All that logic will be added in the next PR.

src/lightly_train/_transforms/random_order.py
liopeer: Good catch, `n=1` as default is a mistake and I will fix it. For `replace` it should be correct, we want sampling without replacement.

src/lightly_train/_transforms/object_detection_transform.py
stegmuel: Nice.

src/lightly_train/_transforms/random_order.py
stegmuel: Yes, I agree with you that it should be False.
My point is why would we want to allow changing that to True or n to be something else than len(transforms)`.

liopeer: Fair point. Instead of changing it, how would you feel if we will just limit `RandomOrderArgs` to the arguments that are also supported in torchvision?

stegmuel: Works for me.

src/lightly_train/_transforms/random_photometric_distort.py
```diff
@@ -0,0 +1,109 @@
...
+        self.transform = RandomOrder(
+            [
+                ColorJitter(
+                    brightness=self.brightness,
+                    contrast=self.contrast,
+                    saturation=self.saturation,
+                    hue=self.hue,
+                    p=p,
+                ),
+                ChannelShuffle(p=p),
```
stegmuel, line 93: Looks like we might end-up with stronger augmentation than the original one as the granularity of the transform is different.
Can you add a TODO, that we might have to tune down p?

stegmuel (commented):
Left some comments.

stegmuel (approved):
LGTM.

src/lightly_train/_transforms/random_photometric_distort.py
liopeer: Added!

PR #297, opened 2025-09-15
Add fallback for onnx export when tiling is not necessary

simonschoelly:
## What has changed and why?

Added a fallback to skip tiling when exporting a model as ONNX if it is not needed.
## How has it been tested?

Only tested with the CPU runtime so far as currently CUDA is broken on our Jetson Orin. So far I can say that we don't experience memory issues anymore for large square images with the CPU runtime.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/export_task.py
+++ b/src/lightly_train/_commands/export_task.py
@@ -113,9 +113,9 @@ def _export_task(
         num_channels:
             Number of channels in input tensor.
         height:
-            Height of the input tensor.
+            Height of the input tensor. For efficiency reasons we recomment this to be the same as width.
         width:
-            Width of the input tensor.
+            Width of the input tensor. For efficiency reasons we recomment this to be the same as height.
         precision:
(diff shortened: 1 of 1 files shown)


src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/task_model.py
```diff
@@ -466,11 +466,18 @@ def _forward_logits(self, x: Tensor) -> Tensor:
         """Forward pass that returns the logits of the last layer. Intended for
         inference."""
         # x is a batch of images with shape (B, C, H, W).
+        _, _, H, W = x.shape
+        # The current implementation of tile and untile leads to large amounts of memory being consumed when
+        # running the model as ONNX. Therefore we add a fallback for the case when these methods are not necessary.
```
liopeer, line 471: Can we add a recommendation to the docstring in `src/lightly_train/_commands/export_task.py` that for EoMT we currently recommend using square input shapes for above reason?

liopeer: Has been addressed.

PR #300, opened 2025-09-16
Add Obj Det Transform Pt 2

liopeer:
## What has changed and why?
- added ScaleJitter transform
- made transforms step/epoch aware so that we can have a stop_step/stop_epoch
- also made some small changes to semantic segmentation transform. we can also leave these away for now and refactor later (the refactoring is anyway not complete with this PR, so might make sense to do everything separately)

What remains:
- RandomIoUCrop

## How has it been tested?
- unit tests (might make sense to add more sophisticated ones before merging)

- [ ] Yes

- [ ] Yes

### DIFF
--- a/README.md
+++ b/README.md
@@ -407,4 +407,5 @@ We're committed to supporting both open-source and commercial users.
 [![Discord](https://img.shields.io/discord/752876370337726585?style=for-the-badge&logo=discord&logoColor=white&label=discord&color=7289da)](https://discord.gg/xvNJW94) <br>
 [![GitHub](https://img.shields.io/badge/GitHub-lightly--ai-black?style=for-the-badge&logo=github&logoColor=white)](https://github.com/lightly-ai/lightly-train) <br>
 [![X](https://img.shields.io/badge/X-lightlyai-black?style=for-the-badge&logo=x&logoColor=white)](https://x.com/lightlyai) <br>
+[![YouTube](https://img.shields.io/badge/YouTube-lightly--tech-blue?style=for-the-badge&logo=YouTube&logoColor=white)](https://www.youtube.com/channel/UCAz60UdQ9Q3jPqqZi-6bmXw) <br>
 [![LinkedIn](https://img.shields.io/badge/LinkedIn-lightly--tech-blue?style=for-the-badge&logo=linkedin&logoColor=white)](https://www.linkedin.com/company/lightly-tech)


src/lightly_train/_transforms/object_detection_transform.py
```diff
@@ -41,17 +50,135 @@ class ObjectDetectionTransformOutput(TaskTransformOutput):
...
+            )
+            self.past_stop = True
+
+        transformed = self.transform(
+            image=input["image"],
+            bboxes=input["bboxes"],
+            class_labels=input["class_labels"],
+        )
+
+        # TODO: Lionel (09/25): Remove in favor of Normalize transform.
```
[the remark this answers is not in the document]
liopeer: This is temporary and reflects how RT-DETR does their normalization.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
```diff
@@ -42,27 +51,76 @@ class DINOv2LTDetrObjectDetectionRandomFlipArgs(RandomFlipArgs):
     vertical_prob: float = 0.0
 
 
-class DINOv2LTDetrObjectDetectionResizeArgs(ResizeArgs):
-    height: int = 644
-    width: int = 644
+class DINOv2LTDetrObjectDetectionStopPolicyArgs(StopPolicyArgs):
+    stop_step: int = 71
```
liopeer:
```suggestion
    # TODO: Lionel (09/25): Set steps instead of epochs.
    stop_step: int = 71
```

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
```diff
@@ -42,27 +51,76 @@ class DINOv2LTDetrObjectDetectionRandomFlipArgs(RandomFlipArgs):
...
+        default_factory=DINOv2LTDetrObjectDetectionScaleJitterArgs
     )
     # We use the YOLO format internally for now.
     bbox_params: BboxParams = Field(
         default_factory=lambda: BboxParams(
             format="yolo", label_fields=["class_labels"], min_width=0.0, min_height=0.0
         ),
     )
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
liopeer: I think this should be moved up to `ObjectDetectionTransformArgs`
```suggestion
```
It's necessary in order to be able to use the `StopPolicyArgs`: This has classes and not objects in its attributes, which can't be serialized.

src/lightly_train/_transforms/semantic_segmentation_transform.py
```diff
@@ -185,7 +187,12 @@ def __init__(self, transform_args: SemanticSegmentationTransformArgs) -> None:
 
         # Optionally apply random horizontal flip.
         if transform_args.random_flip is not None:
-            transform += [HorizontalFlip(p=transform_args.random_flip.horizontal_prob)]
+            if transform_args.random_flip.horizontal_prob > 0.0:
```
liopeer, line 198: Unrelated, but I saw that the vertical flip had been completely ignored until now.

src/lightly_train/_transforms/semantic_segmentation_transform.py
```diff
@@ -67,6 +68,7 @@ class SemanticSegmentationTransformArgs(TaskTransformArgs):
     normalize: NormalizeArgs
     random_flip: RandomFlipArgs | None
     color_jitter: ColorJitterArgs | None
+    # TODO: Lionel(09/25): These are currently not fully used.
     scale_jitter: ScaleJitterArgs | None
```
liopeer, line 72:
```suggestion
    scale_jitter: ScaleJitterArgs | None # TODO: Lionel(09/25): These are currently not fully used.
```

src/lightly_train/_transforms/transform.py
```diff
@@ -57,11 +59,11 @@ class RandomFlipArgs(PydanticConfig):
 
 
 class RandomPhotometricDistortArgs(PydanticConfig):
-    brightness: tuple[float, float] = Field(strict=False, ge=0.0)
-    contrast: tuple[float, float] = Field(strict=False, ge=0.0)
-    saturation: tuple[float, float] = Field(strict=False, ge=0.0)
-    hue: tuple[float, float] = Field(strict=False, ge=-0.5, le=0.5)
-    prob: float = Field(ge=0.0, le=1.0)
```
liopeer, line 64: These don't actually work for tuples and we have good checks in the transforms themselves, so it's not really necessary and makes the code complicated.

guarin: Huh, I am pretty sure `Field(strict=False)` is required to allow anything else than a tuple as input.

liopeer: Oh damn, you're right. Should have kept the `strict=False`. What doesn't work are only the `ge=0.0`: Those only work for single numbers, not for tuples. Tuple would require custom resolvers. I'll address it.

src/lightly_train/_commands/train_task.py
```diff
@@ -393,6 +401,9 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
             optimizer.zero_grad()
             scheduler.step()
 
+            train_dataloader.dataset.transform.global_step = state["step"]
```
stegmuel: Can you leave a comment why you do this. Otherwise it might be confusing, especially since the step is also passed to the validation transform.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
stegmuel: Why not call that `stop_epoch` then? 
Recent experiments showed better results when this was set to 72, i.e. disabled can you leave a comment or set the default such that this by default disabled?

src/lightly_train/_transforms/scale_jitter.py
```diff
@@ -0,0 +1,87 @@
...
+        self.target_size = target_size
+        self.scale_range = scale_range
+        self.divisible_by = divisible_by
+        self.p = p
+        self.seed_offset = seed_offset
+        self.step_seeding = step_seeding
+
+        self._step = 0
+
+        factors = np.linspace(
```
stegmuel: In RT-DETR the distribution over the resolution is not uniform. (see [this](https://github.com/lyuwenyu/RT-DETR/blob/f9417e3acfa48bcb649e5ec0bc3de1e8677c8961/rtdetrv2_pytorch/configs/rtdetrv2/include/dataloader.yml#L22)). 640 or 644 is sampled more often.

src/lightly_train/_commands/train_task.py
stegmuel: I'm afraid this won't work. 
My understanding is that each worker will have its own copy of the transform created by the main process and that the update will only affect the version from the main process.

stegmuel (commented):
First review.

src/lightly_train/_transforms/object_detection_transform.py
```diff
@@ -41,17 +52,148 @@ class ObjectDetectionTransformOutput(TaskTransformOutput):
...
+        )
+
+    def __call__(
         self, input: ObjectDetectionTransformInput
     ) -> ObjectDetectionTransformOutput:
-        pass
+        # Adjust transform after stop_step is reached.
+        if (
+            self.stop_step is not None
+            and self.global_step >= self.stop_step
```
stegmuel, line 171: Reminder that the default is `stop_step=71`, which is an epoch value.

src/lightly_train/_commands/train_task.py
liopeer: This is now actually a `WorkerSharedStep` instance. And it's not used anymore for the `ScaleJitter`. In there, it would set the step for all the workers, but we would need to update only ever one worker at a time. I now therefore moved the `ScaleJitter` to the batch collation like in the original repo. There would be another way, but this would additionally require a custom `BatchSampler`, which I omitted for now.

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -431,43 +427,27 @@ def get_dataset(
     )
 
 
-# TODO(Guarin, 08/25): Move this function to the _data module.
```
liopeer, line 434: moved

src/lightly_train/_data/dataloader_helpers.py
```diff
@@ -0,0 +1,27 @@
...
+# Copyright (c) Lightly AG and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+#
+from multiprocessing import Value
+
+
+class WorkerSharedStep:
```
liopeer: If you have a better idea where to put this I'm all ears.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -55,16 +59,20 @@ class MultiChannelClassInfo(PydanticConfig):
 ClassInfo = Union[MultiChannelClassInfo, SingleChannelClassInfo]
 
 
-class MaskSemanticSegmentationDataset(Dataset[MaskSemanticSegmentationDatasetItem]):
+class MaskSemanticSegmentationDataset(TaskDataset):
+    batch_collate_fn_cls: ClassVar[type[BaseCollateFunction]] = (
```
liopeer, line 63: Wasn't sure if it makes more sense to put it in the Dataset class or the Transform class. Wdyt?

src/lightly_train/_data/task_dataset.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class TaskDataset(Dataset[TaskDatasetItem]):
+    batch_collate_fn_cls: ClassVar[type[BaseCollateFunction]] = BaseCollateFunction
+
+    def __init__(self, transform: TaskTransform) -> None:
+        self._transform = transform
+
+    @property
+    def transform(self) -> TaskTransform:
```
liopeer, line 24: Not sure if we should instead have the `global_step` directly here as well.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
liopeer: I now disabled it by default (in the full TransformArgs I set StopPolicyArgs to `None`. But since we use steps in all the task_training, I would also specify it as a `stop_step`.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
liopeer: Done.

src/lightly_train/_commands/train_task.py
liopeer: Completely removed.

src/lightly_train/_data/dataloader_helpers.py
liopeer: Removed.

src/lightly_train/_transforms/transform.py
liopeer: Addressed.

src/lightly_train/_transforms/scale_jitter.py
liopeer: Addressed.

src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/transforms.py
```diff
@@ -40,10 +40,14 @@ class DINOv2EoMTSemanticSegmentationColorJitterArgs(ColorJitterArgs):
...
 class DINOv2EoMTSemanticSegmentationScaleJitterArgs(ScaleJitterArgs):
-    min_scale: float = 0.5
-    max_scale: float = 2.0
-    num_scales: int = 20
+    sizes: Sequence[tuple[int, int]] | None = None
+    min_scale: float | None = 0.5
+    max_scale: float | None = 2.0
+    num_scales: int | None = 20
     prob: float = 1.0
+    step_seeding: bool = False
```
stegmuel: Is this still needed?

src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/transforms.py
```diff
@@ -40,10 +40,15 @@ class DINOv2LinearSemanticSegmentationColorJitterArgs(ColorJitterArgs):
...
-    min_scale: float = 0.5
-    max_scale: float = 2.0
-    num_scales: int = 20
+    sizes: Sequence[tuple[int, int]] | None = None
+    min_scale: float | None = 0.5
+    max_scale: float | None = 2.0
+    num_scales: int | None = 20
     prob: float = 1.0
+    # TODO: Lionel(09/25): These are currently not used.
+    step_seeding: bool = False
```
stegmuel: Is this still needed?

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
```diff
@@ -42,27 +51,90 @@ class DINOv2LTDetrObjectDetectionRandomFlipArgs(RandomFlipArgs):
     vertical_prob: float = 0.0
 
 
-class DINOv2LTDetrObjectDetectionResizeArgs(ResizeArgs):
-    height: int = 644
-    width: int = 644
+class DINOv2LTDetrObjectDetectionStopPolicyArgs(StopPolicyArgs):
+    # In the original implementation, stop_epoch is 71. We assume a dataset of 100_000
+    # and batch size 16.
+    stop_step: int = 71 * 100_000 // 16
```
stegmuel: Should we keep this?
Let say we keep it, shouldn't we set it by default to None?

src/lightly_train/_task_models/dinov3_eomt_semantic_segmentation/transforms.py
```diff
@@ -40,10 +40,14 @@ class DINOv3EoMTSemanticSegmentationColorJitterArgs(ColorJitterArgs):
...
 class DINOv3EoMTSemanticSegmentationScaleJitterArgs(ScaleJitterArgs):
-    min_scale: float = 0.5
-    max_scale: float = 2.0
-    num_scales: int = 20
+    sizes: Sequence[tuple[int, int]] | None = None
+    min_scale: float | None = 0.5
+    max_scale: float | None = 2.0
+    num_scales: int | None = 20
     prob: float = 1.0
+    step_seeding: bool = False
```
stegmuel: Still needed?

src/lightly_train/_transforms/transform.py
```diff
@@ -11,13 +11,15 @@
 from collections.abc import Sequence
```
stegmuel, line 11: shouldn't the file be name transforms.py with an "s" ?

src/lightly_train/_transforms/transform.py
```diff
@@ -147,10 +149,21 @@ def from_dict(cls, config: dict[str, list[float]]) -> NormalizeArgs:
...
-    min_scale: float
-    max_scale: float
-    num_scales: int
-    prob: float
+    sizes: Sequence[tuple[int, int]] | None
+    min_scale: float | None
+    max_scale: float | None
+    num_scales: int | None
+    prob: float = Field(ge=0.0, le=1.0)
+    divisible_by: int | None
```
stegmuel, line 157: Shouldn't the list of target resolution sizes be provided here?

stegmuel (commented):
Left some questions/remarks?

src/lightly_train/_transforms/object_detection_transform.py
liopeer: The default is now a step (very heuristic though).

src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/transforms.py
liopeer: I will remove the `step_seeding` and `seed_offset` (also from the parent class). The other things are useful for the EoMT transforms which I will eventually migrate to the `ScaleJitter` transform as well.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
liopeer: I actually set the entire `stop_policy` as `None` further down, i.e.
```python
stop_policy: DINOv2LTDetrObjectDetectionStopPolicyArgs | None = None
```
But that means I could remove the entire class `DINOv2LTDetrObjectDetectionStopPolicyArgs`, which I'll do to avoid confusion.

src/lightly_train/_transforms/transform.py
liopeer: The filename is not very descriptive anymore, I agree. I would probably take the entire `MethodTransform` related things out and rename it to `transform_args.py` and `method_transform.py`. Wdyt? Would happen in a separate PR though since this will refactor a lot of paths in many files.

src/lightly_train/_transforms/transform.py
liopeer: This is just the "abstract" class. The list with the actual resolution sizes is in `src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py`

```python
class DINOv2LTDetrObjectDetectionScaleJitterArgs(ScaleJitterArgs):
    sizes: Sequence[tuple[int, int]] | None = [
        (490, 490),
        (518, 518),
        (546, 546),
        (588, 588),
        (616, 616),
        (644, 644),
        (644, 644),
        (644, 644),
        (686, 686),
        (714, 714),
        (742, 742),
        (770, 770),
        (812, 812),
    ]
```

src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/transforms.py
liopeer: Same answer as above.

src/lightly_train/_task_models/dinov3_eomt_semantic_segmentation/transforms.py
liopeer: Same answer as above.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
liopeer: Done.

src/lightly_train/_transforms/transform.py
stegmuel: Yes, I saw that `src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py` has the correct values for DINOv2, but we need to either have a new equivalent file for DINOv3 or set the correct values in the abstract one, no?

liopeer: Ah yes. That's the idea. For now we wanted to keep things separate, i.e. one model = one directory. We will have to move things further up the chain eventually, but I would not do that right now.

PR #301, opened 2025-09-17
Resume scheduler properly

LinkinPony:
## What has changed and why?

When training DINOv3_EoMT (or anything else with `TwoStageWarmupPolySchedule`), the learning rate will reset to the initial stage when `resume_interrupted=True`. This is because the scheduler status is not saved and resumed properly, so `last_epoch` is always equal to -1 when resumed from checkpoints.

lr_1
lr_2

What changed:
- Added scheduler: LRScheduler to TrainTaskState so the LR scheduler state is saved and restored with checkpoints.
- Included the scheduler instance in the saved state in train_task.py.
- Ensured the same scheduler object is preserved on load by asserting identity in load_checkpoint.

## How has it been tested?

Rerun the same train and resume from checkpoints to make sure the learning rate is consistent.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/train_task.py
+++ b/src/lightly_train/_commands/train_task.py
@@ -342,6 +342,7 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
         state = TrainTaskState(
             train_model=train_model,
             optimizer=optimizer,
+            scheduler=scheduler,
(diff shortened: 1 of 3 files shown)


guarin (approved):
Thanks for the contribution! This LGTM

PR #303, opened 2025-09-17
Improve existing mmap file handling

guarin:
## What has changed and why?

* Relax checks if run is resumed or overwritten

This should "fix" most occurrences where we raise an error because a mmap file already exists and we don't know whether it is the correct one. It makes the assumption that `resume_interrupted` and `overwrite` are used without changing the underlying data. For `resume_interrupted` this makes always sense because resume doesn't work properly if you change the data anyways. For `overwrite` this makes sense if we assume that overwrite is mostly used during debugging. Proper runs should always use a new directory.

## How has it been tested?

* Unit tests

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/common_helpers.py
+++ b/src/lightly_train/_commands/common_helpers.py
@@ -391,6 +391,8 @@ def _get_package(model: Module) -> BasePackage:
 def get_dataset_temp_mmap_path(
     data: PathLike | Sequence[PathLike],
     out: PathLike,
+    resume_interrupted: bool,
+    overwrite: bool,
 ) -> Generator[Path, Any, Any]:
     """Generate file in the cache directory to be used for memory-mapping the dataset.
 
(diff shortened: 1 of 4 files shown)


src/lightly_train/_commands/common_helpers.py
```diff
@@ -414,6 +416,8 @@ def get_dataset_temp_mmap_path(
         # Use data as identifier to share the mmap file across multiple runs.
         # NOTE(Guarin, 09/25): Hash of data might be slow if data is a long list of
         # filenames or directories.
+        if isinstance(data, (str, Path)):
+            data = Path(data).resolve()
```
guarin, line 420: To have same behavior as for `out` 3 lines below

yutong-xiang-97: Nit: could simply use `PathLike` here. Doesn't have to be fixed

tests/_commands/test_common_helpers.py
```diff
@@ -851,6 +866,8 @@ def test_get_dataset_mmap_file__reuse(
             out_dir=tmp_path,
             filenames=filenames,
             mmap_filepath=mmap_filepath,
+            resume_interrupted=False,
```
yutong-xiang-97, line 932: Here you could add a test for the reuse logic in different settings

src/lightly_train/_commands/common_helpers.py
guarin: Doesn't work, you can't use type aliases in `isinstance` checks. They can only be used to type code but not in code that is executed at runtime.

tests/_commands/test_common_helpers.py
guarin: Added tests 👍🏼

yutong-xiang-97 (approved):
Approved!

PR #307, opened 2025-09-22
Infer image shape from model for onnx export

simonschoelly:
## What has changed and why?

This PR infers the height and width from the model for the ONNX export if they re not specified.

I also infer the number of channels now from the model. I don't think it makes sense at all that one can specify the number of channels for the ONNX export - so I decided to remove that option at all.

I don't think we even have the ability to train with something else than RGB channels - so I did not write a test for grayscale images for example - let me know if that is actually possible.

## How has it been tested?

- [ ] Yes

Probably not necessary

- [x] Yes


tests/_commands/test_export_task.py
```diff
@@ -89,8 +91,8 @@ def dinov2_vits14_eomt_checkpoint(tmp_path_factory: pytest.TempPathFactory) -> P
 @pytest.mark.skipif(not RequirementCache("onnxslim"), reason="onnxslim not installed")
 def test_onnx_export(
     batch_size: int,
-    height: int,
-    width: int,
+    height: Union[int, None],
```
yutong-xiang-97: Here we use Py3.10 convention `int | None` and add `from __future__ import annotations` to make it work for lower versions.

src/lightly_train/_commands/export_task.py
```diff
@@ -181,16 +179,26 @@ def _export_task_from_config(config: ExportTaskConfig) -> None:
     )
     task_model.eval()
 
+    height = config.height
+    width = config.width
+    # TODO we might also use task_model.backbone.in_chans
+    num_channels = len(task_model.image_normalize["mean"])
```
yutong-xiang-97, line 183: It is nice that you infer the channels from the model!

tests/_commands/test_export_task.py
```diff
@@ -102,6 +104,10 @@ def test_onnx_export(
     model = lightly_train.load_model_from_checkpoint(
         dinov2_vits14_eomt_checkpoint, device="cpu"
     )
+    if height is None:
+        height = model.image_size[0]
+    if width is None:
+        width = model.image_size[1]
     onnx_path = tmp_path / "model.onnx"
     validation_input = torch.randn(batch_size, 3, height, width, device="cpu")
```
yutong-xiang-97: Indeed we are able to train with multi-channel images (#279 ). You could add another test for 4-channel images.

src/lightly_train/_commands/export_task.py
```diff
@@ -72,9 +72,8 @@ def export_onnx(
     out: PathLike,
     checkpoint: PathLike,
     batch_size: int = 1,
-    num_channels: int = 3,
-    height: int = 224,
-    width: int = 224,
+    height: Union[int, None] = None,
```
yutong-xiang-97:
```suggestion
    height: int | None = None,
```

src/lightly_train/_commands/export_task.py
```diff
@@ -72,9 +72,8 @@ def export_onnx(
...
+    width: Union[int | None] = None,
```
yutong-xiang-97:
```suggestion
    width: int | None = None,
```

tests/_commands/test_export_task.py
```diff
@@ -140,6 +162,72 @@ def test_onnx_export(
...
+@pytest.mark.skipif(
+    sys.version_info < (3, 9),
+    reason="Requires Python 3.9 or higher for image preprocessing.",
+)
+@pytest.mark.skipif(not RequirementCache("onnx"), reason="onnx not installed")
+@pytest.mark.skipif(
+    not RequirementCache("onnxruntime"), reason="onnxruntime not installed"
+)
+@pytest.mark.skipif(not RequirementCache("onnxslim"), reason="onnxslim not installed")
+def test_onnx_export_4_channels(
```
yutong-xiang-97: You don't have to add an additional test or create fixtures for this. Simply parameterize the `test_onnx_export` test with `num_channels` like done here: https://github.com/lightly-ai/lightly-train/blob/f32cd12cd03c9a12d630ef1dd7c8cec4235bf10b/tests/_data/test_mask_semantic_segmentation_dataset.py#L423

simonschoelly: I think I need this because I need to add the number of channels when creating the models in the fixture, namely in this line
```python
transform_args={"num_channels": num_channels}
```

yutong-xiang-97: Aren't you able to do `checkpoint= create_dinov2_vits14_eomt_test_checkpoint(tmp_path, num_channels=num_channels)` in `.export_onnx()` and use parameterization for the tests? Also the `tmp_path` is already provided as a fixture in pytest so you don't have to use `tmp_path_factory.mktemp("tmp")`.

PR #308, opened 2025-09-22
Support loading checkpoint for semantic segmentation

yutong-xiang-97:
## What has changed and why?

Add support for training sem seg models from an existing model checkpoint. Like in `train`, we added a `checkpoint` parameter where users could pass the ckpt file path. Also it cannot be specified simutaneously with `resume_interrupted`

## How has it been tested?

- Unit tests
- Manually on a compute machine and compare the metrics from training from scratch.

- [x] Yes

- [ ] Yes
- [x] Not needed (will have a follow-up issue later on)

### DIFF
--- a/tests/_commands/test_train_task.py
+++ b/tests/_commands/test_train_task.py
@@ -186,3 +186,83 @@ def test_train_semantic_segmentation__checkpoint(
             checkpoint=last_ckpt_path,
         )
     assert f"Loading checkpoint from '{last_ckpt_path}'" in caplog.text
+
+
+@pytest.mark.skipif(
+    sys.platform.startswith("win") or is_self_hosted_docker_runner,
+    reason=(
+        "Fails on Windows since switching to Jaccard index "
+        "OR on self-hosted CI with GPU (insufficient shared memory causes worker bus error)"
+    ),
+)
+def test_train_semantic_segmentation__resume_interrupted(
+    tmp_path: Path, caplog: LogCaptureFixture
+) -> None:
+    """Assert that resume_interrupted loads the last checkpoint from the output dir."""
+    out = tmp_path / "out"
+    train_images = tmp_path / "train_images"
+    train_masks = tmp_path / "train_masks"
+    val_images = tmp_path / "val_images"
+    val_masks = tmp_path / "val_masks"
+    helpers.create_images(train_images)
+    helpers.create_masks(train_masks)
+    helpers.create_images(val_images)
+    helpers.create_masks(val_masks)
+
+    # Part 1: Generate a checkpoint that can be resumed.
+    lightly_train.train_semantic_segmentation(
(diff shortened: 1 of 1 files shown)


src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -628,3 +630,30 @@ def load_checkpoint(fabric: Fabric, out_dir: PathLike, state: TrainTaskState) ->
...
+
+    # Sanity check to make sure that checkpoint loading didn't create new objects or
+    # changed the model state.
+    assert state["train_model"] is train_model
+    assert {
+        n: p.requires_grad for n, p in state["train_model"].named_parameters()
+    } == train_model_grads
+    assert {
+        n: m.training for n, m in state["train_model"].named_modules()
+    } == train_model_trainings
```
simonschoelly, line 659: I was a bit wondering if that sanity check makes sense if we add new parameters/modules in the future to our model but load an old checkpoint - but then I thought that probably makes sense as in that case we would need to write some custom logic anyways to populate the new parameters/modules.

tests/_commands/test_train_task.py
```diff
@@ -104,3 +106,83 @@ def test_train_semantic_segmentation(
...
+            "val": {
+                "images": val_images,
+                "masks": val_masks,
+            },
+            "classes": {
+                0: "background",
+                1: "car",
+            },
+        },
+        model="dinov2/vits14-eomt",
```
simonschoelly, line 149: We have a lot of tests for dinov2 already - perhaps we can use dinov3 for this test?

tests/_commands/test_train_task.py
```diff
@@ -104,3 +106,83 @@ def test_train_semantic_segmentation(
...
+        batch_size=2,
+        num_workers=0,
+        steps=1,
+    )
+    last_ckpt_path = out / "checkpoints" / "last.ckpt"
+    assert last_ckpt_path.exists()
+
+    # Part 2: Load the checkpoint via the checkpoint parameter and assert log.
+    with caplog.at_level(logging.INFO):
+        lightly_train.train_semantic_segmentation(
```
simonschoelly, line 163: Perhaps it would be good to also have a test for `resume_interrupted`?

src/lightly_train/_commands/train_task_helpers.py
yutong-xiang-97: Yes indeed.

tests/_commands/test_train_task.py
yutong-xiang-97: We have to use a download link for DINOv3 from Meta which expires constantly, so not suitable for tests.

simonschoelly: Ok then we some how have to tackle this in the future.

PR #312, opened 2025-09-23
Export EoMT model to `.pt` at the end of training

yutong-xiang-97:
## What has changed and why?

Allow exporting the semantic segmentation model weights to `exported_models/exported_last.pt`, similar to what has been done in `.train()`.

The file includes

- `train_model`: state dict of the model
- `model_class_path`: the module path to find the corresponding class for the model
- `model_init_args`: arguments to initialize the model

which are the minimum args needed to reconstruct the model.

This way, instead of using `.ckpt` file which contains also other info to resume training, we could use the exported model for inference and onnx export. It could also be used as the `checkpoint` parameter added in #308 to continue fine-tuning with the model weights. Meanwhile, using a `.ckpt` file still works.

Changes have also been made to the corresponding tests and docs which used the ckpts before but better use the new `exported_last.pt`.

## How has it been tested?

- Unit tests
- Manually on a compute machine

[Semantic Segmentation - LightlyTrain documentation.pdf](https://github.com/user-attachments/files/22497333/Semantic.Segmentation.-.LightlyTrain.documentation.pdf)

- [x] Yes

- [x] Yes

### DIFF
--- a/docs/source/semantic_segmentation.md
+++ b/docs/source/semantic_segmentation.md
@@ -72,7 +72,7 @@ if __name__ == "__main__":
 
 ### Load the Trained Model from Checkpoint and Predict
 
-After the training completes you can load the model for inference with the exported model weights:
+After the training completes you can load the model for inference like this:
 
 ```python
 import lightly_train
--- a/src/lightly_train/_commands/export_task.py
+++ b/src/lightly_train/_commands/export_task.py
@@ -177,7 +177,7 @@ def _export_task_from_config(config: ExportTaskConfig) -> None:
     ).as_posix()  # TODO(Yutong, 07/25): make sure the format corrsponds to the output file extension!
     checkpoint_path = common_helpers.get_checkpoint_path(checkpoint=config.checkpoint)
     task_model = task_model_helpers.load_model_from_checkpoint(
-        exported_model_or_checkpoint=checkpoint_path
+        checkpoint=checkpoint_path
     )
     task_model.eval()
 
@@ -247,7 +247,7 @@ def _export_task_from_config(config: ExportTaskConfig) -> None:
             outputs_onnx = tuple(torch.from_numpy(y) for y in outputs_onnx)
 
             task_model = task_model_helpers.load_model_from_checkpoint(
-                exported_model_or_checkpoint=checkpoint_path, device="cpu"
+                checkpoint=checkpoint_path, device="cpu"
             )
             outputs_model = task_model(x_model)
 
--- a/src/lightly_train/_commands/train_task.py
+++ b/src/lightly_train/_commands/train_task.py
@@ -440,7 +440,6 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
             if is_save_ckpt_step or is_last_step:
                 helpers.save_checkpoint(fabric=fabric, out_dir=out_dir, state=state)
 
-            if is_last_step:
                 model_dict = {
                     "model_class_path": state["model_class_path"],
                     "model_init_args": state["model_init_args"],
--- a/src/lightly_train/_task_models/task_model_helpers.py
+++ b/src/lightly_train/_task_models/task_model_helpers.py
@@ -18,13 +18,13 @@ from lightly_train.types import PathLike
 
 
 def load_model_from_checkpoint(
-    exported_model_or_checkpoint: PathLike,
+    checkpoint: PathLike,
     device: Literal["cpu", "cuda", "mps"] | torch.device | None = None,
 ) -> TaskModel:
     """Load a model from an exported model file (in .pt format) or a checkpoint file (in .ckpt format).
 
     Args:
-        exported_model_or_checkpoint:
+        checkpoint:
             Path to the exported model file or checkpoint file.
         device:
             Device to load the model on. If None, the model will be loaded onto a GPU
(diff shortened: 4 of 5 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add support for exporting the semantic segmentation model weights to \`\`exported_models/exported_last.pt\` under the output directory at the end of training.
```
JonasWurst: will this render properly?

tests/_commands/test_export_task.py
```diff
@@ -59,7 +59,7 @@ def dinov2_vits14_eomt_checkpoint(tmp_path_factory: pytest.TempPathFactory) -> P
         steps=1,
     )
 
-    checkpoint_path = out / "checkpoints/last.ckpt"
+    checkpoint_path = out / "exported_models" / "exported_last.pt"
```
JonasWurst, line 62: Both need to exist no?

tests/_commands/test_train_task.py
```diff
@@ -98,7 +98,7 @@ def test_train_semantic_segmentation(
     assert (out / "train.log").exists()
 
     model = lightly_train.load_model_from_checkpoint(
-        checkpoint=out / "checkpoints" / "last.ckpt"
+        exported_model_or_checkpoint=out / "exported_models" / "exported_last.pt"
```
JonasWurst: Both are required or?

CHANGELOG.md
```suggestion
- Add support for exporting the semantic segmentation model weights to `exported_models/exported_last.pt` under the output directory at the end of training.
```

docs/source/semantic_segmentation.md
```diff
@@ -72,13 +72,13 @@ if __name__ == "__main__":
 
 ### Load the Trained Model from Checkpoint and Predict
 
-After the training completes you can load the model for inference like this:
+After the training completes you can load the model for inference with the exported model weights:
```
guarin: Would keep the previous version

src/lightly_train/_commands/export_task.py
```diff
@@ -247,7 +247,7 @@ def _export_task_from_config(config: ExportTaskConfig) -> None:
             outputs_onnx = tuple(torch.from_numpy(y) for y in outputs_onnx)
 
             task_model = task_model_helpers.load_model_from_checkpoint(
-                checkpoint=checkpoint_path, device="cpu"
+                exported_model_or_checkpoint=checkpoint_path, device="cpu"
```
guarin: Please don't rename input argument name, otherwise it is a breaking change

src/lightly_train/_commands/train_task.py
```diff
@@ -440,6 +440,15 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
             if is_save_ckpt_step or is_last_step:
                 helpers.save_checkpoint(fabric=fabric, out_dir=out_dir, state=state)
 
+            if is_last_step:
```
guarin: Would run this on every `if is_save_ckpt_step or is_last_step:` same as above

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -601,6 +601,20 @@ def save_checkpoint(fabric: Fabric, out_dir: Path, state: TrainTaskState) -> Non
...
+    out_dir = Path(out_dir).resolve()
+    model_path = out_dir / "exported_models" / "exported_last.pt"
+    return model_path
+
+
+def export_model(out_dir: Path, model_dict: dict[str, Any]) -> None:
+    model_path = get_exported_model_path(out_dir)
+    model_path.parent.mkdir(parents=True, exist_ok=True)
+
+    logger.info(f"Exporting models to '{model_path}'")
```
guarin:
```suggestion
    logger.info(f"Exporting model to '{model_path}'")
```

src/lightly_train/_task_models/task_model_helpers.py
```diff
@@ -18,14 +18,14 @@
 
 
 def load_model_from_checkpoint(
-    checkpoint: PathLike,
+    exported_model_or_checkpoint: PathLike,
```
guarin:
```suggestion
    checkpoint: PathLike,
```

tests/_commands/test_export_task.py
guarin: Yes

tests/_commands/test_train_task.py
```diff
@@ -108,6 +108,80 @@ def test_train_semantic_segmentation(
     assert prediction.max() <= 1
 
 
+pytest.mark.skipif(
```
guarin: Otherwise it is not applied to the function. Also move down, there should be no emtpy lines between decorator and the function

```suggestion
@pytest.mark.skipif(
```

src/lightly_train/_task_models/task_model_helpers.py
guarin: I think we can add a more general `def load_model(model: str, device)` function in the future that either loads from checkpoint, exported model, url, or model name. You can even add it now already and make it load only from checkpoint and exported model for now.

yutong-xiang-97: I would make a separate issue for this and push it later.

tests/_commands/test_train_task.py
yutong-xiang-97: I am not 100% sure if I get you, but we would prefer to use the exported model for the testing in this case.

PR #319, opened 2025-09-26
Mention license info for DINOv3 weights in README

yutong-xiang-97:
## What has changed and why?

As stated by title

## How has it been tested?

N/A

- [ ] Yes

- [ ] Yes

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -6,19 +6,24 @@ This project incorporates the following third-party components:
 
 - Source: https://github.com/facebookresearch/dinov2
 - License: Apache License 2.0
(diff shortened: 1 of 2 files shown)


liopeer (changes requested):
LGTM, but I realized that we are actually missing the DINOv3 license in the `NOTICE` file. Could you add that? Something like this probably:

```markdown
**DINOv3**
- Source: https://github.com/facebookresearch/dinov3
- License: DINOv3 License (custom)
- License file: licenses/DINOv3_LICENSE.md
```

Optimally you would quickly check to which files we made changes – I definitely know that we changed all relative imports to absolute imports.

PR #320, opened 2025-09-28
Add RandomIoUCrop

liopeer:
## What has changed and why?
- adds RandomIoUCrop transforms, implemented in albumentations

## How has it been tested?
- unit tests with minimal/maximal dependencies
- manually tested with all albumentation versions in between

- [ ] Yes

- [ ] Yes

### DIFF
--- a/docs/source/semantic_segmentation.md
+++ b/docs/source/semantic_segmentation.md
@@ -253,7 +253,7 @@ The following mask formats are supported:
 
 - png
 
-### Supported Mask Dataset Format
+### Specify Mask Filepaths
 
 We support two ways of specifying the mask filepaths in relation to the image filepaths:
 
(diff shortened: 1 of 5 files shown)


src/lightly_train/_transforms/random_iou_crop.py
```diff
@@ -0,0 +1,165 @@
...
+                if left == right or top == bottom:
+                    continue
+
+                # Convert bboxes from [0, 1] to absolute image coordinates.
+                bboxes_absolute = orig_bboxes * np.array(
+                    [orig_w, orig_h, orig_w, orig_h]
+                )
+
+                # Get bboxes whose center is in the crop.
+                cx = (bboxes_absolute[:, 0] + bboxes_absolute[:, 2]) / 2
```
stegmuel: Confused here. I thought we used the YOLO format, which is [cx, cy, w, h], no?

liopeer: Fair enough and good question, that's really confusing and maybe we need to eventually change our internal format. Following how it currently is:

1. For our dataset we use [cx, cy, w, h] with normalized coordinates (i.e. w, h in [0,1]). This is also what the output of our models will provide to the user.
2. Albumentations allows inputting different formats, which you specify via the `BboxParams` argument to `Compose`. In the unit test I used `pascal_voc` which means [x_min, y_min, x_max, y_max] in *unnormalized* (i.e. full pixel) coordinates.
3. Albumentations inside however will not use that format: When you write your own transforms you need to use the Albumentation internal format which is normalized [x_min, y_min, x_max, y_max].

Hope that answers it. I know it's not super beautiful and I will add some comments.

stegmuel: - I'm not sure, to follow where the conversion occurs?
- In the RandomIoUCrop you explicitly convert from normalized to unnormalized values.

liopeer: To conclude:

- conversion happens at the beginning and end of a `Compose` pipeline
- the explicit conversion inside the class is due to the fact that `crop_coordinates` are unnormalized, i.e. pixel values, whereas bbox coordinates are in a normalized format

PR #324, opened 2025-09-29
Fix model weights loading when continuing training from a model weight file

yutong-xiang-97:
## What has changed and why?

Fix model weights loading when continuing training from a model weight file so that `class_head` weights for EoMT are reinitialized (while `mask_head` and `upscale` weights are loaded from the file). 

Also add a flag `reuse_class_head` to control this behavior. It is set to `false` by default for fine-tuning from an existing model weights for a downstream task with different num_classes. The `class_head` may also be reused for cases like the fine-tuning phase in auto-labeling.

## How has it been tested?

Add a sanity check in the function and it passes the tests calling the function.

- [ ] Yes

- [ ] Yes


src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -650,6 +650,7 @@ def load_checkpoint_from_file(
     fabric: Fabric,
     ckpt_path: PathLike,
     state: TrainTaskState,
+    reuse_class_head: bool = False,
```
liopeer:
```suggestion
    reuse_class_head: bool,
```
For these internal functions I would not set defaults unless necessary.

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -659,15 +660,42 @@ def load_checkpoint_from_file(
...
+    class_head_param_ids_before: set[int] | None = None
+    if not reuse_class_head and hasattr(train_model, "class_head"):
+        class_head_before = getattr(train_model, "class_head")
+        class_head_param_ids_before = {
+            id(param) for param in class_head_before.parameters()
+        }
+        delattr(train_model, "class_head")
+
     logger.info(f"Loading checkpoint from '{ckpt_path}'")
     fabric.load(path=ckpt_path, state={"train_model": train_model})  # type: ignore[arg-type]
```
liopeer: Does `fabric.load` use a `strict=False` setting? Because some parameters will now be missing from the checkpoint.

tests/_commands/test_train_task.py
```diff
@@ -259,6 +259,37 @@ def test_train_semantic_segmentation__checkpoint(
...
+            },
+            model="dinov2/vits14-eomt",
+            accelerator="auto" if not sys.platform.startswith("darwin") else "cpu",
+            devices=1,
+            batch_size=2,
+            num_workers=0,
+            steps=1,
+            overwrite=True,
+            checkpoint=last_ckpt_path,
+            reuse_class_head=False,
```
liopeer, line 289: I think this unit test is wrong, right? Because in above one you don't load the class head (since the default is `False`). So here you should check with `reuse_class_head=True` and then check that the debug level logs don't include `""Skipping class-dependent parameters from checkpoint: %s""`.

liopeer (changes requested):
Loading etc. looks good! I am confused about the unit test though.

tests/_commands/test_train_task.py
yutong-xiang-97: I tuned `num_classes` here so that the class head should be reinited. Indeed though, I should check the param skipping message in the logs.

liopeer: Oh okay, did not see that the number of classes changed. Thanks!

PR #326, opened 2025-09-29
Support saving best model and metric with EoMT

yutong-xiang-97:
## What has changed and why?

Allow saving the checkpoints and model weights with the best `val_miou` for EoMT training. 

Add two flags in the `save_checkpoint_args`, `save_best and `save_last`, to control the checkpoint and model weights saving behavior.

Add corresponding docs and use `exported_best.pt` everywhere in the examples.

## How has it been tested?

Manually (using unit tests for the flags would be cumbersome).

- [x] Yes

- [ ] Yes


src/lightly_train/_commands/train_task.py
```diff
@@ -478,6 +485,31 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
...
+                                    "train_model": train_model.state_dict(),
+                                }
+
+                                helpers.export_model(
+                                    out_dir=out_dir,
+                                    model_dict=model_dict,
+                                    best_or_last="best",
+                                )
+
+                            max_val_miou = val_miou
```
stegmuel, line 518: Could you add a message stating that the best mIoU was reached and that the model is saved.
Also make sure to show the value of the mIoU in the message.

src/lightly_train/_task_checkpoint.py
```diff
@@ -10,3 +10,5 @@
 
 class TaskSaveCheckpointArgs(PydanticConfig):
     save_every_num_steps: int = 1000
+    save_last: bool = True
```
stegmuel, line 20: Now it is possible for a user to set args such that the model is never saved. Do you think we should add a warning in case both flags are false?

stegmuel (commented):
Left some comments.
Additionally, shouldn't we save the best val mIoU either in the checkpoint or in a separate file?

src/lightly_train/_commands/train_task.py
yutong-xiang-97: There are indeed messages for model saving already, but I will add the message for the max val miou.

yutong-xiang-97:
> Additionally, shouldn't we save the best val mIoU either in the checkpoint or in a separate file?

I think it is good enough to have them in the training logs. What would be the benefit if it is save in the checkpoint file though?

PR #334, opened 2025-10-01
Resume MLflow runs when resuming from interrupted

yutong-xiang-97:
## What has changed and why?

Resume the same MLflow run when resuming from an interrupted training session by:

- searching with `experiment_name` and `run_name` under the same `tracking_uri`
- only logging `resume_interrupted` as a hyperparam when it is set to True
- removing logging `logger_args/mlflow/run_id` as it is redundant info and not used elsewhere.

In cases where we could not found a matching run, we start a new run and throw an warning; if multiple runs are found, we resume from the last started run.

## How has it been tested?

Manually.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ image_with_boxes = utils.draw_bounding_boxes(
 
 fig, ax = plt.subplots(figsize=(30, 30))
 ax.imshow(image_with_boxes.permute(1, 2, 0))
-fig.savefig(f"predictions_{model.model_name}.png")
+fig.savefig(f"predictions.png")
 ```
 
 </details>


src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -143,7 +144,56 @@ def get_logger_args(
...
+            f"Multiple MLflow runs found for experiment {experiment_name} and run name {run_name} when trying to resume. Resuming the most recent run."
+        )
+
+    resume_run_id: str = runs[0].info.run_id
+
+    return resume_run_id
+
+
+def get_loggers(
+    logger_args: TaskLoggerArgs, out: Path, resume_interrupted: bool
```
stegmuel, line 199: Could you add the extra arg to the doc string?

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -143,7 +144,56 @@ def get_logger_args(
...
+        )
+        return None
+
+    filter_string = f"""
+        attributes.run_name LIKE "{run_name}"
+        """
+    runs = mlflow.search_runs(
+        experiment_names=[experiment_name],
+        filter_string=filter_string,
+        order_by=["attributes.start_time DESC"],
```
stegmuel, line 179: Do you confirm that last run in at index 0 in `runs`?

stegmuel (commented):
Left two minor comments otherwise looks good.

src/lightly_train/_commands/train_task_helpers.py
yutong-xiang-97: I am not 100% sure what you mean, but it should be the last started run. An alternative is to use the last created run, but I think the former makes more sense.

PR #335, opened 2025-10-01
Use floor operation in determining DINOv2 epochs

liopeer:
## What has changed and why?
- the ceiling operation leads to a warning when using the "auto" epoch setting
- changing to `floor` from `ceil` should fix it
Screenshot 2025-10-01 at 17 00 13

## How has it been tested?
- not

- [x] Yes

- [x] Yes


stegmuel (approved):
LGTM.

PR #341, opened 2025-10-08
Prepare Release 0.11.4

liopeer:
## What has changed and why?
- bump version to 0.11.4
- rewrite CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add support for saving the best semantic segmentation checkpoints and model weights during training
-
 ### Changed
-
-- Change default precision to `bf16-mixed` for training on GPUs that support it.
 
 ### Deprecated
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,19 +9,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### Added
+
+- Add support for saving the best semantic segmentation checkpoints and model weights during training.
+- Add object detection inference support with LT-DETR models.
 
 ### Changed
 
-- Change default precision to `bf16-mixed` for training on GPUs that support it.
+- Change default precision to `bf16-mixed` for pretraining on GPUs that support it.
```
simonschoelly, line 30: Isn't that considered as a breaking change?

liopeer: It could indeed influence the training a bit, however we have never seen adverse effects with bf16 pretraining. I would therefore not consider this a breaking change. Also checkpoint will still be saved in full fp32 precision etc. so this will not have any influence on how people will load their models after resuming or into their finetuning pipeline.

PR #349, opened 2025-10-13
Add Custom Mask Names to Docs

yutong-xiang-97:
## What has changed and why?

Add changed made in #275 to docs

## How has it been tested?

[Semantic Segmentation - LightlyTrain documentation.pdf](https://github.com/user-attachments/files/22888558/Semantic.Segmentation.-.LightlyTrain.documentation.pdf)

- [ ] Yes

- [x] Yes

### DIFF
--- a/docs/source/semantic_segmentation.md
+++ b/docs/source/semantic_segmentation.md
@@ -235,7 +235,7 @@ and checkpoints organized.
 ## Data
 
 Lightly**Train** supports training semantic segmentation models with images and masks.
-Every image must have a corresponding mask whose filepath either matches the image filename (under a different directory) or follows a specific template pattern. The masks must be PNG images in either grayscale integer format, where each pixel value corresponds to a class ID, or multi-channel (e.g., RGB) format.
+Every image must have a corresponding mask whose filename either matches that of the image (under a different directory) or follows a specific template pattern. The masks must be PNG images in either grayscale integer format, where each pixel value corresponds to a class ID, or multi-channel (e.g., RGB) format.
 
 The following image formats are supported:
 


docs/source/semantic_segmentation.md
```diff
@@ -205,9 +205,7 @@ and checkpoints organized.
 ## Data
 
 Lightly**Train** supports training semantic segmentation models with images and masks.
-Every image must have a corresponding mask with the same filename except for the file
-extension. The masks must be PNG images in grayscale integer format, where each pixel
-value corresponds to a class ID.
+Every image must have a corresponding mask whose filepath either matches the image filename (under a different directory) or follows a specific template pattern. The masks must be PNG images in either grayscale integer format, where each pixel value corresponds to a class ID, or multi-channel (e.g., RGB) format.
```
stegmuel:
```suggestion
Every image must have a corresponding mask whose filename either matches that of the image (under a different directory) or follows a specific template pattern. The masks must be PNG images in either grayscale integer format, where each pixel value corresponds to a class ID, or multi-channel (e.g., RGB) format.
```

stegmuel: It would make sense to also provide an example for the case where the labels are encoded in RGB format.
We should also explain somewhere that labels can be merged (multiple values mapped to a single class). This is exemplified [here](https://github.com/lightly-ai/lightly-train/blob/f32cd12cd03c9a12d630ef1dd7c8cec4235bf10b/src/lightly_train/_data/mask_semantic_segmentation_dataset.py#L342).
Can you create a follow-up issue for  this?

yutong-xiang-97: This would be in a following PR which I will post after merging this (to avoid merge conflicts).

PR #355, opened 2025-10-14
Add class mapping support to docs

yutong-xiang-97:
## What has changed and why?

Add docs reflecting the changes in #266 and #278 

## How has it been tested?

[Semantic Segmentation - LightlyTrain documentation.pdf](https://github.com/user-attachments/files/22907391/Semantic.Segmentation.-.LightlyTrain.documentation.pdf)

- [ ] Yes

- [x] Yes

### DIFF
--- a/docs/source/semantic_segmentation.md
+++ b/docs/source/semantic_segmentation.md
@@ -380,10 +380,10 @@ The template string always uses `image_path` (of type `pathlib.Path`) to refer t
 
 We support two mask formats:
 
-- Single-channel integer masks
-- Multi-channel masks (e.g., RGB masks)
+- Single-channel integer masks, where each integer value determines a label
+- Multi-channel masks (e.g., RGB masks), where each pixel value determines a label
 
-Use the `classes` dictionary to map class IDs to labels. In this document, a **class ID** is a key in the `classes` dictionary and a **label** is its value.
+Use the `classes` dict in the `data` dict to map class IDs to labels. In this document, a **class ID** is a key in the `classes` dictionary and a **label** is its value.
 
 #### Using Integer Masks
 
@@ -398,7 +398,7 @@ For single-channel integer masks (each pixel value is a label), the default is a
     },
 ```
 
-Alternatively, to merge multiple labels into one class during training, specify in a dictionary the label name under the "name" key, and list the integer labels as the value of "labels" (type `list[int]`):
+Alternatively, to merge multiple labels into one class during training, use a dictionary like the following:
 
 ```
     "classes": {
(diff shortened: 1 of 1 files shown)


docs/source/semantic_segmentation.md
```diff
@@ -376,6 +376,63 @@ if __name__ == "__main__":
 
 The template string always uses `image_path` (of type `pathlib.Path`) to refer to the filepath of the corresponding image. Only this parameter is allowed in the template string, which is used to calculate the mask filepath.
 
+### Specify Training Classes
+
+We support two mask formats:
+
+- Single-channel integer masks
+- Multi-channel masks (e.g., RGB masks)
```
stegmuel: Can you add that each color defines a class or something similar?

docs/source/semantic_segmentation.md
```diff
@@ -376,6 +376,63 @@ if __name__ == "__main__":
...
+
+Use the `classes` dictionary to map class IDs to labels. In this document, a **class ID** is a key in the `classes` dictionary and a **label** is its value.
```
stegmuel: Can you specify that `classes` is an entry in the `data` dict argument. Otherwise it is a bit hard to grasp. Maybe add a pointer to the corresponding section in the doc and vice-versa.

docs/source/semantic_segmentation.md
```diff
@@ -376,6 +376,63 @@ if __name__ == "__main__":
...
+```
+    "classes": {                   
+        0: "background",
+        1: "car",
+        2: "bicycle",
+        # ...
+    },
+```
+
+Alternatively, to merge multiple labels into one class during training, specify in a dictionary the label name under the "name" key, and list the integer labels as the value of "labels" (type `list[int]`):
```
stegmuel: The example is quite telling. You can just write:
"To merge multiple labels into a single class, use a dictionary:"

docs/source/semantic_segmentation.md
```diff
@@ -376,6 +376,63 @@ if __name__ == "__main__":
...
+```
+
+Or:
+
+```
+    "classes": {
+        0: {"name": "background", "labels": [0]}, # The dictionary syntax also works for singleton mapping
+        1: {"name": "vehicle", "labels": [1, 2]},
+        # ...
+    },
```
stegmuel, line 418: Do you think this example is needed?
Not sure about "singleton". Maybe remove the comment.

docs/source/semantic_segmentation.md
```diff
@@ -376,6 +376,63 @@ if __name__ == "__main__":
...
+```
+
```
stegmuel, line 420: Suggestion:
Labels may share IDs, but each label maps to one class ID only.

docs/source/semantic_segmentation.md
```diff
@@ -376,6 +376,63 @@ if __name__ == "__main__":
...
+It is fine if source labels coincide with class IDs, as in the example. Only class IDs define the internal classes for training and prediction masks. Note that you cannot map the same label to multiple class IDs.
+
+#### Using Multi-channel Masks
+
+Mapping for multi-channel masks follows the same rules, except pixel values must be specified as a list of integer tuples (type `list[tuple[int, ...]]`) in the `"labels"` field:
```
stegmuel: Suggestion:
For multi-channel masks, specify pixel values as lists of integer tuples (list[tuple[int, ...]]) in the "labels" field:

stegmuel (commented):
Left some comments/suggestions.

docs/source/semantic_segmentation.md
yutong-xiang-97: This is already under the `data` section though, but I will mention it here.

docs/source/semantic_segmentation.md
yutong-xiang-97: I do think it is needed to explain that the syntax can be mixed. But I will remove the singleton comment.

docs/source/semantic_segmentation.md
yutong-xiang-97: I wouldn't use "ID" for the original labels at all to prevent confusion with class IDs. But I will take your suggestion to not use negation in the second sentence.

PR #361, opened 2025-10-20
Replace types-pillow with pillow

simonschoelly:
This PR replaces `types-pillow` with `pillow` as from `pillow` version `10.3` `types-pillow` is not necessary anymore. [See](https://pypi.org/project/types-Pillow/).

I also added a restriction for to `pillow<3.12.0` as the `helpers.create_images` function currently fails with newer pillow versions. This would theoretically be prevented by the `EXCLUDE_NEWER_DATE` variable in our Makefile but that variable is not used in targets such as `install-dev`.

I noticed that for our docker files we replace `pillow` with `pillow-simd`. We don't have that version restriction there and I am not sure if we would also need that there.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,12 +51,12 @@ dev = [
     "mdformat>=0.7.21; python_version>='3.9'",  # Older versions change formatting behavior
     "mdformat-frontmatter>=2.0.0",
     "mypy>=1.10.0",
+    "pillow>=10.3.0,<12.0.0",  # helpers.create_image currently fails for newer pillow versions
     "pre-commit>=3.5.0",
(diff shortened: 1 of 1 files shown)


src/lightly_train/_data/file_helpers.py
```diff
@@ -208,7 +208,7 @@ def _open_image_numpy__with_pil(
         ImageMode.UNCHANGED: None,
         ImageMode.MASK: None,
     }[mode]
-    image = Image.open(image_path)
+    image: PILImage | ImageFile.ImageFile = Image.open(image_path)
```
yutong-xiang-97, line 211: Why is this change needed, because of MyPy?

simonschoelly: Yes exactly. Further below the variable `image` is assigned a value of another type - and that causes MyPy to complain. I am not sure why that did not happen with `types-pillow`.

PR #372, opened 2025-10-22
Fix loading model weights from model init args

yutong-xiang-97:
## What has changed and why?

Load the `model_init_arg` when loading a checkpoint to do fine-tuning by:

- setting the relevant args as "auto" by default. The process logic is that we use the value provided by user in any case; if not, load the value in `model_init_arg` when specified; otherwise, use the default value.
- using a `CheckpointContext` class to manage model weight file loading. I kept this design cuz I do think it is much simpler and neater than the prev two functions.
- using a `load_state_dict_pre_hook` to manage potentially different `num_queries` from the model weights: the tensor value is truncated or depending on which is bigger.

Note:
- This PR does not include the changes to object detection, but it should be applied as part of the training pipeline.
- for an accidentally different `model_name`, raise an error.
- for different `classes` when `reuse_class_head` is true, raise an error.

## How has it been tested?

Manually.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/train_task.py
+++ b/src/lightly_train/_commands/train_task.py
@@ -232,14 +232,6 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
     logger.info(f"Args: {helpers.pretty_format_args(args=initial_config)}")
     logger.info(f"Using output directory: '{out_dir}")
 
-    # Load checkpoint context if resuming or further fine-tuning.
-    checkpoint = helpers.load_checkpoint(
-        fabric=fabric,
-        out_dir=out_dir,
-        resume_interrupted=config.resume_interrupted,
-        ckpt_path=config.checkpoint,
-    )
-
     # Log system information.
     system_information = _system.get_system_information()
     _system.log_system_information(system_information=system_information)
@@ -256,6 +248,14 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
     )
     train_model_cls = helpers.get_train_model_cls(
         model_name=config.model,
+    )
+
+    # Load checkpoint context if resuming or further fine-tuning.
+    checkpoint = helpers.load_checkpoint(
+        fabric=fabric,
+        out_dir=out_dir,
+        resume_interrupted=config.resume_interrupted,
+        ckpt_path=config.checkpoint,
     )
 
     if checkpoint:


src/lightly_train/_task_models/task_model_helpers.py
```diff
@@ -70,3 +73,49 @@ def _resolve_device(device: str | torch.device | None) -> torch.device:
...
+        return
+
+    num_queries_state = queries_weight.shape[0]
+    if num_queries_state == num_queries_module:
+        return
+    elif num_queries_state > num_queries_module:
+        logger.info(
+            "Checkpoint provides %d queries but module expects %d. Truncating.",
+            num_queries_state,
+            num_queries_module,
```
guarin: Please use f-strings as we do everywhere. Please change everywhere in the PR

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -72,6 +74,9 @@
 
 logger = logging.getLogger(__name__)
 
+if TYPE_CHECKING:
+    from lightly_train._commands.train_task import TrainTaskConfig
+
```
guarin: This will create circular import:
* `train_task.py` imports from `train_task_helpers.py` (good)
* `train_task_helpers.py` imports from `train_task.py` (bad)

Usually the solution for this is to move the class in question to a new file. In this case this would mean moving `TrainTaskConfig` to `train_task_config.py`.

But in this case I think we should rather simplify the code and remove the context classes.

src/lightly_train/_commands/train_task.py
```diff
@@ -230,6 +230,20 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
...
+    )
+    if checkpoint_ctx:
+        model_init_args = checkpoint_ctx.metadata.model_init_args
+        if (saved_model_name := model_init_args.get("model_name")) != config.model:
+            raise ValueError(
+                f"The model name in the checkpoint or model weights file('{saved_model_name}') "
+                f"does not match the provided model name ('{config.model}')."
+            )
+    else:
+        model_init_args = {}
```
guarin: Move the logic as far down in the function as possible, right before `model_init_args` is used the first time. There is code that is more important that should be run before just below (e.g. logging system information).

src/lightly_train/_commands/train_task.py
```diff
@@ -230,6 +232,14 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
     logger.info(f"Args: {helpers.pretty_format_args(args=initial_config)}")
     logger.info(f"Using output directory: '{out_dir}")
 
+    # Load checkpoint context if resuming or further fine-tuning.
+    checkpoint = helpers.load_checkpoint(
+        fabric=fabric,
+        out_dir=out_dir,
```
guarin: Load this further down right before the `if checkpoint:` line. Logging system information etc. is more important code and must run as early as possible in the code.

PR #373, opened 2025-10-22
Download checkpoint if it does not exist as a file

simonschoelly:
## What has changed and why?

The `load_model_from_checkpoint` function now tries to download a model/checkpoint from our s3 bucket if the checkpoint path does not exist locally.

Current behavior:
- If the path specified is a local file then that one is loaded.
- If not and the path is a `str` it tries to download the model.
- Models are cached and a checksum is uses to avoid download multiple times.
- The cache directory is specified by the environment variable `LIGHTLY_TRAIN_MODEL_CACHE_DIR`.
- If a model is already downloaded but has a different checksum then it is downloaded again and will be replaced.
- The name of the model is given by the last part in the download url. For example when using the name `dinov2/vits14-ltdetr-coco` then the url is `/dinov2_ltdetr/ltdetr_vits14dinov2_coco.ckpt` and the file will be stored under `{LIGHTLY_TRAIN_MODEL_CACHE_DIR}/ltdetr_vits14dinov2_coco.ckpt`.
- As it is always the case with file handling there are some edge cases that might not be covered and will not fail graciously - for example if the path to a cached model already exists but is a directory.

Things that we should do in the future:
- The list of models and their checksum is currently hard coded. Instead we should download the list of available models.
- I will create an issue for that.

TODO
- Update docs.

## How has it been tested?

Some small tests have been written. I also verified manually that all models that we have can be downloaded.

- [x] Yes

Updated the README.md - @yutong-xiang-97 will update `semantic_segmentation.md` later.

- [ ] Yes

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,17 +146,13 @@ LightlyTrain’s LT-DETR models, powered by DINOv2 and DINOv3 backbones, demonst
 
 🚀 We are actively working on new models with improved speed and accuracy. Updates coming soon, so stay tuned!
 
-```bash
-wget <MODEL-WEIGHTS-URL> -O model.ckpt
-```
-
 ```python
 import lightly_train
 from torchvision import utils, io
 import matplotlib.pyplot as plt
 
 model = lightly_train.load_model_from_checkpoint(
-    checkpoint="model.ckpt",
+    checkpoint="dinov3/convnext-tiny-ltdetr-coco"
 )
 
 labels, boxes, scores = model.predict("<image>.jpg").values()


README.md
```diff
@@ -25,47 +25,46 @@ With LightlyTrain you can train your very own foundation model like DINOv2 on yo
...
-| LightlyTrain | dinov3/convnext-base-ltdetr | 58.6 | 24.68 | 121.0 | 640×640 | [🔗](https://lightly-train-checkpoints.s3.us-east-1.amazonaws.com/dinov3_ltdetr/ltdetr_convnext-base_coco.ckpt) |
-| LightlyTrain | dinov3/convnext-large-ltdetr | 60.0 | 42.30 | 230.0 | 640×640 | [🔗](https://lightly-train-checkpoints.s3.us-east-1.amazonaws.com/dinov3_ltdetr/ltdetr_convnext-large_coco.ckpt) |
+| Implementation | Backbone Model | AP<sub>50:95</sub> | Latency (ms) | # Params (M) | Input Size | Checkpoint Name |
+|:--------------:|:----------------------------:|:------------------:|:------------:|:------------:|:----------:|:---------------------------------:|
+| LightlyTrain | dinov2/vits14-ltdetr | 55.7 | 16.87 | 55.3 | 644×644 | dinov2/vits14-ltdetr-coco |
+| LightlyTrain | dinov3/convnext-tiny-ltdetr | 54.4 | 13.29 | 61.1 | 640×640 | dinov3/convnext-tiny-ltdetr-coco |
+| LightlyTrain | dinov3/convnext-small-ltdetr | 56.9 | 17.65 | 82.7 | 640×640 | dinov3/convnext-small-ltdetr-coco |
+| LightlyTrain | dinov3/convnext-base-ltdetr | 58.6 | 24.68 | 121.0 | 640×640 | dinov3/convnext-base-ltdetr-coco |
```
guarin, line 33: Could you update this part further down in the README to use the model name instead of downloading the model with wget?

````
```bash
wget <MODEL-WEIGHTS-URL> -O model.ckpt
```
```python
import lightly_train
from torchvision import utils, io
import matplotlib.pyplot as plt
    checkpoint="model.ckpt",
)
````

simonschoelly: Done

PR #374, opened 2025-10-23
Fix NormalizeArgs from dict

guarin:
## What has changed and why?

* Fix NormalizeArgs.from_dict for multi-channel images


yutong-xiang-97 (approved):
Approved.

PR #379, opened 2025-10-27
Thomas trn 1623 object detection metrics

stegmuel:
## What has changed and why?

Added mAP metrics for object detection.
Documentation will be added in following PRs.

## How has it been tested?

Manual tests.

- [x] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_data/file_helpers.py
+++ b/src/lightly_train/_data/file_helpers.py
@@ -230,18 +230,10 @@ def open_yolo_object_detection_label_numpy(
     for line in _iter_yolo_label_lines(label_path=label_path):
         parts = [float(x) for x in line.split()]
         class_id = parts[0]
-        # Box format: class_id x_center y_center width height
-        if len(parts) == 5:
-            x_center = parts[1]
-            y_center = parts[2]
-            width = parts[3]
-            height = parts[4]
-        # Polygon format: class_id x0 y0 x1 y1 ...
-        else:
-            x_center, y_center, width, height = _bbox_from_polygon(
-                polygon=np.array(parts[1:], dtype=np.float64)
-            )
-
+        x_center = parts[1]
+        y_center = parts[2]
+        width = parts[3]
+        height = parts[4]
         bboxes.append([x_center, y_center, width, height])
         classes.append(int(class_id))
     bboxes_np = np.array(bboxes) if bboxes else np.zeros((0, 4), dtype=np.float64)


src/lightly_train/_task_models/object_detection_components/hybrid_encoder.py
```diff
@@ -303,9 +303,12 @@ def __init__(
...
         for _ in range(len(in_channels) - 1):
-            self.downsample_convs.append(
-                ConvNormLayer(hidden_dim, hidden_dim, 3, 2, act=act)
-            )
+            # Allocate only when upsampling is enabled. Otherwise, this layer is detected as an
+            # unused parameter during distributed training, causing training to fail.
+            if self.upsample:
+                self.downsample_convs.append(
+                    ConvNormLayer(hidden_dim, hidden_dim, 3, 2, act=act)
+                )
```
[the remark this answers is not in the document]
stegmuel: `downsample_feat = self.downsample_convs[idx](feat_low)` is under an if statement. I think copilot is wrong.

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -82,6 +82,22 @@
 ]
 
 
+TASK_TO_METRICS: dict[str, dict[str, str]] = {
```
liopeer, line 86: Wouldn't address it right now, but optimally we create a new type here e.g. a Literal in `types.py` for defining metrics. And then this would change to:

```python
from lightly_train.types import TrainMetric

TASK_TO_METRICS: dict[str, TrainMetric[str, str]] = {
```

and we could accordingly define the `TaskSaveCheckpointArgs`'s `watch_metric` with the same type.

I would just create a follow-up issue.

src/lightly_train/_data/file_helpers.py
```diff
@@ -230,10 +230,18 @@ def open_yolo_object_detection_label_numpy(
...
-        width = parts[3]
-        height = parts[4]
+        # Box format: class_id x_center y_center width height
+        if len(parts) == 5:
+            x_center = parts[1]
+            y_center = parts[2]
+            width = parts[3]
+            height = parts[4]
+        # Polygon format: class_id x0 y0 x1 y1 ...
+        else:
```
liopeer: Not sure if we want to merge this. I'd rather have this fail if it doesn't follow the specification defined in our documentation (I know we use it for our internal benchmarking). But okay with leaving it.

src/lightly_train/_data/yolo_object_detection_dataset.py
```diff
@@ -80,6 +80,7 @@ def __getitem__(self, index: int) -> ObjectDetectionDatasetItem:
             image=image,
             bboxes=bboxes,
             classes=class_labels,
+            original_size=(w, h),
```
liopeer: Would prefer `h, w` since that's PyTorch convention, but doesn't need to be addressed rn. We need to anyway define some better types and conventions for passing stuff around in internal code once OD is finished.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
```diff
@@ -120,6 +124,9 @@ def __init__(
         val_transform_args: DINOv2LTDETRObjectDetectionValTransformArgs,
     ) -> None:
         super().__init__()
+        # Lazy import torchmetrics because it is an optional dependency.
+        from torchmetrics.detection.mean_ap import MeanAveragePrecision
```
liopeer: According to the current `pyproject.toml` it is not an optional dependency, but a core dependency. If we leave it that way can you print a nice error message with a try:, except: that guides the user to install the correct optional dependencies.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -119,6 +123,9 @@ def __init__(
         val_transform_args: DINOv3LTDETRObjectDetectionValTransformArgs,
     ) -> None:
         super().__init__()
+        # Lazy import torchmetrics because it is an optional dependency.
+        from torchmetrics.detection.mean_ap import MeanAveragePrecision
```
liopeer: Same here.

src/lightly_train/_task_models/object_detection_components/utils.py
```diff
@@ -12,12 +12,14 @@
 # limitations under the License.#
 """Copyright(c) 2023 lyuwenyu. All Rights Reserved."""
```
liopeer: Since this is an external file, we should document the diffs somewhere. Either right here or in the `NOTICE` file. Alternatively, you could also move the new functions to a different module (not inside `object_detection_utils.py`) because this is all marked as copy-pasted code, see `Makefile` where do create the headers.

pyproject.toml
```diff
@@ -106,6 +106,7 @@ onnxruntime = [
 onnxslim = [
     "onnxslim>=0.1.67",
 ]
+object-detection = ["pycocotools>=2.0.7"]
```
liopeer: This would also have to be mentioned in the documentation (otherwise the run will fail at the beginning of the validation). Since it's a super light-weight dependency I would actually put it into the core dependencies for the sake of convenience for the users. We can later maybe refactor and create a group `finetuning` that includes transformers, torchmetrics and pycocotools (and anything else that comes).

src/lightly_train/_commands/train_task_helpers.py
stegmuel: Left a TODO.

src/lightly_train/_task_models/object_detection_components/utils.py
stegmuel: Did both.

pyproject.toml
guarin: I assume you are using torchmetrics, in that case it is better to install `torchmetrics[detection]` which will handle the pycocotols or coco-fast-eval install behind the scenes.

src/lightly_train/_data/file_helpers.py
stegmuel: What about updating the doc?

src/lightly_train/_data/yolo_object_detection_dataset.py
stegmuel: Left a TODO.
The reason it is (w, h) is to match the expected format of the post-processor, which is the only place it is used.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
stegmuel: Updated the import to reflect the `pyproject.toml`.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
stegmuel: Updated the import to reflect the `pyproject.toml`.

src/lightly_train/_data/file_helpers.py
liopeer: Is that a common thing to do? Using instance segmentation datasets for OD? I just don't want to make it too complicated with too many options for the user. But as I said: Also fine with just leaving it.

stegmuel: reverted.

liopeer (approved):
LGTM!

PR #382, opened 2025-10-28
Add EoMT instance segmentation loss

guarin:
## What has changed and why?

* Add EoMT instance segmentation loss
* Fix license headers for EoMT files

## How has it been tested?

-

- [ ] Yes

- [ ] Yes


src/lightly_train/_task_models/dinov3_eomt_instance_segmentation/mask_loss.py
```diff
@@ -0,0 +1,121 @@
+#
+# ---------------------------------------------------------------
```
liopeer, line 2: While reviewing I realized that that the `mask_loss.py` in `dinov3_eomt_semantic_segmentation` does not have the correct header. Could you check the `Makefile` and add any missing paths?

liopeer: Nevermind. You already did that. 😄

src/lightly_train/_task_models/dinov3_eomt_semantic_segmentation/mask_loss.py
```diff
@@ -1,9 +1,8 @@
 #
-# Copyright (c) Lightly AG and affiliates.
-# All rights reserved.
-#
-# This source code is licensed under the license found in the
-# LICENSE file in the root directory of this source tree.
+# ---------------------------------------------------------------
```
liopeer, line 2: ❤️

PR #384, opened 2025-10-28
Implement predict semantic segmentation with `model.predict`

yutong-xiang-97:
## What has changed and why?

Added to `_commands/predict_task.py` a `predict_semantic_segmentation` function to predict masks for a full dataset with a semantic segmentation model checkpoint. We've also created necessary helper functions and transforms and added them to `_commands/predict_task_helpers.py` and `_transforms/` respectively.

## How has it been tested?

by:
- integration test
- manually comparing the masks with the masks generated by the script we used for the benchmark results.

- [x] Yes

- [ ] Yes
- [x] Not needed (will do in a follow-up)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -193,7 +193,7 @@ DOCKER_EXTRAS := --extra mlflow --extra tensorboard --extra timm --extra wandb -
 
 # Date until which dependencies installed with --exclude-newer must have been released.
 # Dependencies released after this date are ignored.
-EXCLUDE_NEWER_DATE := "2025-09-07"
+EXCLUDE_NEWER_DATE := "2025-10-28"
 
 # Pinned versions for Torch and TorchVision to avoid issues with the CUDA/driver version
 # on the CI machine. These versions are compatible with CUDA 11.4 and Python 3.8.
--- a/docs/source/models/rtdetr.md
+++ b/docs/source/models/rtdetr.md
@@ -53,9 +53,10 @@ be executed from inside the `RT-DETR/rtdetrv2_pytorch` directory or the
 `RT-DETR/rtdetrv2_pytorch` directory must be added to the Python path.
 
 ```python
+from src.core import YAMLConfig
+
 import lightly_train
 from lightly_train.model_wrappers import RTDETRModelWrapper
-from src.core import YAMLConfig
 
 if __name__ == "__main__":
     # Load the RT-DETR model
--- a/src/lightly_train/_models/ultralytics/ultralytics.py
+++ b/src/lightly_train/_models/ultralytics/ultralytics.py
@@ -22,7 +22,7 @@ from lightly_train._models.model_wrapper import (
 )
 
 if TYPE_CHECKING:
-    from ultralytics import YOLO
+    from ultralytics import YOLO  # type: ignore[attr-defined]
     from ultralytics.nn.modules.block import SPPF, C2f
     from ultralytics.nn.modules.conv import Conv
 
--- a/src/lightly_train/_models/ultralytics/ultralytics_package.py
+++ b/src/lightly_train/_models/ultralytics/ultralytics_package.py
@@ -16,7 +16,7 @@ import torch
 from torch.nn import Module
 
 if TYPE_CHECKING:
-    from ultralytics import YOLO
+    from ultralytics import YOLO  # type: ignore[attr-defined]
 
 from lightly_train._models import log_usage_example
 from lightly_train._models.model_wrapper import ModelWrapper
@@ -75,7 +75,7 @@ class UltralyticsPackage(Package):
     @classmethod
     def is_supported_model(cls, model: Module | ModelWrapper) -> bool:
         try:
-            from ultralytics import YOLO
+            from ultralytics import YOLO  # type: ignore[attr-defined]
         except ImportError:
             return False
         if isinstance(model, ModelWrapper):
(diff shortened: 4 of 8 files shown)


src/lightly_train/_commands/predict_task.py
```diff
@@ -0,0 +1,208 @@
...
+        )
+        try:
+            predict_task_helpers.save_mask(mask_numpy, mask_filepath)
+        except Exception as e:
+            logger.error(
+                f"Could not save predicted mask for image "
+                f"'{image_filename}' at '{mask_filepath}': {e}"
+            )
+        if idx % config.log_every_num_steps == 0:
+            logger.info(f"{idx}/{num_images} completed.")
```
guarin: Suggestion: this mirrors the `Step 5/100` message from training
```suggestion
            logger.info(f"Images {(idx + 1) * config.batch_size}/{num_images}")
```

src/lightly_train/_commands/predict_task.py
```diff
@@ -0,0 +1,208 @@
...
+    ):
+        logger.error(
+            f"Some data paths could not be found: {diff}. Skipping them during prediction."
+        )
+
+    predict_model = fabric.setup_module(model)
+    predict_model.mark_forward_method("predict")
+
+    # TODO(Yutong, 10/25): re-implement with predict_batch
+    logger.info("Starting prediction...")
```
guarin: Suggestion:
```suggestion
    logger.info("Creating predictions for {num_images} images...")
```

src/lightly_train/_commands/predict_task_helpers.py
```diff
@@ -0,0 +1,203 @@
...
+
+    return out_dir
+
+
+def get_transform_cls(model_cls_name: str) -> type[PredictTransform]:
+    for task_model_cls in SEM_SEG_TASK_MODEL_CLASSES:
+        if task_model_cls == model_cls_name:
+            return PredictSemanticSegmentationTransform
+    # TODO(Yutong, 10/25): add more task model classes here once implemented
+    raise ValueError(f"Unsupported model class '{model_cls_name}'.")
```
guarin: Request: Instead of listing classes explicitly in SEM_SEG_TASK_MODEL_CLASSES I would check whether the class name contains "semanticsegmentation". `if "semanticsegmentation" in model_cls_name.lower():` Otherwise it is easy to forget adding new classes to the `SEM_SEG_TASK_MODEL_CLASSES` list.

src/lightly_train/_commands/predict_task_helpers.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+def get_transform(
+    model: TaskModel,
+) -> PredictTransform:
+    model_cls_name = model.class_path.split(".")[-1]
+    transform_cls = get_transform_cls(model_cls_name)
+
+    transform_args = transform_cls.transform_args_cls.from_model(model)
```
guarin: No

src/lightly_train/_transforms/predict_semantic_segmentation_transform.py
```diff
@@ -0,0 +1,83 @@
...
+from lightly_train._transforms.predict_transform import (
+    PredictTransform,
+    PredictTransformArgs,
+)
+from lightly_train.types import TransformInput, TransformOutput
+
+
+class PredictSemanticSegmentationNormalizeArgs(PydanticConfig):
+    mean: tuple[float, ...]
+    std: tuple[float, ...]
```
guarin: No need to create a new PredictSemanticSegmentationNormalizeArgs class. You can import `NormalizeArgs` from `_transforms/transform.py` instead.

src/lightly_train/_transforms/predict_semantic_segmentation_transform.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class PredictSemanticSegmentationTransformArgs(PredictTransformArgs):
+    image_size: tuple[int, int]
+    normalize: PredictSemanticSegmentationNormalizeArgs
+
+    @classmethod
+    def from_model(cls, model: TaskModel) -> "PredictSemanticSegmentationTransformArgs":
+        if not (image_size := getattr(model, "image_size", None)):
```
guarin: Request: This method shouldn't take `TaskModel` as input to reduce risk of cyclic imports and keep code well separated. If we import models here the transforms suddenly start depending on models which themselves depend on many other classes which creates a lot of dependencies in the codebase. Best to keep the two separated. 

The solution to this is to pass the values directly instead of the model when the class is initialized:
```
PredictSemanticSegmentationTransformArgs(
    image_size=model.image_size,
    normalize=NormalizeArgs(mean=model.normalize["mean"], std=model.normalize["std"]),
)
```

src/lightly_train/_commands/predict_task.py
```diff
@@ -0,0 +1,208 @@
...
+    initial_data = initial_config["data"]
+    resolved_data = resolved_config["data"]
+    if (
+        isinstance(initial_data, list)
+        and isinstance(resolved_data, list)
+        and (diff := (set(initial_data) - set(resolved_data)))
+    ):
+        logger.error(
+            f"Some data paths could not be found: {diff}. Skipping them during prediction."
+        )
```
guarin: Don't think this is needed. `list_image_filenames_from_iterable` which is called in `get_dataset` already throws an error if invalid paths are encountered.

src/lightly_train/_transforms/predict_transform.py
```diff
@@ -0,0 +1,35 @@
...
+from lightly_train._task_models.task_model import TaskModel
+from lightly_train.types import TransformInput, TransformOutput
+
+
+class PredictTransformArgs(PydanticConfig):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    @classmethod
+    def from_model(cls, model: TaskModel) -> "PredictTransformArgs":
+        raise NotImplementedError
```
guarin:
```suggestion
```

PR #387, opened 2025-10-28
Fix same mmapfile

stegmuel:
## What has changed and why?

Use data and out as identifier for hashing to create the names of the mmap files.
This avoid clashes between validation and train sets.

## How has it been tested?

Manual tests.

- [x] Yes

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Same train and validation mmap file.
```
yutong-xiang-97: Could you elaborate on what is wrong in which situtation? E.g.:
```suggestion
- Fix the collision between the train and validation mmap files when using `out` as the sole filename identifier during fine-tuning.
```

PR #389, opened 2025-10-28
EoMT instance segmentation task model

guarin:
## What has changed and why?

* Add EoMT instance segmentation task model

Biggest difference to semantic segmentation model is that tile/untile are not used and instead all images are resized to 640x640. Another difference is how the labels/masks/scores are calculated. Check https://github.com/tue-mps/eomt/blob/660778b9641c1bacbb5b0249ee3dcb684d9c94d9/training/mask_classification_instance.py#L88-L89 for comparison.

## How has it been tested?

* Not tested yet, need train model first

- [ ] Yes

- [ ] Yes

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -13,6 +13,7 @@ When importing:
   - Third-party imports
   - Local imports
     and group them accordingly with a blank line between each group.
+- Always add `from __future__ import annotations` at the top of the file.
 
 ## File Layout
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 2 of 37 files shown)


src/lightly_train/_task_models/dinov3_eomt_instance_segmentation/task_model.py
```diff
@@ -0,0 +1,544 @@
...
+                The URL to the DINOv3 backbone weights. This is used to download the
+                weights.
+            backbone_args:
+                Additional arguments to pass to the DINOv3 backbone.
+        """
+        super().__init__(locals(), ignore_args={"backbone_weights", "backbone_url"})
+        parsed_name = self.parse_model_name(model_name=model_name)
+        self.model_name = parsed_name["model_name"]
+        self.classes = classes
+        self.class_ignore_index = class_ignore_index
```
stegmuel: Is this still needed for instance segmentation?

guarin: `class_ignore_index` is not needed, I removed it. `classes` is still needed as we have to map internal class ids in [0, num_classes - 1] back to the original ids.

PR #393, opened 2025-10-29
Fix CI errors: add `device` parameter to the predict function

yutong-xiang-97:
## What has changed and why?

Let user set the `device` parameter in `predict_semantic_segmentation` function to setup which or how many GPUs are used for inference.

## How has it been tested?

Passed CI tests.

- [ ] Yes

- [ ] Yes


guarin (approved):
LGTM

PR #399, opened 2025-10-30
Reduce memory usage for predict and update model hash

yutong-xiang-97:
## What has changed and why?

As the title suggests

## How has it been tested?

Manually

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/predict_task.py
+++ b/src/lightly_train/_commands/predict_task.py
@@ -41,6 +41,7 @@ def predict_semantic_segmentation(
     num_workers: int | Literal["auto"] = "auto",
     accelerator: str | Accelerator = "auto",
     devices: int | str | list[int] = 1,
+    remove_cache: bool = False,  # TODO(Yutong, 10/25): remove/improve this when re-implementing with predict_batch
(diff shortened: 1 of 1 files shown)


stegmuel (commented):
Why is this needed? Did you encounter OOM errors?
Did you observe significant memory reduction? batch and masks are in any case overwritten in the next iteration, no? 
What about speed? `torch.cuda.empty_cache()` has an overhead, no?

yutong-xiang-97:
> Why is this needed? Did you encounter OOM errors?

Encountered this when making the tutorial on Colab with T4.

> Did you observe significant memory reduction? batch and masks are in any case overwritten in the next iteration, no?

There is indeed a significant drop of memory consumption, like 40GB vs 10GB on B200. The major drop comes from `torch.cuda.empty_cache()` but deleting batch and masks still helps a bit and prevents from occasional spikes.

> What about speed? `torch.cuda.empty_cache()` has an overhead, no?

This predict function will in any case be improved with a dedicated function with batch processing where we can do some optimizations.

stegmuel (approved):
LGTM

PR #404, opened 2025-11-03
Add EoMT instance segmentation train model

guarin:
## What has changed and why?

* Add EoMT instance segmentation train model

## How has it been tested?

Will test it when adding `train_instance_segmentation` in the next PR

- [ ] Yes

- [ ] Yes

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,11 +2,13 @@
 
 ## Development
 
+### Setting up the Development Environment
+
 ```
 git clone https://github.com/lightly-ai/lightly-train.git
-uv venv
-make install-dev
+uv venv .venv
 source .venv/bin/activate
+make install-dev
 ```
 
 Make sure the environment is activated before running the following commands.
@@ -14,9 +16,18 @@ Make sure the environment is activated before running the following commands.
 > [!WARNING]\
 > Prepending commands with `uv run` might not work properly. Activate the environment directly instead.
 
+### Running Checks and Tests
+
+Before committing code, make sure all tests and checks pass:
+
 ```
 make format
 make static-checks
+```
+
+and if you want to run all the tests:
+
+```
 make test
 ```
 


src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -619,6 +619,7 @@ def log_step(
     name_to_display_name = {
         "train_loss": "Train Loss",
         "train_metric/miou": "Train mIoU",
+        "train_metric/map": "Train mAP@0.5:0.95",
```
stegmuel: We do not keep track of the mAP during  OD training (if I remember correctly).

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -644,8 +645,18 @@ def compute_metrics(log_dict: dict[str, Any]) -> dict[str, Any]:
             for i, v in enumerate(value):
                 metrics[f"{name}_{i}"] = v.item()
         if isinstance(value, dict):
-            for class_name, class_value in value.items():
-                metrics[f"{name}{class_name}"] = class_value.item()
+            if "map" in value:
+                # Special case for detection metrics which return results like this:
+                # {"map": 0.5, "map_50": 0.7, ...}
+                if name.endswith("/map"):
```
stegmuel, line 701: This should fix itself with the merging in the main, but the classes are return by the mAP metrics and should be handled.

src/lightly_train/_commands/train_task_helpers.py
stegmuel: Ah now I see that it's done for instance segmentation...

src/lightly_train/_task_models/dinov3_eomt_instance_segmentation/train_model.py
```diff
@@ -0,0 +1,486 @@
...
+    def __init__(
+        self,
+        *,
+        model_name: str,
+        model_args: DINOv3EoMTInstanceSegmentationTrainArgs,
+        data_args: YOLOInstanceSegmentationDataArgs,
+        val_transform_args: DINOv3EoMTInstanceSegmentationValTransformArgs,
+    ) -> None:
+        super().__init__()
+        # Lazy import because torchmetrics is an optional dependency.
```
stegmuel, line 167: If torchmetrics is not optional should the imports be done at the top of the file? And similarly for semantics segmentation?

stegmuel (commented):
First round of comments/questions

src/lightly_train/_commands/train_task_helpers.py
guarin: Updated

src/lightly_train/_task_models/dinov3_eomt_instance_segmentation/train_model.py
guarin: I guess, but would prefer to do that in a different pr across the codebase

stegmuel (approved):
LGTM.

PR #406, opened 2025-11-03
Add event tracking

IgorSusmelj:
## What has changed and why?

- adds `_event` module to track anonymous usage
- tracks two event types (training started and inference started) with minimal metadata (training method, model architecture, OS, GPU name)
- we rate limit (only send a request every 30 seconds
- we also use a queue of size 100 to avoid consuming too much memory (e.g. if there is no internet connection)
- we use background daemon threads that are non blocking (they are automatically killed if the program ends)
- only use built-in python libs (no new dependencies)
- Opt-out via environment variable: LIGHTLY_TRAIN_EVENTS_DISABLED=1
- add a note about usage tracking to readme

## How has it been tested?

- added new unit tests for events
- update unit tests for callbacks

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/predict_task_helpers.py
+++ b/src/lightly_train/_commands/predict_task_helpers.py
@@ -7,6 +7,7 @@
 #
 from __future__ import annotations
 
+import json
 import logging
 from pathlib import Path
 from typing import Any, Sequence
@@ -15,6 +16,7 @@ import numpy as np
 from lightning_fabric import Fabric
 from lightning_fabric import utilities as fabric_utilities
 from PIL import Image
+from torch import Tensor
 from torch.utils.data import DataLoader
 
 from lightly_train._configs.validate import pydantic_model_validate
@@ -204,3 +206,48 @@ def save_mask(mask: NDArrayMask, mask_filepath: Path) -> None:
     mask_np = mask_np.astype(dtype, copy=False)
 
     Image.fromarray(mask_np).save(mask_filepath)
+
+
+def prepare_coco_entries(
+    predictions: dict[str, Tensor],
+    image_size: tuple[int, int],
+) -> list[dict[str, Any]]:
+    width, height = image_size
+
+    labels: list[int] = predictions["labels"].detach().cpu().tolist()
+    boxes: list[list[float]] = predictions["bboxes"].detach().cpu().tolist()
+    scores: list[float] = predictions["scores"].detach().cpu().tolist()
+
+    entries = []
+    for label, box, score in zip(labels, boxes, scores):
+        x1, y1, x2, y2 = box
+        x1 = max(0.0, min(float(x1), float(width)))
+        y1 = max(0.0, min(float(y1), float(height)))
+        x2 = max(0.0, min(float(x2), float(width)))
+        y2 = max(0.0, min(float(y2), float(height)))
+
+        x = int(round(x1))
+        y = int(round(y1))
+        x2_rounded = int(round(x2))
+        y2_rounded = int(round(y2))
+        w = max(0, x2_rounded - x)
+        h = max(0, y2_rounded - y)
+
+        rounded_score = round(score, 2)
+
+        entries.append(
+            {
+                "category_id": label,
+                "bbox": [x, y, w, h],
+                "score": rounded_score,
+            }
+        )
+    return entries
+
+
+def save_coco_json(entries: list[dict[str, Any]], coco_filepath: Path) -> None:
+    coco_filepath.parent.mkdir(parents=True, exist_ok=True)
+    coco_filepath.write_text(
+        json.dumps({"predictions": entries}, indent=2) + "\n",
+        encoding="utf-8",
+    )


src/lightly_train/_commands/train.py
```diff
@@ -333,13 +334,21 @@ def train_from_config(config: TrainConfig) -> None:
         config.callbacks = callback_helpers.get_callback_args(
             callback_args=config.callbacks
         )
+        event_info = TrainingEventInfo(
+            method=config.method,
+            model=str(config.model),
```
guarin:
```suggestion
            model=config.model if isinstance(config.model, str) else config.model.__class__.__name__,
```

Otherwise it will log the full pytorch model structure, which is also ok if you intended that

src/lightly_train/_events/__init__.py
```diff
@@ -0,0 +1,10 @@
+#
+# Copyright (c) Lightly AG and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+#
+from lightly_train._events.tracker import track_event
+
+__all__ = ["track_event"]
```
guarin: No need for this, we only import in `__init__` for publicly exposed functions/classes

src/lightly_train/_events/config.py
```diff
@@ -0,0 +1,11 @@
...
+# Copyright (c) Lightly AG and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+#
+import os
+
+POSTHOG_API_KEY: str = os.getenv("LIGHTLY_TRAIN_POSTHOG_KEY", "phc_placeholder")
+EVENTS_DISABLED: bool = os.getenv("LIGHTLY_TRAIN_EVENTS_DISABLED", "0") == "1"
```
guarin: Add to `_env.py` instead

src/lightly_train/_callbacks/events.py
```diff
@@ -0,0 +1,35 @@
...
+
+class EventsCallback(Callback):
+    """Callback to track training events."""
+
+    def __init__(self, event_info: TrainingEventInfo) -> None:
+        super().__init__()
+        self.event_info = event_info
+
+    def on_train_start(self, trainer: Trainer, pl_module: LightningModule) -> None:
+        """Track training_started event."""
```
guarin: Did you check that this only runs on rank 0?

src/lightly_train/_events/tracker.py
```diff
@@ -0,0 +1,88 @@
...
+        "api_key": config.POSTHOG_API_KEY,
+        "event": event_name,
+        "distinct_id": _session_id,
+        "properties": {**properties, **_get_system_info()},
+    }
+    _events.append(event_data)
+
+    if current_time - _last_flush >= _RATE_LIMIT_SECONDS:
+        _last_flush = current_time
+        threading.Thread(target=_flush, daemon=True).start()
```
guarin, line 100: Why not pass the event directly to the thread instead of adding it to `_events`?

src/lightly_train/_events/tracker.py
```diff
@@ -0,0 +1,88 @@
...
+            pass
+
+
+def _get_system_info() -> Dict[str, Any]:
+    """Collect minimal system info."""
+    global _system_info
+    if _system_info is None:
+        gpu_name = None
+        if torch.cuda.is_available():
+            gpu_name = torch.cuda.get_device_name(0)
```
guarin, line 60: There is also `lightly_train/_system.py::get_system_information` which contains additional information

src/lightly_train/_callbacks/events.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class EventsCallback(Callback):
+    """Callback to track training events."""
+
+    def __init__(self, event_info: TrainingEventInfo) -> None:
+        super().__init__()
+        self.event_info = event_info
+
+    def on_train_start(self, trainer: Trainer, pl_module: LightningModule) -> None:
```
guarin: This will not get triggered for `embed` and `predict_task` as they don't involve training

src/lightly_train/_callbacks/events.py
IgorSusmelj: No, I added a check now!

src/lightly_train/_callbacks/events.py
IgorSusmelj: Yes, I think for them we did not use the callback. So we manually call to track events there. I wasn't sure if we want to use callbacks there as well.

src/lightly_train/_events/tracker.py
IgorSusmelj: Two motivations:
1. have fewer threads (we use one thread to go through the whole queue)
2. we won't miss any events that way (we can always add to the list and only the thread actively processing the list can flush it)

IgorSusmelj: Added a comment to track_event

src/lightly_train/_events/tracker.py
IgorSusmelj: I didn't know about it. I will leave a TODO to see if we should/can merge them or not.

src/lightly_train/_events/tracker.py
```diff
@@ -0,0 +1,96 @@
...
+    return _system_info
+
+
+def track_event(event_name: str, properties: Dict[str, Any]) -> None:
+    """Track an event.
+
+    Events are buffered so flushes can send them in batches. This limits how many
+    threads we spawn and ensures no events are dropped while a flush runs.
+    """
+    global _last_flush
```
guarin, line 77:
```suggestion
    """
    if not _distributed.is_global_rank_zero():
        return

    global _last_flush
```
Import from:

```
from lightly_train import _distributed
```

This way we can be sure that it never logs from any other rank than rank 0 no matter from where `track_event` is called. Right now it is manually called in e.g. `_predict_task_from_config` so the rank zero check in the callback is not enough.

src/lightly_train/_events/tracker.py
guarin: > have fewer threads (we use one thread to go through the whole queue)

Aren't events only added to the queue through `track_events` and every `track_events` call spawns a new thread?

I don't think it is an issue, just don't completely understand the logic

src/lightly_train/_callbacks/events.py
guarin: I wouldn't use a callback but instead call `track_event` directly. I am not sure if the callback is needed in general as we only log at the start of training.

guarin (approved):
LGTM!

PR #408, opened 2025-11-04
EoMT instance segmentation handle image size

guarin:
## What has changed and why?

* Handle validation image size

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_task_models/dinov3_eomt_instance_segmentation/transforms.py
+++ b/src/lightly_train/_task_models/dinov3_eomt_instance_segmentation/transforms.py
@@ -128,9 +128,7 @@ class DINOv3EoMTInstanceSegmentationValTransformArgs(InstanceSegmentationTransfo
     random_flip: RandomFlipArgs | None = None
     color_jitter: ColorJitterArgs | None = None
     scale_jitter: ScaleJitterArgs | None = None
-    smallest_max_size: SmallestMaxSizeArgs = Field(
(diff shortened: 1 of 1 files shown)


src/lightly_train/_task_models/dinov3_eomt_instance_segmentation/train_model.py
```diff
@@ -179,7 +181,11 @@ def __init__(
         self.model_args = model_args
         num_queries = no_auto(self.model_args.num_queries)
         num_joint_blocks = no_auto(self.model_args.num_joint_blocks)
-        image_size = no_auto(val_transform_args.image_size)
+        image_size_train = no_auto(train_transform_args.image_size)
+        image_size_val = no_auto(val_transform_args.image_size)
+        image_size = (
+            image_size_val if isinstance(image_size_val, tuple) else image_size_train
```
stegmuel, line 187: Did not get why whether or not image_size_val is a tuple or not determines to use image_size_train or image_size_val.

stegmuel (commented):
Left one question.

src/lightly_train/_task_models/dinov3_eomt_instance_segmentation/train_model.py
guarin: It can either be a tuple or None. If it is None (the default) we use the train image size (640 by default). If it is a tuple it means the user overwrote the value to something else and then we use that value.

PR #409, opened 2025-11-05
Use float32 for numpy image and separate image and mask opening

yutong-xiang-97:
## What has changed and why?

During training when NumPy arrays are used, only allow `uint8` and `float32` dtypes for image input dtypes and `integer` for mask input dtypes, so as to accommodate Albumentation's input type specs. As a consequence, we also have to write separate helper functions to open masks.

Major changes:
- Update type specs for `NDArrayImage` and add `NDArrayMask`. Also update other relevant types accordingly.
- Load images with `uint8`, or `float32` if not feasible. The float32 images are standardized to `[0,1]` for Albumentations' transformation to work.
- Load masks with the original dtypes. Masks are always assumed to be of integer types.
- Update corresponding tests for images and masks, incl. 
  - tests with even more dtypes, channels, and image modes.
  - tests for floating range in `[0,1]`
- Use `decode_image` instead of `read_image` wherever possible. `read_image` is to be deprecated in favor of `decode_image` with `torchvision>=0.20.0`; additionally, `decode_image` can also handle `uint16` images. However, str path input for `decode_image` is only possible with `torchvision>=0.20.0` even if it is present in the older versions, so we fall back to `read_image` in this case.
- Finally, remove RGB specs in `open_image_tensor` as they are not necessary.

## How has it been tested?

- Unit tests.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/train_task.py
+++ b/src/lightly_train/_commands/train_task.py
@@ -661,6 +661,7 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
 
         hyperparams = helpers.pretty_format_args_dict(config.model_dump())
         hyperparams.pop("resume_interrupted", None)
+        hyperparams.pop("overwrite", None)
         logger_args = hyperparams.get("logger_args")
         if isinstance(logger_args, dict):
             mlflow_logger_args = logger_args.get("mlflow")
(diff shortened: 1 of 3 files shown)


stegmuel (approved):
LGTM.

PR #410, opened 2025-11-05
Allow pretrained models as model argument

guarin:
## What has changed and why?

* Allow passing pretrained model names and paths as `model="<pretrained name/path>` to the task model train functions
* Rename `load_model_from_checkpoint` to `load_model` for simplicity
* Update docs

## How has it been tested?

* Manually tested the different workflows

[Semantic Segmentation - LightlyTrain documentation - model.pdf](https://github.com/user-attachments/files/23363377/Semantic.Segmentation.-.LightlyTrain.documentation.-.model.pdf)

- [x] Yes

- [x] Yes

### DIFF
--- a/src/lightly_train/_commands/train_task.py
+++ b/src/lightly_train/_commands/train_task.py
@@ -585,10 +585,10 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
             and config.reuse_class_head
             and config.data.included_classes != model_init_args.get("classes")
         ):
-            logger.warning(
+            raise ValueError(
                 f"The included classes in the data configuration ({config.data.included_classes}) "
                 f"do not match the classes used in the checkpoint weights file ({model_init_args.get('classes')}). "
-                f"It is not advisable to reuse the class head when you have a different classes config."
+                f"Set reuse_class_head=False when you have a different classes config."
             )
 
         logger.info(


yutong-xiang-97:
@codex review

src/lightly_train/_commands/train_task.py
```diff
@@ -514,31 +544,15 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
...
     )
 
-    # Load checkpoint context if resuming or further fine-tuning.
-    checkpoint = helpers.load_checkpoint(
-        fabric=fabric,
-        out_dir=out_dir,
-        resume_interrupted=config.resume_interrupted,
-        ckpt_path=config.checkpoint,
+    model_init_args = (
+        {} if checkpoint is None else checkpoint.get("model_init_args", {})
```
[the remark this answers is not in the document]
yutong-xiang-97: This now defaults to load the model specified in `checkpoint`.

src/lightly_train/_commands/train_task.py
```diff
@@ -506,6 +524,18 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
...
+        out_dir=out_dir,
+        resume_interrupted=config.resume_interrupted,
+        model=config.model,
+        checkpoint=config.checkpoint,
+        task=config.task,
+    )
+    if model_name is not None:
+        config.model = model_name
+
     train_model_cls = helpers.get_train_model_cls(
```
yutong-xiang-97, line 537: `get_train_model_cls` is also called in `load_checkpoint` to use only the error message but not the returns. Since the only parameter used in the try-except block is `model` which only affects the logger message in `load_checkpoint`, maybe we can move the block in the main function and pass `model_path` and `model_name` as parameters to `load_checkpoint` instead of `model`? The logger message can be rephrased without `model` I believe.

src/lightly_train/_commands/train_task.py
```diff
@@ -569,13 +583,13 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
         )
 
         if (
-            config.checkpoint
+            checkpoint is not None
             and config.reuse_class_head
             and config.data.included_classes != model_init_args.get("classes")
         ):
-            raise ValueError(
+            logger.warning(
```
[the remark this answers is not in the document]
yutong-xiang-97: Without raising an explicit error, the error messages fall back to the plain size mismatch:

```
[rank0]:        size mismatch for model.class_head.weight: copying a param with shape torch.Size([151, 384]) from checkpoint, the shape in current model is torch.Size([149, 384]).
[rank0]:        size mismatch for model.class_head.bias: copying a param with shape torch.Size([151]) from checkpoint, the shape in current model is torch.Size([149]).
[rank0]:        size mismatch for criterion.empty_weight: copying a param with shape torch.Size([151]) from checkpoint, the shape in current model is torch.Size([149]).
```

I would prefer keeping the explicit error message here as a warning won't keep the training go anyway.

src/lightly_train/_commands/train_task.py
```diff
@@ -681,13 +696,13 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
             model_init_args=train_model.get_task_model().init_args,
         )
 
-        if config.resume_interrupted and checkpoint:
+        if config.resume_interrupted and checkpoint is not None:
```
yutong-xiang-97, line 700: with the new `load_checkpoint`, if `config.resume_interrupted` is `True` then the `checkpoint` must exist. You don't have to use `checkpoint` here.

src/lightly_train/_commands/train_task.py
guarin: This is not completely true, we call `get_train_model_cls` in `load_checkpoint` to figure out if the model name is a valid. If it is an invalid model name we know that it must be a path instead. Once we have the path we can download the weights and figure out the actual model name.

src/lightly_train/_commands/train_task.py
guarin: It is still good to check for typing

src/lightly_train/_commands/train_task.py
yutong-xiang-97: Oh just saw that `config.model` is overwritten. Thanks for the explanation!

PR #415, opened 2025-11-06
Support URLs in predict

guarin:
## What has changed and why?

* Support URLs in predict

This improves UX and makes the solution easier to test. I used fsspec because it handles all types of paths/URLs including s3/gcs/azure buckets.

## How has it been tested?

```python
import lightly_train
model = lightly_train.load_model("dinov3/vits16-eomt-coco")
out = model.predict("https://farm6.staticflickr.com/5466/9514492689_8c327cffb3_z.jpg")
```

- [ ] Yes

- [ ] Yes


yutong-xiang-97 (approved):
Approved.

PR #416, opened 2025-11-06
Add DICOM image loading support

yutong-xiang-97:
## What has changed and why?

Add DICOM image support for single-frame DICOM images.

Major changes:
- DICOM imaged loading as NumPy and Tensor using `pydicom`, including:
  - converting color space from YBR to RGB if necessary
  - decode palette color images to RGB images if necessary
  - rescale image to HU values if necessary
  - converting the non-uint8 images to float32 in range `[0,1]`
- Add `.dcm` to supported extensions
- Add dtype-aware Normalize transform because Albumentations' normalize would still default to use 255 for `max_pixel_value` when float image in `[0,1]` is provided, and tweaking the `max_pixel_value` inside the transform pipeline based on image dtypes is not feasible.
- Update `open_image_tensor` to also support DICOM image loading from URL.
- Add corresponding tests

## How has it been tested?

- Unit tests
- Integration tests (pretraining + semseg, inference is currently omitted)

- [x] Yes

- [ ] Yes
- [x] Not needed (will do a follow-up later)

### DIFF
--- a/src/lightly_train/_data/file_helpers.py
+++ b/src/lightly_train/_data/file_helpers.py
@@ -171,7 +171,6 @@ def open_image_tensor(image_path: PathLike) -> Tensor:
     image: Tensor
 
     suffix = Path(image_path).suffix.lower()
-
     if suffix in _TORCHVISION_SUPPORTED_IMAGE_EXTENSIONS:
         try:
             # Fast path when loading local file with torch.
@@ -205,7 +204,7 @@ def open_image_numpy(
     image_np: NDArrayImage
 
     # Torchvision supported images
-    suffix = Path(image_path).suffix.lower()
+    suffix = image_path.suffix.lower()
     if suffix in _TORCHVISION_SUPPORTED_IMAGE_EXTENSIONS:
         try:
             image_np = _open_image_numpy__with_torch(image_path=image_path, mode=mode)
@@ -252,8 +251,6 @@ def _open_image_numpy__with_torch(
 def _open_image_numpy__with_pydicom(
     image_src: PathLike | BinaryIO,
 ) -> NDArrayImage:
-    from lightning_utilities.core.imports import RequirementCache
-
     if not RequirementCache("pydicom"):
         raise ImportError(
             "pydicom is required to read DICOM images. "


CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Automatically download DINOv3 weights.
 - Add support for passing pretrained model names or checkpoint paths as `model` argument
   to the model training functions like `train_semantic_segmentation`.
+- Add support for loading DICOM images as input data for training and inference.
```
guarin: Add to `[Unreleased]` section

src/lightly_train/_data/file_helpers.py
```diff
@@ -109,22 +111,20 @@ def list_image_filenames_from_dir(image_dir: PathLike) -> Iterable[ImageFilename
         yield ImageFilename(filename)
 
 
-def _pil_supported_image_extensions() -> set[str]:
+def _supported_image_extensions() -> set[str]:
```
guarin: Please leave as two separate functions, you can call `_pil_supported_image_extensions` from within `_supported_image_extensions`. Otherwise we don't have a function anymore to get just the PIL supported image extensions.

src/lightly_train/_data/file_helpers.py
```diff
@@ -157,7 +157,9 @@ def open_image_tensor(image_path: PathLike) -> Tensor:
         image_path: Path to the image file. Can be a local path or URL.
     """
     image: Tensor
-    if Path(image_path).suffix.lower() in _TORCHVISION_SUPPORTED_IMAGE_EXTENSIONS:
+    if (
+        suffix := Path(image_path).suffix.lower()
```
guarin: The value assigned by the walrus operator should only be used in the same code section. Otherwise the code flow gets hard to track.

```
if suffix := ...:
   # use suffix here
elif suffix == : # don't use suffix here
  ... 
else:
   # don't use suffix here
```

In these cases it is clearer if you do:
```
suffix = ...
if suffix == ...:
elif suffix == ...: 
```

src/lightly_train/_data/file_helpers.py
```diff
@@ -166,6 +168,15 @@ def open_image_tensor(image_path: PathLike) -> Tensor:
             pass
         else:
             return image
+    elif suffix == ".dcm":
+        image_np = _open_image_numpy__with_pydicom(image_path=image_path)
```
guarin: This won't work as expected. The current logic of the function is:
1. Try to read image with torch
  * This should only fail if the image path is not supported by torch, for example if it is a URL 
3. If above fails, try to read image with fsspec and PIL
  * This handles any image type and also urls/s3/gcs/azure paths

With the current implementation loading from urls won't work with dicom

src/lightly_train/_transforms/normalize.py
```diff
@@ -0,0 +1,146 @@
...
+    def _apply_geq_1_4_4(self, img: NDArrayImage, **params: Any) -> NDArrayImage:
+        if self.normalization == "standard":
+            if img.dtype == np.float32:
+                # float32 input is assumed to be in [0.0, 1.0]
+                self.mean_np = np.array(self.mean, dtype=np.float32)
+                self.denominator = np.reciprocal(
+                    np.array(self.std, dtype=np.float32),
+                )
+            else:
+                # uint8 input is assumed to be in [0, 255] unless max_pixel_value is explicitly set
```
guarin: Let's check here that it is indeed uint8 and throw an error otherwise.

src/lightly_train/_transforms/normalize.py
```diff
@@ -0,0 +1,146 @@
...
+                std=std,
+                max_pixel_value=max_pixel_value,
+                p=p,
+            )
+
+    def apply(self, img: NDArrayImage, **params: Any) -> NDArrayImage:
+        if ALBUMENTATIONS_GEQ_1_4_4:
+            return self._apply_geq_1_4_4(img, **params)
+        else:
+            return self._apply_lt_1_4_4(img, **params)
```
guarin: The logic seems rather complicated, would the following also work? 
```
    def apply(self, img: NDArrayImage, **params: Any) -> NDArrayImage:
        if img.dtype == np.float32:
            # Image is in [0, 1] but albumentations expects [0, self.max_pixel_value]
            img = img * self.max_pixel_value
        return super().apply(img, **params)

File: tests/_commands/test_train.py [hunk 7]
```diff
@@ -546,3 +551,43 @@ def test_train__multichannel(
...
+    tmp_path: Path,
+    module_attribute: str,
+    num_channels: int,
+) -> None:
+    pydicom_examples = pytest.importorskip(
+        "pydicom.examples",
+        reason="pydicom examples not supported",
+    )
+    get_path = getattr(pydicom_examples, "get_path")
+    data_path: Path = get_path(module_attribute)
```
guarin:
```suggestion
    data_path: Path = pydicom_examples.get_path(module_attribute)
```

tests/_commands/test_train.py
```diff
@@ -546,3 +551,43 @@ def test_train__multichannel(
...
+        ("mr", 1),
+        ("overlay", 1),
+        ("rgb_color", 3),
+        ("palette_color", 3),
+        ("jpeg2k", 3),
+    ],
+)
+def test_train__dicom(
+    tmp_path: Path,
+    module_attribute: str,
```
guarin: Why is this called `module_attribute`? Looks more like a data format to me

src/lightly_train/_transforms/normalize.py
yutong-xiang-97: Yep it works w/o the need to deal with `albucore`. Thanks for the suggestion!

src/lightly_train/_data/file_helpers.py
yutong-xiang-97: Oh I indeed didn't consider supporting this. Proposed a fix.

yutong-xiang-97:
@codex review

src/lightly_train/_data/file_helpers.py
```diff
@@ -179,12 +203,20 @@ def open_image_numpy(
 ) -> NDArrayImage:
     """Returns image as (H, W, C) or (H, W) numpy array."""
     image_np: NDArrayImage
-    if image_path.suffix.lower() in _TORCHVISION_SUPPORTED_IMAGE_EXTENSIONS:
+
+    # Torchvision supported images
+    suffix = Path(image_path).suffix.lower()
```
guarin:
```suggestion
    suffix = image_path.suffix.lower()
```

src/lightly_train/_data/file_helpers.py
```diff
@@ -217,6 +249,76 @@ def _open_image_numpy__with_torch(
     return image_np
 
 
+def _open_image_numpy__with_pydicom(
+    image_src: PathLike | BinaryIO,
+) -> NDArrayImage:
+    from lightning_utilities.core.imports import RequirementCache
```
guarin: Move to top of file

tests/_data/test_file_helpers.py
```diff
@@ -202,18 +235,104 @@ def test_open_image_numpy(
...
+            "pydicom.examples",
+            reason="pydicom examples not supported",
+        )
+        image_path: Path = pydicom_examples.get_path("mr")
+    else:
+        image_path = tmp_path / f"image{extension}"
+
+        max_value = int(np.iinfo(dtype).max) if np.issubdtype(dtype, np.integer) else 1
+        helpers.create_image(
+            path=image_path,
```
guarin, line 249: These tests are getting too complicated, we should split them once we have time

yutong-xiang-97: Yeah may be one test per backend.

PR #430, opened 2025-11-12
Thomas trn 1628 small od models

stegmuel:
## What has changed and why?

Add ViT-S/16, ViT+/16 and ViT/16 models for object detection.

## How has it been tested?

Manual tests (COCO eval, finetuning on PVOC).

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
+++ b/src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
@@ -139,8 +139,12 @@ class DINOv2LTDETRObjectDetectionTrainTransformArgs(ObjectDetectionTransformArgs
                 field.resolve_auto(height=height, width=width)
 
         if self.normalize == "auto":
-            normalize = model_init_args.get("image_normalize")
+            normalize = model_init_args.get("image_normalize", "none")
+            # Normalize is specifically set to None in model_init_args.
             if normalize is None:
+                self.normalize = None
+            # Normalize is not set in model_init_args.
+            elif normalize == "none":
                 self.normalize = NormalizeArgs()
             else:
                 assert isinstance(normalize, dict)
@@ -194,8 +198,12 @@ class DINOv2LTDETRObjectDetectionValTransformArgs(ObjectDetectionTransformArgs):
                 field.resolve_auto(height=height, width=width)
 
         if self.normalize == "auto":
-            normalize = model_init_args.get("image_normalize")
+            normalize = model_init_args.get("image_normalize", "none")
+            # Normalize is specifically set to None in model_init_args.
             if normalize is None:
+                self.normalize = None
+            # Normalize is not set in model_init_args.
+            elif normalize == "none":
                 self.normalize = NormalizeArgs()
             else:
                 assert isinstance(normalize, dict)
--- a/src/lightly_train/_task_models/dinov3_ltdetr_object_detection/transforms.py
+++ b/src/lightly_train/_task_models/dinov3_ltdetr_object_detection/transforms.py
@@ -139,8 +139,12 @@ class DINOv3LTDETRObjectDetectionTrainTransformArgs(ObjectDetectionTransformArgs
                 field.resolve_auto(height=height, width=width)
 
         if self.normalize == "auto":
-            normalize = model_init_args.get("image_normalize")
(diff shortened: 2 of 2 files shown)


src/lightly_train/_task_models/dinov2_ltdetr_object_detection/task_model.py
```diff
@@ -213,9 +207,13 @@ def predict(
         h, w = x.shape[-2:]
 
         x = transforms_functional.to_dtype(x, dtype=torch.float32)
+
```
guarin, line 210: I believe you have to set `scale=True` in `to_dtype` to rescale to [0,1] before normalization

src/lightly_train/_task_models/task_model_helpers.py
```diff
@@ -40,6 +40,18 @@
         "dinov2_ltdetr_2/ltdetr_vits14dinov2_coco_dsp.pt",
         "7e1f91b251ba0b796d88fb68276a24a52341aa6e8fb40abe9f730c2a093a5b40",
     ),
+    "dinov3/vitt16-ltdetr-coco": (
+        "dinov3/dinov3_vitt16_ltdetr_coco.pt",
+        "1a4c20a114bf202f5f68d771b9f90276be8bb9c8ffc180b8721a11cbad5578ad",
+    ),
```
guarin, line 46: My goal is that in the future we can follow the checkpoint naming pattern from the dinov3 convnext ltdetr models:
```
<package>_<model>_<date>_<hash>.pt
```
This way we'll be able to skip adding the hash as an extra entry and we won't need different subdirectory and version naming schemes. 

Doesn't have to be changed now, just FYI

tests/_data/test_yolo_object_detection_dataset.py
```diff
@@ -38,7 +39,7 @@
 
 class DummyTransformArgs(ObjectDetectionTransformArgs):
     channel_drop: ChannelDropArgs | None = None
-    num_channels: int | Literal["auto"] = "auto"
+    num_channels: int | Literal["auto"] = 3
```
guarin, line 42: Why did you have to change this?

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/transforms.py
```diff
@@ -141,6 +138,23 @@ def resolve_auto(self, model_init_args: dict[str, Any]) -> None:
             if hasattr(field, "resolve_auto"):
                 field.resolve_auto(height=height, width=width)
 
+        if self.normalize == "auto":
+            normalize = model_init_args.get("image_normalize")
+            if normalize is None:
+                self.normalize = None
```
guarin: This means that we do not normalize by default. Is this correct and intended? IMO that is fine, just want to double check.

tests/_data/test_yolo_object_detection_dataset.py
stegmuel: Because `ObjectDetectionTransformArgs` does not implement `resolve_auto`, only the sub-classes do.
But I thinking I'm missing something because it used to work and  `resolve_auto` was also not implemented.

src/lightly_train/_task_models/task_model_helpers.py
stegmuel: Updated.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
```diff
@@ -141,7 +141,7 @@ def resolve_auto(self, model_init_args: dict[str, Any]) -> None:
         if self.normalize == "auto":
             normalize = model_init_args.get("image_normalize")
             if normalize is None:
-                self.normalize = None
+                self.normalize = NormalizeArgs()
```
guarin, line 148: I think we have to make a finer-grained check here. There are 3 scenarios:
1. `model_init_args` is empty or "image_normalize" is not set
2. `model_init_args` is not empty and "image_normalize" is set to `None`
3. `model_init_args` is not empty and "image_normalize" is a dict

And we would want the following outcome:
1. Default to `NormalizeArgs()` or `None` --> You decide what makes more sense
2. `normalize = None`
3. `normalize = NormalizeArgs.from_dict(...)`

guarin (approved):

PR #432, opened 2025-11-12
Document Non-RGB & DICOM loading support

yutong-xiang-97:
## What has changed and why?

As title suggests.

- Added a "Data Input" section containing two subpages for now:
  - Non-RGB images (for multi-channel support)
  - DICOM images
- Link this section to necessary pages
- Remove relevant content from the respective pages

## How has it been tested?

[Data Input - LightlyTrain documentation.pdf](https://github.com/user-attachments/files/23522387/Data.Input.-.LightlyTrain.documentation.pdf)
[Single- and Multi-Channel Images - LightlyTrain documentation.pdf](https://github.com/user-attachments/files/23522416/Single-.and.Multi-Channel.Images.-.LightlyTrain.documentation.pdf)
[DICOM Images - LightlyTrain documentation.pdf](https://github.com/user-attachments/files/23522388/DICOM.Images.-.LightlyTrain.documentation.pdf)

[Train - LightlyTrain documentation.pdf](https://github.com/user-attachments/files/23505305/Train.-.LightlyTrain.documentation.pdf)
[Object Detection - LightlyTrain documentation.pdf](https://github.com/user-attachments/files/23505299/Object.Detection.-.LightlyTrain.documentation.pdf)
[Semantic Segmentation - LightlyTrain documentation.pdf](https://github.com/user-attachments/files/23505304/Semantic.Segmentation.-.LightlyTrain.documentation.pdf)

- [ ] Yes

- [x] Yes


docs/source/data/dicom.md
```diff
@@ -0,0 +1,42 @@
...
+
+To install LightlyTrain with PyDICOM support, do:
+
+```bash
+pip install lightly-train[dicom]
+```
+
+For DICOM images that may require additional processing, LightlyTrain automatically applies the following using `pydicom` functions:
+
+- converting color space from `YBR` to `RGB` via [`convert_color_space`](https://pydicom.github.io/pydicom/stable/reference/generated/pydicom.pixels.convert_color_space.html)
```
guarin, line 23: Can the conversions be enabled/disabled?

docs/source/data/index.md
```diff
@@ -0,0 +1,26 @@
+(data-input)=
+
+# Data Input
+
+LightlyTrain supports images beyond standard RGB. For domain-specific use cases, such as medical imaging, it also loads images from specialized formats.
```
guarin:
```suggestion

LightlyTrain supports standard RGB images and images for domain-specific use cases, such as medical imaging. It also loads images from specialized formats.
```

Let's add a list below with all the image formats and link to the respective pages if applicable.

docs/source/data/non_rgb.md
```diff
@@ -0,0 +1,62 @@
+(non-rgb)=
+
+# Non-RGB Images
```
guarin: Would name the page `Single- and Multi-Channel Images`. Non-RGB can mean other things like BGR or other color formats.

docs/source/data/non_rgb.md
```diff
@@ -0,0 +1,62 @@
...
+    data={
+        ... # multi-channel image data (e.g. RGB-NIR)
+    },
+    logger_args={
+        "mlflow": {
+            "experiment_name": "my_experiment",
+            "run_name": "my_run",
+            "tracking_uri": "tracking_uri",
+        },
+    },
```
guarin: Remove mlflow

docs/source/data/non_rgb.md
```diff
@@ -0,0 +1,62 @@
...
+    transform_args={
+        "num_channels": 4 # specify number of channels here
+    },
+)
+```
+
+```{note}
+You may also want to customize normalization with the `normalize` parameter in `transform_args`. If you don't set it, LightlyTrain repeats the ImageNet default RGB `mean` and `std` values for any additional channels.
```
guarin: Show this directly in the code snippet

guarin (changes requested):

docs/source/data/dicom.md
yutong-xiang-97: We don't disable on our end, but we will determine what needs to be converted and apply the changes if needed.

guarin: Ok, I was just thinking that in some scenarios it makes sense to load the images as is and not convert to RGB. We don't really care about the color format. Maybe we can make the conversion optional in the future and ideally move it into the transforms.

docs/source/data/dicom.md
```diff
@@ -0,0 +1,72 @@
...
+```
+
+Also be aware that some transform options may be inappropriate depending on the acquisition protocol. For example, vertical flips in `RandomFlip` may not be suitable for certain medical images. To disable vertical flips while keeping horizontal flips:
+
+```python
+transform_args={
+    "random_flip": {
+        "vertical_prob": 0.0,
+    },
+},
```
guarin: Would remove this part as vertical flips are anyways disabled by default.

docs/source/data/multi_channel.md
```diff
@@ -0,0 +1,55 @@
+(multi-channel)=
+
+# Single- and Multi-Channel Images
+
+In addition to standard RGB images, LightlyTrain supports single- and multi-channel inputs for pretraining, distillation, and fine-tuning.
```
guarin: It is not supported for distillation because the DINOv2/v3 teacher models expect 3-channel inputs. What is supported is loading n-channel images and then reducing them to 3-channels with the channel drop augmentation.

docs/source/data/dicom.md
yutong-xiang-97: Was just using it as an example. I may come up with another one to replace it.

PR #438, opened 2025-11-13
Disable unit test tracker

guarin:
## What has changed and why?

## How has it been tested?

- [ ] Yes

- [ ] Yes

### DIFF
--- a/docs/source/data/dicom.md
+++ b/docs/source/data/dicom.md
@@ -1,0 +1,72 @@
+(dicom-support)=
+
+# DICOM Images
+
+LightlyTrain supports loading [DICOM](https://www.dicomstandard.org/) images for pretraining, distillation, and fine-tuning.
+
+```{note}
+Currently, we do not support loading DICOM images as segmentation masks.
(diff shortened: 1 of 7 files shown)


yutong-xiang-97:
@codex review

yutong-xiang-97:
Does the `LIGHTLY_TRAIN_POSTHOG_KEY` has to be empty in this case by design? If so, there could be issues because `EnvVar.convert_empty_str_to_default` is automatically set to `True`, so the following test won't pass:

```
def test_events_disabled_env_for_tests():
    # Should be True because the fixture sets "1"
    assert Env.LIGHTLY_TRAIN_EVENTS_DISABLED.value is True
    assert Env.LIGHTLY_TRAIN_POSTHOG_KEY.value == ""
```

You may want to set `convert_empty_str_to_default=False` for `LIGHTLY_TRAIN_POSTHOG_KEY` unless only the `LIGHTLY_TRAIN_EVENTS_DISABLED` needs to function in this case.

PR #448, opened 2025-11-17
Fix resume interrupted for fine-tuning

### LINKED ISSUE #428: Resume training for my segment task
I want to resume a previously interrupted training run, and here is my running code: 
```
import lightly_train
if __name__ == "__main__":
    lightly_train.train_semantic_segmentation(
        out="out/my_experiment",
        resume_interrupted=True,
        model="dinov2/vits14-eomt", 
        data={
                   ……
        },
    )
```
The original mIoU for 30000 steps was 0.7012, but when I ran this code, it showed:
```
Resuming training from step 30000/40000...
Train Step 30100/40000 | Train Loss: 29.6862 | Train mIoU: 0.0289
```
Train mIoU is close to 0. I want to know how to resume training correctly. Thanks.
(I see words in document: “Like in pretraining, there are two distinct ways to continue training, depending on your intention. Therefore, the resume_interrupted=True parameter cannot be combined with passing a checkpoint path to the model parameter.” However, when I commented out the statement `model="dinov2/vits14-eomt", ` it gave an error：
```
lightly_train.train_semantic_segmentation(
TypeError: train_semantic_segmentation() missing 1 required keyword-only argument: 'model'
```)

### PR DESCRIPTION
**Pull request by guarin**
## What has changed and why?

* Fix resume interrupted for fine-tuning

The problem was that we loaded the checkpoint with `fabric.load` and then assigned `train_dataloader = ckpt["train_dataloader"]` but this doesn't work correctly as fabric has to handle the state loading completely. This also solves issues where the train dataloader would load images on the wrong device.

Closes #428

## How has it been tested?

* Manually

After resuming I get the expected train mIoU:
```
Saving checkpoints every 1000 steps.[0m
Train Step 3100/10000 | Train Loss: 14.1718 | Train mIoU: 0.3722[0m

```

- [x] Yes

- [ ] Yes

### DIFF
--- a/tests/_commands/test_train_task.py
+++ b/tests/_commands/test_train_task.py
@@ -464,5 +464,5 @@ def test_train_semantic_segmentation__resume_interrupted(
             resume_interrupted=True,
         )
 
-    assert f"Loading model checkpoint from '{last_ckpt_path}'" in caplog.text
+    assert f"Resuming training from model checkpoint '{last_ckpt_path}'" in caplog.text
     assert "Resuming training from step 1/1..." in caplog.text

### CODE REVIEW

File: src/lightly_train/_commands/train_task_helpers.py [hunk 1]
```diff
@@ -863,77 +870,30 @@ def load_checkpoint(
...
-                step=step,
-                model_class_path=model_class_path,
-                model_init_args=model_init_args,
-            ),
-            model_name,
-        )
-
     return (
-        ExportedCheckpoint(
+        CheckpointDict(
```
yutong-xiang-97, line 874: Since we, in any case, load these info from `ckpt_path` and the returned `checkpoint` is not used elsewhere in `train_task.py` before we actually use the loaded ckpt, why not simply change this function to return `ckpt_path` and `model_name` and push the model loading with `fabric` back to `resume_from_checkpoint` and `finetune_from_checkpoint`? This function is doing too many things imo.

src/lightly_train/_commands/train_task.py
```diff
@@ -697,16 +697,16 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
             model_init_args=train_model.get_task_model().init_args,
         )
 
-        if config.resume_interrupted and checkpoint is not None:
+        if config.resume_interrupted and checkpoint_path is not None:
```
yutong-xiang-97, line 725: Also here with the changes below like:
```
    if resume_interrupted:
        ...
        ckpt_path = get_checkpoint_path(out_dir, best_or_last="last")
        ...
        return (
            None,
            ckpt_path,
            model_name,
        )
```

`config.resume_interrupted` would imply `checkpoint_path is not None` already. `checkpoint=None` would also hold in this case.

Overall, I think the if conditions we use in `train_task.py` to control the ckpt loading logic can be optimized.

src/lightly_train/_commands/train_task_helpers.py
guarin: We have to load the checkpoint before `finetune_from_checkpoint` to figure out `model_name` and `model_init_args`.

guarin:
> The issue only appears after [this PR](https://github.com/lightly-ai/lightly-train/pull/410/files#diff-e0e2a8313cdb9012a869fd373193512cd4307c37d6df9eea65191513af9228da) with changes to how to load the train_dataloader.

I tried and it still fails on multi-GPU training. It seems that code never worked correctly because `fabric.load` wasn't used.

src/lightly_train/_commands/train_task.py
guarin: I agree that it could be improved but I would like to keep the working code as is.

src/lightly_train/_commands/train_task_helpers.py
yutong-xiang-97: I would personally prefer that we only return what is used (in this case `ckpt_path`, `model_name`, and `model_init_args`) to keep things clean, but yes we could proceed with the working code as-is, and it has also the benefit of loading the checkpoint only once. I just want to say that something might have been poorly designed that have made our whole checkpoint loading logic so complex and error-prone. We might still want to look into this later on.

guarin: The checkpoint weights are also used and passed to `finetune_from_checkpoint`.

PR #466, opened 2025-11-24
Deprecate reuse class head and use hooks for segmentation

yutong-xiang-97:
## What has changed and why?

Deprecate the `reuse_class_head` argument in `train_task.py` because 

- it is not suitable for handling object detection (so this PR is a precursor of fixing OD class-dependent weight loading issues)
- the logic is supposed to be handled by `load_state_dict_pre_hooks`

Instead we add two hooks:

- `class_head_reuse_or_reinit_hook` in `TaskModel`: only load the `class_head` weights from the state dict if the `num_classes` match; otherwise, the `class_head` is re-inited.
- `criterion_empty_weight_reinit_hook` in `TrainModel`: Reinit the `criterion.empty_weight` deterministically based on the new `num_classes`.

NOT included: `dinov2-linear-semantic-segmentation` cuz it is not publicized in the docs.

## How has it been tested?

Manually

- [x] Yes

- [x] Yes

### DIFF
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -903,4 +903,17 @@ def finetune_from_checkpoint(
     """
 
     train_model = state["train_model"]
-    train_model.load_state_dict(checkpoint["train_model_state_dict"], strict=True)
+    incompatible = train_model.load_state_dict(
+        checkpoint["train_model_state_dict"], strict=False
+    )
+
+    if incompatible.missing_keys:
+        logger.warning(
+            "Missing keys after loading checkpoint: %s",
+            incompatible.missing_keys,
+        )
+    if incompatible.unexpected_keys:
+        logger.warning(
+            "Unexpected keys after loading checkpoint: %s",
+            incompatible.unexpected_keys,
+        )
--- a/src/lightly_train/_task_models/task_model_helpers.py
+++ b/src/lightly_train/_task_models/task_model_helpers.py
@@ -306,7 +306,7 @@ def class_head_reuse_or_reinit_hook(
         return
     else:
         logger.info(
-            f"Checkpoint provides {num_classes_state} classes but module expects {num_classes_module}. Reinitializing class head.",
+            f"Checkpoint provides {num_classes_state - 1} classes but module expects {num_classes_module - 1}. Reinitializing class head.",
         )
 
         # Re-initialize class head weights and biases
--- a/tests/_commands/test_train_task.py
+++ b/tests/_commands/test_train_task.py
@@ -380,7 +380,7 @@ def test_train_semantic_segmentation__checkpoint(
             steps=1,
             overwrite=True,
         )
-    assert "Checkpoint provides 3 classes but module expects 4." in caplog.text
+    assert "Checkpoint provides 2 classes but module expects 3." in caplog.text
 
 
 @pytest.mark.skipif(


src/lightly_train/_task_models/task_model_helpers.py
```diff
@@ -280,3 +281,33 @@ def queries_adjust_num_queries_hook(
...
+    num_classes_module = class_head_module.out_features
+    if num_classes_state == num_classes_module:
+        return
+    else:
+        logger.info(
+            f"Checkpoint provides {num_classes_state} classes but module expects {num_classes_module}. Reinitializing class head.",
+        )
+        # Remove class head weights/bias to force reinitialization.
+        del state_dict[class_head_weight_key]
+        del state_dict[class_head_bias_key]
```
guarin: Doesn't removing the entries from the state dict trigger an error that the weights are missing from the state dict?

tests/_commands/test_train_task.py
```diff
@@ -353,38 +353,6 @@ def test_train_semantic_segmentation__checkpoint(
...
-            accelerator="auto" if not sys.platform.startswith("darwin") else "cpu",
-            devices=1,
-            batch_size=2,
-            num_workers=0,
-            steps=1,
-            overwrite=True,
-            checkpoint=last_ckpt_path,
-            reuse_class_head=False,
-        )
-    assert "Skipping class-dependent parameters from checkpoint:" in caplog.text
```
guarin, line 386: Shouldn't we keep this test but check for the new "Reinitializing class head." message in the logs instead?

guarin (changes requested):

tests/_commands/test_train_task.py
yutong-xiang-97: Then we have to create a ViT checkpoint fixture or use an existing publicized checkpoint for this, which might increase memory consumption. Do you think it is absolutely needed?

src/lightly_train/_task_models/task_model_helpers.py
yutong-xiang-97: The `finetune_from_checkpoint` function uses `strict=False` which returns the incompatible keys as a warning: https://github.com/lightly-ai/lightly-train/pull/466/files#diff-f327769a1fc8881e2ac97f636a31f94897d71f89208ce2e2ab2779cf38a95646R906

guarin: Ok but in that case we don't want the head weights to raise this warning no? Otherwise it might confuse users. I would prefer to re-initialize here and log an info message.

tests/_commands/test_train_task.py
guarin: I don't understand. The first run creates a checkpoint with 2 classes. The second run re-uses that checkpoint with the same number of classes. And the third run loads the checkpoint with `checkpoint=last_ckpt_path` and 3 classes (here the logging should happen).

yutong-xiang-97: Sorry completely unaware that step one is to generate the checkpoint. Will add the test.

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -903,17 +903,4 @@ def finetune_from_checkpoint(
...
-    if incompatible.missing_keys:
-        logger.warning(
-            "Missing keys after loading checkpoint: %s",
-            incompatible.missing_keys,
-        )
-    if incompatible.unexpected_keys:
-        logger.warning(
-            "Unexpected keys after loading checkpoint: %s",
-            incompatible.unexpected_keys,
-        )
```
guarin: I would still keep these checks and warnings. This lets users fine-tune even if for some reason or another something is wrong with the checkpoint. My point before was that when the number of classes changes we know what has to be done and can fix it accordingly and give the user a better message.

tests/_commands/test_train_task.py
```diff
@@ -341,18 +341,47 @@ def test_train_semantic_segmentation__checkpoint(
...
+            model=str(last_ckpt_path),
+            model_args={"num_joint_blocks": 1},
+            accelerator="auto" if not sys.platform.startswith("darwin") else "cpu",
+            devices=1,
+            batch_size=2,
+            num_workers=0,
+            steps=1,
+            overwrite=True,
+        )
+    assert "Checkpoint provides 3 classes but module expects 4." in caplog.text
```
guarin: This message is now a bit awkward because the user sets 3 classes in the config and the message complains that there are 3 classes in the model but 4 are expected. --> Should we just remove -1 from the class counts in the message?

src/lightly_train/_commands/train_task_helpers.py
yutong-xiang-97: I am 100% sure if I get your message. Do you mean you still want `strict=False` and keep the checks for missing and incompatible keys, while we only use an info log for the mismatched `class_head` weights?

guarin: Yes exactly!

guarin (approved):
LGTM

PR #467, opened 2025-11-24
Fix OD `num_classes` issue.

yutong-xiang-97:
## What has changed and why?

The object detection decoder and postprocessor are not initialized with the actual `num_classes` in the dataset, but use the default. This has resulted in issues like label mismatch when using the model for inference and improper weight loading.

This PR passes the correct `num_classes` in the decoder and postprocessor while handling the weight loading by re-init class-dependent parameters once `num_classes` do not match.

## How has it been tested?

Manually.

- [x] Yes

- [ ] Yes

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
   <img width="1373" height="322" alt="Book meeting with Lightly at RSNA and NeurIPS 2025" src="https://github.com/user-attachments/assets/5e9177e7-b872-46f1-8b9f-eaadd379d289"/>
   </a>
 </p>
+
 # LightlyTrain - SOTA Pretraining, Fine-tuning and Distillation
 
 [![Python](https://img.shields.io/badge/Python-3.8%7C3.9%7C3.10%7C3.11%7C3.12-blue.svg)](https://docs.lightly.ai/train/stable/installation.html)


src/lightly_train/_task_models/object_detection_components/rtdetrv2_decoder.py
```diff
@@ -484,6 +486,25 @@ def __init__(
             self.register_buffer("anchors", anchors, persistent=False)
             self.register_buffer("valid_mask", valid_mask, persistent=False)
 
+        if hasattr(self, "register_load_state_dict_pre_hook"):
```
stegmuel, line 493: Can you add a note (either at the top or in notice.md) about the changes to this file?

src/lightly_train/_task_models/task_model_helpers.py
```diff
@@ -309,6 +340,85 @@ def class_head_reuse_or_reinit_hook(
...
+    state_dict: dict[str, Any],
+    *,
+    weight_key: str,
+    bias_key: str,
+) -> bool:
+    score_head_weight = state_dict.get(weight_key)
+    if score_head_weight is None:
+        return False
+
+    num_classes_state = score_head_weight.shape[0]
```
stegmuel, line 414: What about the case where [`query_select_method==agnostic`](https://github.com/lightly-ai/lightly-train/blob/1d8aff2c1671f82dafe20e416e713f546c3a836a/src/lightly_train/_task_models/object_detection_components/rtdetrv2_decoder.py#L465)? Should we leave a TODO and/or a warning?

stegmuel (commented):
reviewed

src/lightly_train/_task_models/task_model_helpers.py
yutong-xiang-97: In this case, the dimension of `enc_score_head` is independent of `num_classes`. With the current impl, we simply load the weights when the dimensions match, and that's consistent with the behavior of other hooks. I think it is fine to load the weights in this case, unless it does not make sense at all or affects the results significantly. 

I don't know the details of how the `score_head` and `query_select_method` work. Should the head be in any case re-inited theoretically speaking when we train on a different dataset?

stegmuel: Let's keep it like this for the moment as it is not the default setting and unlikely to be used.

PR #473, opened 2025-11-26
Fix bug in object detection model.predict

guarin:
## What has changed and why?

* Fix bug in object detection model.predict

Fixes bug introduced in #462

## How has it been tested?

* Manually

Will add unit tests as soon as tiny models are out

- [x] Yes

- [ ] Yes


stegmuel (approved):
LGTM

PR #475, opened 2025-11-27
add log of best result

masakljun:
This PR adds logging of the final best metric at the end of training. The value is written right before the `Training completed `message so the user can see which score was achieved.

I verified the change by running `pytest tests/_commands/test_train_task.py::test_train_semantic_segmentation__checkpoint --log-cli-level=INFO -s`. The logs show that the best result is recorded correctly.
Screenshot 2025-11-27 at 14 32 36

- [ ] Yes

- [ ] Yes

### DIFF
--- a/mlruns/0/meta.yaml
+++ b/mlruns/0/meta.yaml
@@ -1,6 +1,0 @@
-artifact_location: file:///Users/masakljun/Documents/git/lightly-train/mlruns/0
-creation_time: 1764235413456
-experiment_id: '0'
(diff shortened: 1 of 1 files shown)


mlruns/0/meta.yaml
guarin, line 1: Looks like this file was accidentally committed :)

masakljun: Oh thanks, I see, removed now

PR #484, opened 2025-12-04
Masa trn 1695 load ema weights only

masakljun:
## What has changed and why?

This PR enables the training pipeline to initialize/fine-tune a model using checkpoints that only contain EMA (Exponential Moving Average) weights. 

Key Changes
- Added `load_train_state_dict()` to the base `TrainModel` class. This defaults to the standard `load_state_dict()` behavior for most models.
- Updated `finetune_from_checkpoint()` in `train_task_helpers.py` to use this new method.

## How has it been tested?

- Will create a test for this TODO

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/_warnings.py
+++ b/src/lightly_train/_commands/_warnings.py
@@ -66,6 +66,12 @@ def filter_train_warnings() -> None:
         category=UserWarning,
         module="torch.optim.lr_scheduler",
     )
+    # PyTorch Lightning warning at beginning of distillation because teacher is in
+    # eval mode.
+    warnings.filterwarnings(
+        "ignore",
+        message=r"Found .* in eval mode at the start of training",
+    )
 
 
 def filter_embed_warnings() -> None:


src/lightly_train/_task_models/task_model.py
```diff
@@ -82,6 +82,8 @@ def predict(self, image: PathLike | PILImage | Tensor) -> Any:
         """
         raise NotImplementedError()
 
-    def load_train_state_dict(self, state_dict: dict[str, Any]) -> None:
+    def load_train_state_dict(
+        self, state_dict: dict[str, Any], strict: bool = True, assign: bool = False
```
stegmuel: Was it really needed to change this method? And subsequently that of all subclasses?

src/lightly_train/_task_models/train_model.py
```diff
@@ -87,6 +87,16 @@ def get_task_model(self) -> TaskModel:
         """
         raise NotImplementedError()
 
+    def load_train_state_dict(
+        self, state_dict: dict[str, Any], strict: bool = True, assign: bool = False
+    ) -> Any:
+        """Loads the model state dict.
+
+        This acts as a wrapper around load_state_dict. Subclasses can overload
```
stegmuel, line 95: NIT: you can find an example of a docstring with our formatting guideline [here](https://github.com/lightly-ai/lightly-train/blob/bce3de6531fe95cbf80b50d9372101e1fd9a44db/src/lightly_train/_commands/train_task_helpers.py#L779)

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -175,6 +175,18 @@ def __init__(
         self.map_metric = MeanAveragePrecision()
         self.map_metric.warn_on_many_detections = False
 
+    def load_train_state_dict(
+        self, state_dict: dict[str, Any], strict: bool = True, assign: bool = False
+    ) -> Any:
+        """Loads the model state dict.
+
+        Overloads the default implementation to use the task model's loading logic.
+        This allows loading weights from an EMA model into the training model.
```
stegmuel, line 192: We should probably also load the weights inside the EMA model.

src/lightly_train/_task_models/train_model.py
guarin: @stegmuel what is wrong with the formatting? For functions that are only really used internally I don't expected args/returns to be documented in the docstring.

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -935,8 +935,10 @@ def finetune_from_checkpoint(
         checkpoint: Checkpoint context the state dicts to load.
     """
 
-    train_model = state["train_model"]
-    incompatible = train_model.load_state_dict(
+    # train_model = state["train_model"]
```
stegmuel: Reminder to remove the comment.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -183,6 +183,35 @@ def __init__(
...
+        """
+        # Load the student using task-model logic
```
stegmuel: Not sure if student is appropriate in this context.
Proposition: Initialize the model using the task-model logic.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
```diff
@@ -461,14 +461,16 @@ def list_model_names(cls) -> list[str]:
             if "convnext" in name
         ]
 
-    def load_train_state_dict(self, state_dict: dict[str, Any]) -> None:
+    def load_train_state_dict(
+        self, state_dict: dict[str, Any], strict: bool = True, assign: bool = False
+    ) -> Any:
         """Load the EMA state dict from a training checkpoint."""
         new_state_dict = {}
```
stegmuel, line 472: I realize that we have also overlooked the case were the checkpoint is obtained with a training that does not use ema. This implies that we should check if the state dict contains keys starting with `ema_model.model`

stegmuel: Can be ignored.

stegmuel: Potentially you can leave a TODO stating that the currrent code won't work in the situation I mentioned.

tests/_task_models/test_train_model.py
```diff
@@ -0,0 +1,49 @@
...
+
+from lightly_train._task_models.dinov3_ltdetr_object_detection.task_model import (
+    DINOv3LTDETRObjectDetection,
+)
+
+
+def test_dinov3_ltdetr_load_train_state_dict_uses_ema_keys_and_forwards_kwargs() -> (
+    None
+):
+    model = object.__new__(DINOv3LTDETRObjectDetection)
```
stegmuel: Here you should instantiate a `DINOv3LTDETRObjectDetectionTrain` instead:
`train_model = object.__new__(DINOv3LTDETRObjectDetection)`

tests/_task_models/test_train_model.py
```diff
@@ -0,0 +1,49 @@
...
+
+    captured: dict[str, Any] = {}
+
```
stegmuel: Then you can create a fake instance task model class:
```
class FakeTaskModel:
    def load_train_state_dict(self, sd, strict=True, assign=False):
        captured["inner_sd"] = sd
        captured["inner_strict"] = strict
        captured["inner_assign"] = assign
        return "incompatible_keys"
```

You can then attach the fake task model to the train model:
`train_model.model = FakeTaskModel()`

And later call the real `train_model.load_train_state_dict(...)` that we are trying to test.

tests/_task_models/test_train_model.py
```diff
@@ -0,0 +1,49 @@
...
+    def fake_load_state_dict(
+        state_dict: dict[str, Any], strict: bool = True, assign: bool = False
+    ) -> str:
+        captured["state_dict"] = state_dict
+        captured["strict"] = strict
+        captured["assign"] = assign
```
stegmuel: You can test the following scenarios:
1. EMA model is None -> the fake task model receives exactly the provided `state_dict`.
2. EMA model is not None and the `state_dict` contains the ema model -> the fake ema task model receives a `state_dict` that contains the subset of keys that started with the prefix, but with the prefix stripped from them.
3. EMA model is not None and the `state_dict` does not contain the emal model -> the ema_model.model is provided with the state_dict of the task model.

For the last one you need to implement `state_dict()` method (for the task model) that returns a dict:
```
def state_dict(self) -> dict[str, Any]:
    return {
        "backbone.weight": "train_backbone",
        "head.bias": "train_head_bias",
    }
```

masakljun:
> Left recommendations for the test.

Thanks @stegmuel, I tried to incorporate them. I had to add some # type: ignore[...] comments because of the type checker.

PR #487, opened 2025-12-09
Thomas trn 1653 ltdetr onnx export

stegmuel:
## What has changed and why?

Add onnx export method for lt-detr.

## How has it been tested?

Manual tests. 

- [x] Yes

- [x] Yes

### DIFF
--- a/src/lightly_train/_methods/distillation/distillation.py
+++ b/src/lightly_train/_methods/distillation/distillation.py
@@ -68,8 +68,8 @@ def get_teacher(
             raise FileNotFoundError(
                 f"Teacher weights file {teacher_weights} does not exist."
             )
-
-        state_dict = torch.load(teacher_weights, weights_only=True)
+        device = next(teacher_embedding_model.parameters()).device
+        state_dict = torch.load(teacher_weights, weights_only=True, map_location=device)
         teacher_embedding_model.load_state_dict(state_dict)
         logger.debug(f"Loaded teacher weights from {teacher_weights}.")
 
--- a/src/lightly_train/_methods/distillationv2/distillationv2.py
+++ b/src/lightly_train/_methods/distillationv2/distillationv2.py
@@ -63,8 +63,8 @@ def get_teacher(
             raise FileNotFoundError(
                 f"Teacher weights file {teacher_weights} does not exist."
             )
-
-        state_dict = torch.load(teacher_weights, weights_only=True)
+        device = next(teacher_embedding_model.parameters()).device
+        state_dict = torch.load(teacher_weights, weights_only=True, map_location=device)
         teacher_embedding_model.load_state_dict(state_dict)
         logger.debug(f"Loaded teacher weights from {teacher_weights}.")
 


src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
```diff
@@ -581,3 +587,118 @@ def _forward_train(self, x: Tensor, targets):  # type: ignore[no-untyped-def]
...
+
+        # Get the first parameter from the model.
+        first_parameter = next(self.parameters())
+
+        # Infer info from first parameters.
+        # TODO(Thomas, 12/25): Use a more robust approach to infer num_channels.
+        num_channels = first_parameter.shape[1]
+        model_device = first_parameter.device
+        model_dtype = first_parameter.dtype
+
```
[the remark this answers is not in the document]
stegmuel: We could add `in_channels` to the backbones.

docs/source/object_detection.md
```diff
@@ -309,3 +309,25 @@ my_data_dir/
...
+for representing machine learning models in a framework independent manner. In particular, it is useful for deploying our
+models on edge devices where PyTorch is not available.
+
+The following example shows how to export a previously trained model to ONNX.
+
+```python
+import lightly_train
+
+# Instantiate the model from a checkpoint.
+model = lightly_train.load_model_from_checkpoint(
```
guarin:
```suggestion
model = lightly_train.load_model(
```

docs/source/object_detection.md
```diff
@@ -309,3 +309,25 @@ my_data_dir/
...
+    "out/my_experiment/exported_models/exported_best.pt"
+)
+
+# Export to ONNX.
+model.export_to_onnx(
```
guarin: Let's call this `export_onnx` to be more consistent with the other function/method names like `train_object_detection` in `train_task.py` or `export_onnx` in `export_task.py`

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
```diff
@@ -500,7 +501,7 @@ def predict(
         x = transforms_functional.resize(x, self.image_size)
         x = x.unsqueeze(0)
 
-        labels, boxes, scores = self(x, orig_target_size=(h, w))
+        labels, boxes, scores = self(x, orig_target_size=torch.tensor([[h, w]]))
```
guarin, line 504: Why did you have to modify the `predict` method? The ONNX export should only call `forward` and not `predict`.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
```diff
@@ -581,3 +586,142 @@ def _forward_train(self, x: Tensor, targets):  # type: ignore[no-untyped-def]
...
+        dummy_sizes = torch.tensor(
+            [[self.image_size[0], self.image_size[1]]],
+            device=model_device,
+            dtype=torch.int64,
+        )
+
+        # TODO(Thomas, 12/25): Add warm-up forward if needed.
+
+        # Set the input/output names.
+        input_names = ["images", "orig_target_sizes"]
```
guarin: Two questions:
1. Why can we not infer the `orig_target_sizes` from the input images?
2. Can we rename `orig_target_sizes` to something more descriptive like `image_sizes` or `original_image_sizes`? Target doesn't really make sense for me at inference time.

In general I would prefer if we can omit the extra `orig_target_sizes` argument altogether.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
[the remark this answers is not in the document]
guarin: I think it is fine for now, this is very unlikely to change.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
stegmuel: ONNX was not happy with tuple input for the forward, and the predict calls the forward

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
stegmuel: 1. We can, but at this point the image is already resized so the output predictions would be resized relative to the image size after transformation (640x640). This would imply more work from  the user be predictions can be depicted on the original image.
2. Yes, we can rename that.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
guarin: Ok 👍🏼 Could you create the tensor on the same device as `x`?

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
guarin: > We can, but at this point the image is already resized so the output predictions would be resized relative to the image size after transformation (640x640). This would imply more work from the user be predictions can be depicted on the original image.

As the user is doing the resize I think it makes sense to return the predictions in 640x640. It feels a bit weird to return predictions back in a different size than the input image. I would only return bboxes in the original image coordinates if we also do the resizing. Leaving `orig_target_sizes` away will make the model much easier to use.

PR #492, opened 2025-12-10
Fix resuming error with recent torch versions

guarin:
## What has changed and why?

* Fix `fabric.load` with checkpoint that contains more than just weights

## How has it been tested?

* Added unit tests
* Tested manually

- [ ] Yes

- [ ] Yes


stegmuel (commented):
lgtm

PR #498, opened 2025-12-11
Use latency everywhere

stegmuel:
## What has changed and why?

Updated README and docs to use latency in ms everywhere.

## How has it been tested?

Not tested.

- [ ] Yes

- [x] Yes

### DIFF
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -961,8 +961,9 @@ def finetune_from_checkpoint(
         checkpoint: Checkpoint context the state dicts to load.
     """
 
-    train_model = state["train_model"]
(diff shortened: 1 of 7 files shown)


docs/source/instance_segmentation.md
```diff
@@ -23,16 +23,16 @@ You can also explore running inference and training these models using our Colab
 
 ### COCO
 
-| Implementation | Model | Val mAP mask | Avg. FPS | Params (M) | Input Size |
+| Implementation | Model | Val mAP mask | Avg. Latency [ms] | Params (M) | Input Size |
 |----------------|----------------|-------------|----------|-----------|------------|
```
guarin: Could you use the same formatting for the units? Currently we have `[ms]` and `(M)` in the table.

PR #499, opened 2025-12-11
Cleanup transforms dtypes and asserts

guarin:
## What has changed and why?

* Cleanup dtypes for transforms
* Remove unnecessary asserts in transforms

I checked transforms for possible input/output types and adjusted them to match the types we use elsewhere in the codebase.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,6 +7,7 @@
 #
 import os
 import re
+import shutil
 from pathlib import Path
 from typing import Generator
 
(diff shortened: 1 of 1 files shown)


src/lightly_train/_transforms/random_photometric_distort.py
```diff
@@ -106,6 +103,4 @@ def apply(
             Transformed image as numpy array with shape (H, W, C).
         """
         out = self.transform(image=img)["image"]
-        assert isinstance(out, np.ndarray)
-        assert out.dtype == np.uint8
```
guarin, line 110: This dtype assert was too restrictive, images can also be float32

src/lightly_train/_transforms/transform.py
```diff
@@ -119,7 +119,7 @@ def cast_list_to_tuple(cls, value: int | Sequence[int]) -> int | tuple[int, int]
             and all(isinstance(v, int) for v in value)
         ):
             value = tuple(value)
-            assert isinstance(value, tuple)
+            assert len(value) == 2
```
guarin, line 122: Old assert doesn't make sense because we just created the tuple. Check for len 2 is needed however.

PR #505, opened 2025-12-12
Add fine-tune wandb support

guarin:
## What has changed and why?

* Add wandb logger for all fine-tuning tasks

## How has it been tested?

* Manually

- [x] Yes

- [x] Yes


docs/source/semantic_segmentation.md
```diff
@@ -512,6 +512,36 @@ Disable the TensorBoard logger with:
 logger_args={"tensorboard": None}
```
stegmuel, line 514: Shouldn't the beginning of the logging section also be  updated?

guarin: Good catch!

PR #532, opened 2026-01-05
Thomas trn 1744 eomt tiny

stegmuel:
## What has changed and why?

- Allow usage of ViT-Tiny with EoMT for semantic segmentation
- Make patch size settable for faster semantic segmentation models

## How has it been tested?

Manual tests.

- [x] Yes

- [x] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add example jupyter notebooks for ONNX and TensorRT export.
 - Add tiny semantic segmentation models.
 - Add support for Ultralytics RT-DETR models.
+- Add support for different patch size in EoMT and semantic segmentation.
 
 ### Changed
 


docs/source/semantic_segmentation.md
```diff
@@ -193,6 +198,36 @@ if __name__ == "__main__":
...
+
+As shown above, the patch size can be set to a value different from the one used in the
+pretrained model without harming compatibility of the pretrained weights. Internally,
+the patch embedding weights are automatically resized to the requested patch size using
+the method introduced in [FlexiViT](https://arxiv.org/pdf/2212.08013).
+
+As illustrated in the figure below, increasing the patch size leads to a significant
+speed-up with only a moderate impact on performance.
+
+[INSERT IMAGE HERE]
```
guarin: Would add the image also at the top right after the table

src/lightly_train/_models/dinov3/dinov3_src/hub/backbones.py
```diff
@@ -151,6 +151,15 @@ def _make_dinov3_vit(
         state_dict = torch.hub.load_state_dict_from_url(
             url, map_location="cpu", check_hash=check_hash
         )
+
+        # Re-sample the projection weights before loading the statedict.
+        key = "patch_embed.proj.weight"
+        original_conv_weight = state_dict[key]
+        if original_conv_weight.shape[-1] != patch_size:
```
guarin, line 158: FYI this is how we handled adjusting the number of input channels when a checkpoint with a different number of channels is loaded: https://github.com/lightly-ai/lightly-train/blob/e502f9d45267bdf30ea0edfe366f32cea93c1975/src/lightly_train/_models/dinov3/dinov3_src/layers/patch_embed.py#L74-L77

That way you don't have to add the logic in all `load_state_dict` implementations.

guarin (commented):
Looks good! Will test it later.

Could you also add an entry in the CHANGELOG?

src/lightly_train/_models/dinov3/dinov3_src/hub/backbones.py
stegmuel: Thanks I'll do that for the next one.

guarin (approved):
LGTM

PR #535, opened 2026-01-06
Add semantic segmentation tensorrt export

guarin:
## What has changed and why?

* Add TensorRT export for EoMT semantic segmentation models
* Update docs/notebook

## How has it been tested?

* Manually

- [x] Yes

- [x] Yes

### DIFF
--- a/docs/source/semantic_segmentation.md
+++ b/docs/source/semantic_segmentation.md
@@ -36,9 +36,9 @@ You can also explore inferencing with these model weights using our Colab notebo
 | -------------- | --------------------------- | -------- | ----------------- | ---------- | ---------- |
 | LightlyTrain   | dinov3/vitt16-eomt-coco     | 37.9     | 6.0               | 6.0        | 512×512    |
 | LightlyTrain   | dinov3/vitt16plus-eomt-coco | 39.5     | 6.4               | 7.7        | 512×512    |
-| LightlyTrain   | dinov3/vits16-eomt-coco     | 46.5     | 11.3              | 21.6       | 512×512    |
-| LightlyTrain   | dinov3/vitb16-eomt-coco     | 52.0     | 23.1              | 85.7       | 512×512    |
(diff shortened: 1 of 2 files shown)


docs/source/semantic_segmentation.md
```diff
@@ -764,6 +764,20 @@ is a standard format for representing machine learning models in a framework ind
 manner. In particular, it is useful for deploying our models on edge devices where
 PyTorch is not available.
 
+### Requirements
+
+Exporting to ONNX requires some additional packages to be installed. Namely
```
stegmuel, line 769: Not related to this PR, but above I see that the training image size is 518 and the val one is 512.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
```diff
@@ -943,99 +949,13 @@ def export_tensorrt(
...
-
-        # Set the ONNX export path.
-        if onnx_args is None:
-            onnx_args = {}
-        onnx_args.setdefault("out", Path(out).with_suffix(".onnx"))
-
-        # Export the model to ONNX.
-        self.export_onnx(
-            **onnx_args,
+        tensorrt_helpers.export_tensorrt(
```
stegmuel, line 952: Not sure if it is a good idea to use the same tensorrt detection and segmentation export function. I guess it can be easily changed later.

docs/source/semantic_segmentation.md
guarin: Yes this is the section where we explain how to set different image sizes for training and validation :)

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
guarin: I re-used it because so far it was 100% identical.

docs/source/semantic_segmentation.md
stegmuel: Ooops ...

PR #540, opened 2026-01-07
Add instance segmentation onnx export

guarin:
## What has changed and why?

* Add instance segmentation onnx export
* Add docs and notebook

## How has it been tested?

* Manual

- [x] Yes

- [x] Yes


src/lightly_train/_task_models/dinov3_eomt_instance_segmentation/task_model.py
```diff
@@ -572,3 +574,164 @@ def load_train_state_dict(self, state_dict: dict[str, Any]) -> None:
...
+        )
+
+        if simplify:
+            import onnxslim  # type: ignore [import-not-found,import-untyped]
+
+            # Simplify.
+            onnxslim.slim(
+                model=str(out),
+                output_model=out,
+                # We skip constant folding as this currently increases the model size by
```
stegmuel: Tiling is not implemented for instance segmentation, I believe.
Should the comment and `skip_optimizations` arg be removed?

guarin: Good point, removed and tested it

PR #562, opened 2026-01-14
Fix LTDETR ONNX FP16 export

guarin:
## What has changed and why?

* Fix LTDETR ONNX FP16 export
* Use same verify code in `export_onnx` for all models
* Handle fp16 overflow in LTDETR ONNX FP16 verify

## How has it been tested?

* Manually

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
+++ b/src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
@@ -13,7 +13,7 @@ from typing import Any, ClassVar, Literal
 
 import torch
 from lightning_fabric import Fabric
-from pydantic import Field
+from pydantic import AliasChoices, Field
 from torch import Tensor
 from torch.nn.modules.module import _IncompatibleKeys
 from torch.optim import AdamW, Optimizer  # type: ignore[attr-defined]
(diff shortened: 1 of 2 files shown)


src/lightly_train/_task_models/object_detection_components/hybrid_encoder.py
```diff
@@ -367,12 +367,13 @@ def forward(self, feats):
...
-                    ).to(src_flatten.device)
-                else:
-                    pos_embed = getattr(self, f"pos_embed{enc_ind}", None).to(
-                        src_flatten.device
                     )
+                else:
+                    pos_embed = getattr(self, f"pos_embed{enc_ind}", None)
 
+                pos_embed = pos_embed.to(
+                    device=src_flatten.device, dtype=src_flatten.dtype
```
guarin, line 375: Converting to `src_flatten.dtype` is the actual fix

stegmuel: Is it correct that if I now train with fp16 the pos_embed is now in fp16 whereas before the fix it would have been in fp32?

guarin: Yes

guarin: With bf16 it will still be in fp32

stegmuel: How?

guarin: Sorry i mixed things up. With bf16 the pos embed will be in bf16 (or whatever autocast decides is the correct dtype for src_flatten). What I meant with the bf16 still in fp32 is that if you train with bf16 autocast then the final model weights will still be saved as fp32. But pos embed is not a model weight during training so it doesn't apply here.

stegmuel: OK, my point was that the proposed change also changes the behavior of the training, but this is not really required for the onnx issue at hand.

stegmuel (commented):
answered

PR #571, opened 2026-01-16
Thomas trn 1752 classification dataset

stegmuel:
## What has changed and why?

Adding a dataset class for image classification.

## How has it been tested?

Custom tests to verify the behavior.

- [x] Yes

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - Add Image classification dataset.
+- Add random rotation transforms for all fine-tuning tasks.
 
 ### Changed
 
--- a/src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/transforms.py
+++ b/src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/transforms.py
@@ -21,6 +21,8 @@ from lightly_train._transforms.transform import (
     NormalizeArgs,
     RandomCropArgs,
     RandomFlipArgs,
+    RandomRotate90Args,
+    RandomRotationArgs,
     ScaleJitterArgs,
     SmallestMaxSizeArgs,
 )
@@ -78,6 +80,8 @@ class DINOv2EoMTSemanticSegmentationTrainTransformArgs(
     num_channels: int | Literal["auto"] = "auto"
     normalize: NormalizeArgs | Literal["auto"] = "auto"
     random_flip: RandomFlipArgs | None = Field(default_factory=RandomFlipArgs)
+    random_rotate_90: RandomRotate90Args | None = None
+    random_rotate: RandomRotationArgs | None = None
     color_jitter: DINOv2EoMTSemanticSegmentationColorJitterArgs | None = Field(
         default_factory=DINOv2EoMTSemanticSegmentationColorJitterArgs
     )
@@ -125,6 +129,8 @@ class DINOv2EoMTSemanticSegmentationValTransformArgs(SemanticSegmentationTransfo
     num_channels: int | Literal["auto"] = "auto"
     normalize: NormalizeArgs | Literal["auto"] = "auto"
     random_flip: RandomFlipArgs | None = None
+    random_rotate_90: RandomRotate90Args | None = None
+    random_rotate: RandomRotationArgs | None = None
     color_jitter: ColorJitterArgs | None = None
     scale_jitter: ScaleJitterArgs | None = None
     smallest_max_size: SmallestMaxSizeArgs = Field(
--- a/src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/transforms.py
+++ b/src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/transforms.py
@@ -21,6 +21,8 @@ from lightly_train._transforms.transform import (
     NormalizeArgs,
     RandomCropArgs,
     RandomFlipArgs,
+    RandomRotate90Args,
+    RandomRotationArgs,
     ScaleJitterArgs,
     SmallestMaxSizeArgs,
 )
@@ -78,6 +80,8 @@ class DINOv2LinearSemanticSegmentationTrainTransformArgs(
     num_channels: int | Literal["auto"] = "auto"
     normalize: NormalizeArgs | Literal["auto"] = "auto"
     random_flip: RandomFlipArgs | None = Field(default_factory=RandomFlipArgs)
+    random_rotate_90: RandomRotate90Args | None = None
+    random_rotate: RandomRotationArgs | None = None
     color_jitter: DINOv2LinearSemanticSegmentationColorJitterArgs | None = Field(
         default_factory=DINOv2LinearSemanticSegmentationColorJitterArgs
     )
@@ -127,6 +131,8 @@ class DINOv2LinearSemanticSegmentationValTransformArgs(
     num_channels: int | Literal["auto"] = "auto"
     normalize: NormalizeArgs | Literal["auto"] = "auto"
     random_flip: RandomFlipArgs | None = None
+    random_rotate_90: RandomRotate90Args | None = None
+    random_rotate: RandomRotationArgs | None = None
     color_jitter: ColorJitterArgs | None = None
     scale_jitter: ScaleJitterArgs | None = None
     smallest_max_size: SmallestMaxSizeArgs = Field(
--- a/src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
+++ b/src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
@@ -22,6 +22,8 @@ from lightly_train._transforms.transform import (
     RandomFlipArgs,
     RandomIoUCropArgs,
     RandomPhotometricDistortArgs,
+    RandomRotate90Args,
+    RandomRotationArgs,
     RandomZoomOutArgs,
     ResizeArgs,
     ScaleJitterArgs,
@@ -110,6 +112,8 @@ class DINOv2LTDETRObjectDetectionTrainTransformArgs(ObjectDetectionTransformArgs
     random_flip: DINOv2LTDETRObjectDetectionRandomFlipArgs | None = Field(
         default_factory=DINOv2LTDETRObjectDetectionRandomFlipArgs
     )
+    random_rotate_90: RandomRotate90Args | None = None
+    random_rotate: RandomRotationArgs | None = None
     image_size: ImageSizeTuple | Literal["auto"] = "auto"
     # TODO: Lionel (09/25): Remove None, once the stop policy is implemented.
     stop_policy: StopPolicyArgs | None = None
@@ -175,6 +179,8 @@ class DINOv2LTDETRObjectDetectionValTransformArgs(ObjectDetectionTransformArgs):
     random_zoom_out: None = None
     random_iou_crop: None = None
     random_flip: None = None
+    random_rotate_90: RandomRotate90Args | None = None
+    random_rotate: RandomRotationArgs | None = None
     image_size: ImageSizeTuple | Literal["auto"] = "auto"
     stop_policy: None = None
     resize: ResizeArgs | None = Field(
--- a/src/lightly_train/_task_models/dinov3_eomt_instance_segmentation/transforms.py
+++ b/src/lightly_train/_task_models/dinov3_eomt_instance_segmentation/transforms.py
@@ -22,6 +22,8 @@ from lightly_train._transforms.transform import (
     NormalizeArgs,
     RandomCropArgs,
     RandomFlipArgs,
+    RandomRotate90Args,
+    RandomRotationArgs,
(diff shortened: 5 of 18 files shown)


docs/source/image_classification.md
```diff
@@ -0,0 +1,166 @@
...
+        model="dinov3/vits16-classification",
+        data={
+            "train": "my_data_dir/train/",
+            "val": "my_data_dir/val/",
+            "classes": {                    
+                0: "cat",
+                1: "car",
+                2: "dog",
+                # ...
+            },
```
guarin, line 81: This doesn't match the `class_0`, `class_1`, etc. naming of the directories. I would actually rename the directories to `cat`, `car`, etc. This makes it explicit that any class names are supported

docs/source/image_classification.md
```diff
@@ -0,0 +1,166 @@
...
+        },
+    )
+```
+
+In this setup:
+
+- Each image belongs to exactly one class.
+- Class names are taken from the directory names.
+- Class IDs are assigned according to the names mapping passed to
+  ImageClassificationDataArgs.
```
guarin:
```suggestion
- Class IDs are assigned according to the passed `classes` dictionary
```

Users don't know about `ImageClassificationDataArgs`.

docs/source/image_classification.md
```diff
@@ -0,0 +1,166 @@
...
+- Labels can be specified either as class IDs or class names.
+
+#### Supported CSV Options
+
+The behavior of CSV parsing can be configured via the `data` argument:
+
+- `csv_image_col`: Name of the image path column (default: `"image_path"`).
+- `csv_label_col`: Name of the label column (default: `"label"`).
+- `csv_label_type`: "name" or `"id"` (default: `"name"`).
+- `label_delimiter`: Delimiter for multiple labels (default: `","`).
```
guarin:
```suggestion
- `csv_image_column`: Name of the image path column (default: `"image_path"`).
- `csv_label_column`: Name of the label column (default: `"label"`).
- `csv_label_type`: "name" or `"id"` (default: `"name"`).
- `label_delimiter`: Delimiter for multiple labels (default: `","`).
```

Should there also be a `csv_column_delimiter`?  Would use `column` instead of `col` to make it clearer what the field represents

docs/source/image_classification.md
```diff
@@ -0,0 +1,166 @@
...
+Notes:
+
+- Image paths must be absolute paths.
+- Multiple labels are separated by a delimiter (default: ,).
+- When using commas as label delimiters, the label field must be quoted.
+- Labels can be specified either as class IDs or class names.
+
+#### Supported CSV Options
+
+The behavior of CSV parsing can be configured via the `data` argument:
```
guarin, line 193: A code snippet would be great here

src/lightly_train/_data/image_classification_dataset.py
```diff
@@ -0,0 +1,289 @@
...
+        Args:
+            class_ids_str: Delimiter-separated class ID string (e.g. "3,7,12").
+
+        Returns:
+            1D tensor of internal class IDs (dtype=torch.long).
+        """
+        class_ids_str = class_ids_str.strip()
+        internal_class_ids = []
+        for class_id_str in class_ids_str.split(self.dataset_args.label_delimiter):
+            if class_id_str.strip() != "":
```
guarin:
```suggestion
            class_id_str = class_id_str.strip()
            if class_id_str:
```

Then the same value is reused later on in `int(class_id_str)`

src/lightly_train/_data/image_classification_dataset.py
```diff
@@ -0,0 +1,289 @@
...
+
+    def __getitem__(self, index: int) -> ImageClassificationDatasetItem:
+        # Load the image.
+        image_info = self.image_info[index]
+        image_path = Path(image_info["image_path"])
+        class_ids_str = image_info["class_id"]
+
+        # Verify that the image exists.
+        if not image_path.exists():
+            raise FileNotFoundError(f"Image file {image_path} does not exist.")
```
guarin: This should already be checked in `list_image_info`

src/lightly_train/_data/image_classification_dataset.py
```diff
@@ -0,0 +1,289 @@
...
+            if self.csv_label_col not in reader.fieldnames:
+                raise ValueError(
+                    f"CSV {self.annotations_csv} missing required column '{self.csv_label_col}'. "
+                    f"Found columns: {reader.fieldnames}"
+                )
+
+            # Iterate over the csv's rows.
+            for row in reader:
+                image_path = (row.get(self.csv_image_col) or "").strip()
+                labels_str = (row.get(self.csv_label_col) or "").strip()
```
guarin:
```suggestion
                image_path = (row.get(self.csv_image_col, "")).strip()
                labels_str = (row.get(self.csv_label_col, "")).strip()
```

src/lightly_train/_data/image_classification_dataset.py
```diff
@@ -0,0 +1,289 @@
...
+                # TODO(Thomas, 01/26): Add a flag to disable skipping invalid rows.
+                if image_path == "" or labels_str == "":
+                    continue
+
+                # Verify that the image path is absolute.
+                if not Path(image_path).is_absolute():
+                    raise ValueError(
+                        f"CSV {self.annotations_csv}: '{self.csv_image_col}' must be an absolute path "
+                        f"but got '{image_path}'."
+                    )
```
guarin: Do we really need absolute paths?

src/lightly_train/_data/image_classification_dataset.py
```diff
@@ -0,0 +1,289 @@
...
+                        for class_name_str in labels_str.split(self.label_delimiter)
+                        if class_name_str.strip() != ""
+                    ]
+                else:
+                    # Handle potential spaces in class IDs.
+                    class_ids = [
+                        int(class_id.strip())
+                        for class_id in labels_str.split(self.label_delimiter)
+                        if class_id.strip() != ""
+                    ]
```
guarin: Would create class_ids as `set` instead of `list` to avoid duplicate entries

src/lightly_train/_data/image_classification_dataset.py
```diff
@@ -0,0 +1,289 @@
...
+
+                # Don't collect from ignore classes.
+                if self.ignore_classes is not None:
+                    class_ids = [
+                        class_id
+                        for class_id in class_ids
+                        if class_id not in self.ignore_classes
+                    ]
```
guarin: Set instead of list

src/lightly_train/_data/image_classification_dataset.py
```diff
@@ -0,0 +1,289 @@
...
+        }
+
+
+class ImageClassificationDatasetArgs(TaskDatasetArgs):
+    image_dir: Path
+    classes: dict[int, str]
+    ignore_classes: set[int] | None
+
+    # Optional .csv file with mapping from image path to label.
+    annotations_csv: Path | None = None
```
guarin:
```suggestion
    dir_or_file: Path
    classes: dict[int, str]
    ignore_classes: set[int] | None
```

Suggestion: Instead of saving a separate `annotations_csv` file I would use the same `dir_or_file` field for both `image_dir` and `annotations_csv` (and any potential future format). We can check whether it is a directory/csv/txt/json etc. based on `dir_or_file.is_dir()` or file extensions.

src/lightly_train/_data/image_classification_dataset.py
```diff
@@ -0,0 +1,289 @@
...
+    val: PathLike
+    test: PathLike | None = None
+    classes: dict[int, str]
+    label_delimiter: str = ","
+    ignore_classes: set[int] | None = Field(default=None, strict=False)
+
+    # Paths to .csv files mapping image paths and labels.
+    train_csv: PathLike | None = None
+    val_csv: PathLike | None = None
+    test_csv: PathLike | None = None
```
guarin:
```suggestion
```

Remove and let user set `train="annotations.csv"` and handle directory/csv based on path type and file extension

src/lightly_train/_data/image_classification_dataset.py
```diff
@@ -0,0 +1,156 @@
...
+                continue
+
+            for image_filename in file_helpers.list_image_filenames_from_dir(
+                image_dir=class_dir
+            ):
+                image_filepath = class_dir / Path(image_filename)
+                # Labels are comma-separated to support multi-labels.
+                yield {
+                    "image_path": str(image_filepath),
+                    "class_id": ",".join(map(str, [class_id])),
```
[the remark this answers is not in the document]
guarin: Agree

src/lightly_train/_data/image_classification_dataset.py
stegmuel: I had this version first, but if the key exists and is None (csv generated with pandas, etc.) it would crash.

src/lightly_train/_data/image_classification_dataset.py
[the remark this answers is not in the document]
stegmuel: The copilot comment was outdated.

docs/source/image_classification.md
stegmuel: Changed the names.
About the delimiter I don't think so. Right now it should already work with any delimiter as long as the labels are in quotes.

src/lightly_train/_data/image_classification_dataset.py
stegmuel: What would you use as root dir or would you have done it?

src/lightly_train/_data/image_classification_dataset.py
guarin: Just to double check, `row.get()` always returns a string if there is a value right? Because if it parses int/floats automatically the following could happen if the value is `0`
```
(row.get(self.csv_label_col) or "").strip()
# Evaluates to
(0 or "").strip()
# Evaluates to
""  <-- Should be 0 instead
```

src/lightly_train/_data/image_classification_dataset.py
guarin: Hmm I would just use the path the user provided and let pathlib resolve it.

src/lightly_train/_data/image_classification_dataset.py
stegmuel: Yes, it will return a string

src/lightly_train/_data/image_classification_dataset.py
stegmuel: If we just resolve it, the resulting path will depend on the working directory which does not seem ideal.
I can resolve it w.r.t. the directory containing the csv file when the path is not absolute, wdyt?

docs/source/image_classification.md
```diff
@@ -0,0 +1,188 @@
...
+            # Optional, classes that are in the dataset but should be ignored during
+            # training.
+            "ignore_classes": [0], 
+        },
+    )
+```
+
+Notes:
+
+- Image paths must be absolute paths.
```
guarin: This is now outdated

src/lightly_train/_data/image_classification_dataset.py
```diff
@@ -0,0 +1,274 @@
...
+            # Only consider directories that are in `classes`.
+            if not class_dir.exists():
+                continue
+            if not class_dir.is_dir():
+                continue
+
+            for image_filename in file_helpers.list_image_filenames_from_dir(
+                image_dir=class_dir
+            ):
+                image_filepath = class_dir / Path(image_filename)
```
guarin: No need to convert this to a Path before
```suggestion
                image_filepath = class_dir / image_filename
```

src/lightly_train/_data/image_classification_dataset.py
```diff
@@ -0,0 +1,274 @@
...
+                    "image_path": str(image_filepath),
+                    "class_id": str(class_id),
+                }
+
+    def _list_image_info_from_csv(self) -> Iterable[dict[str, str]]:
+        # Map directory/class name to class id.
+        name_to_id = {name: class_id for class_id, name in self.classes.items()}
+
+        # Verify the .csv files is provided and exists.
+        assert self.dir_or_file is not None
```
guarin: Impossible

src/lightly_train/_data/image_classification_dataset.py
```diff
@@ -0,0 +1,275 @@
...
+                # TODO(Thomas, 01/26): Add a flag to disable skipping invalid rows.
+                if image_path == "" or labels_str == "":
+                    continue
+
+                # Resolve relative paths against the CSV file location.
+                image_path_p = Path(image_path)
+                if not image_path_p.is_absolute():
+                    image_path_p = (self.dir_or_file.parent / image_path_p).resolve()
+                else:
+                    image_path_p = image_path_p.resolve()
```
guarin: Instead of calling `.resolve()` I would check if the image exists with `image_path_p.is_file()`, this avoids crashing training if a file doesn't exist. And before calling `is_file()` I would check if the file extension is one of our supported image extensions (this avoids the system call from `.is_file()` in case the extension is not supported).

Would also resolve `self.dir_or_file.parent` outside of the for loop.

PR #575, opened 2026-01-19
get model_init_args  from checkpoint

stegmuel:
## What has changed and why?

Fix resuming from interrupted runs.

## How has it been tested?

Manual tests.

- [x] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/train_task.py
+++ b/src/lightly_train/_commands/train_task.py
@@ -720,10 +720,7 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
         train_model_cls=train_model_cls, checkpoint_args=config.save_checkpoint_args
     )
 
-    if model_init_args is None:
-        model_init_args = (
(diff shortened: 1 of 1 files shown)


src/lightly_train/_commands/train_task.py
```diff
@@ -718,9 +720,10 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
         train_model_cls=train_model_cls, checkpoint_args=config.save_checkpoint_args
     )
 
-    model_init_args = (
-        {} if checkpoint is None else checkpoint.get("model_init_args", {})
-    )
+    if model_init_args is None:
+        model_init_args = (
+            {} if checkpoint is None else checkpoint.get("model_init_args", {})
+        )
```
guarin: Shouldn't be needed anymore
```suggestion
    model_init_args = {} if model_init_args is None else model_init_args
```

The `checkpoint.get("model_init_args", {})` should now already be handled in `load_checkpoint`

PR #580, opened 2026-01-21
fix: ensure boolean mask dtype in YOLOObjectDetectionDataset

ucsk:
## What has changed and why?

Fixed `IndexError` in `YOLOObjectDetectionDataset` by adding `dtype=bool` to boolean mask creation. This ensures the array is always boolean type for indexing, preventing crashes when class labels contain non-integer values.

## How has it been tested?

Tested locally with corrupted label files. The fix resolves the error and training continues normally. Valid label files still work correctly.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- Add Image classification dataset.
+- Add random rotation transforms for all fine-tuning tasks.
+
 ### Changed
 
 ### Deprecated
--- a/docs/source/image_classification.md
+++ b/docs/source/image_classification.md
@@ -1,0 +1,221 @@
+---
+orphan: true
+---
+
+(image-classification)=
+
+# Image Classification
+
+(image-classification-data)=
+
+## Data
+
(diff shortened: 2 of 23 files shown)


guarin:
Thanks for the PR! Could you give an example file for which this crashed? I don't fully understand how the array couldn't be boolean

ucsk:
> Thanks for the PR! Could you give an example file for which this crashed? I don't fully understand how the array couldn't be boolean

During testing, I discovered that empty annotation files trigger an `IndexError` in the current implementation.

Looking into this, I found an interesting discrepancy:
- YOLO's format states that images without objects don't need annotation files ([link](https://docs.ultralytics.com/datasets/detect/#ultralytics-yolo-format))
- LightlyTrain's documentation suggests every image should have one ([link](https://docs.lightly.ai/train/stable/object_detection.html#data))

This difference could confuse users. Technically, when processing empty files, numpy infers `float64` dtype for empty boolean masks instead of `bool`, causing the crash. (`keep.dtype` is `float64`, not `bool`)

This PR adds a minimal fix (`dtype=bool`) that prevents the crash, serving as a safety net regardless of which annotation policy is ultimately adopted.

guarin:
Thanks! Will definitely look into this

PR #582, opened 2026-01-22
Add distillationv3

stegmuel:
## What has changed and why?

Add distillationv3.

## How has it been tested?

Manual tests.

- [x] Yes

- [x] Yes

### DIFF
--- a/docs/prebuild.py
+++ b/docs/prebuild.py
@@ -15,15 +15,6 @@ from lightly_train._methods import method_helpers
 from lightly_train._task_models.dinov2_linear_semantic_segmentation.train_model import (
     DINOv2LinearSemanticSegmentationTrain,
 )
-from lightly_train._task_models.dinov2_ltdetr_object_detection.train_model import (
-    DINOv2LTDETRObjectDetectionTrain,
-)
-from lightly_train._task_models.dinov3_ltdetr_object_detection.train_model import (
-    DINOv3LTDETRObjectDetectionTrain,
-)
-from lightly_train._task_models.picodet_object_detection.train_model import (
-    PicoDetObjectDetectionTrain,
-)
 
 THIS_DIR = Path(__file__).parent.resolve()
 DOCS_DIR = THIS_DIR / "source"
@@ -103,13 +94,8 @@ def dump_transform_args_for_methods(dest_dir: Path) -> None:
 def dump_transform_args_for_tasks(dest_dir: Path) -> None:
     dest_dir.mkdir(parents=True, exist_ok=True)
     for train_model_cls in TASK_TRAIN_MODEL_CLASSES:
-        # TODO(Thomas, 10/25): Allow to dump transform args for object detection tasks too.
-        # TODO(Guarin, 11/25): Allow to dump transform args for instance segmentation tasks too.
         if train_model_cls in {
             DINOv2LinearSemanticSegmentationTrain,
-            DINOv2LTDETRObjectDetectionTrain,
-            DINOv3LTDETRObjectDetectionTrain,
-            PicoDetObjectDetectionTrain,
         }:
             continue
         transform_args_cls = train_model_cls.train_transform_cls.transform_args_cls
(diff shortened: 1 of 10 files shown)


src/lightly_train/_methods/distillationv3/distillationv3.py
```diff
@@ -0,0 +1,463 @@
...
+        # Discard empty spatial dimensions: (B, C, 1, 1) -> (B, C).
+        x_global = self.flatten(x_global)
+
+        # (B, C, H, W) -> (B, H * W, C)
+        x_local = x_local.permute(0, 2, 3, 1)
+        # x_local = x_local.permute(0, 2, 3, 1).flatten(start_dim=1, end_dim=2)
+
+        # Forward the student features through the projection head to match the dimension of the teacher: (B, C) -> (B, D).
+        x_global = self.student_projection_head_global(x_global)
+        x_local = self.student_projection_head_global(x_local)
```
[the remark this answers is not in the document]
stegmuel: Indeed.

src/lightly_train/_methods/distillationv3/distillationv3.py
```diff
@@ -0,0 +1,467 @@
...
+
+    # Default number of teacher embeddings to store in the queue to serve as pseudo classification weights.
+    queue_size: int | Literal["auto"] = "auto"
+
+    # Default temperature parameter to regulate the sharpness of the distributions in the loss.
+    temperature_global: float = 0.07
+    temperature_local: float = 0.07
+
+    # Default teacher.
+    teacher: str = "dinov3/vitb16"
```
guarin, line 99: Would it make sense to increase to L?

src/lightly_train/_methods/distillationv3/distillationv3.py
```diff
@@ -0,0 +1,467 @@
...
+
+    # Optional teacher weight path.
+    teacher_weights: str | Path | None = None
+
+    # Deprecated. Does not have any effect.
+    teacher_url: str | None = None
```
guarin: Drop

src/lightly_train/_methods/distillationv3/distillationv3.py
```diff
@@ -0,0 +1,467 @@
...
+
+    # Scaling method for the learning rate.
+    lr_scale_method: Literal["linear", "sqrt"] = "sqrt"
+    reference_batch_size: int = 1536
+
+    # Loss weighting.
+    local_loss_weight: float = 1.0
```
guarin:
```suggestion
    # Loss weighting.
    loss_local_weight: float = 1.0
```

To match `loss_` prefix naming from task models

src/lightly_train/_methods/distillationv3/distillationv3.py
```diff
@@ -0,0 +1,467 @@
...
+                ]
+            ),
+        )
+
+    def training_step_impl(self, batch: Batch, batch_idx: int) -> TrainingStepResult:
+        # Get the images. In distillation, we only use one view.
+        views = batch["views"][0]
+
+        # Mixup the data.
+        views = self._mixup_data(views)
```
guarin, line 227: Could this hurt the intra-image loss?

src/lightly_train/_methods/distillationv3/distillationv3.py
stegmuel: based on one experiment, no.
With yolov8s: vits16plus < vitl16 < vitb16
L and B were quite similar.

src/lightly_train/_methods/distillationv3/distillationv3.py
stegmuel: As a side note, it would not make sense with the local loss only, which is why I don't apply a (1. - loss_local_weight) term to the global loss.

src/lightly_train/_methods/distillationv3/distillationv3.py
stegmuel: I don't think so. The issue was more that when the loss includes both inter-sample and intra-sample consistency, the intra-sample difference are not "visible" to the student.

PR #588, opened 2026-01-29
DINO: Use by step schedules by default

guarin:
## What has changed and why?

* Update teacher temperature every step
* Update freeze last layer every step

## How has it been tested?

* Manually

- [x] Yes

- [ ] Yes


src/lightly_train/_methods/dino/dino.py
```diff
@@ -63,6 +68,11 @@ class DINOArgs(MethodArgs):
     # weight decay
     weight_decay_start: float | Literal["auto"] = "auto"
     weight_decay_end: float | Literal["auto"] = "auto"
+    # learning rate
+    # Default lr warmup steps based on 10 epochs on ImageNet with batch size 8192:
+    # 10 * 1280000 / 8192 ~= 1560
+    warmup_steps: int = 1560
```
stegmuel: Would have used 12500. 
Would be consistent with a batch size of 1024 (default in DINO) and 12500 is also more aligned with DINOv2

src/lightly_train/_methods/dino/dino.py
```diff
@@ -48,13 +50,16 @@ class DINOArgs(MethodArgs):
...
+    student_freeze_last_layer_steps: int | Literal["auto"] | None = "auto"
     batch_norm: bool = False
     norm_last_layer: bool = True
     # loss
     teacher_temp: float | Literal["auto"] = "auto"
     warmup_teacher_temp: float | Literal["auto"] = "auto"
-    warmup_teacher_temp_epochs: int | Literal["auto"] = "auto"
+    warmup_teacher_temp_epochs: int | None = None
+    warmup_teacher_temp_steps: int | Literal["auto"] | None = "auto"
+    warmup_teacher_temp_max_steps_fraction: float = 0.3  # Max 30% of total steps.
```
stegmuel, line 62: Think max(0.3 * totat_steps, 37500) would be safer.
Increasing the temperature too fast can make the loss stagnate. 
It should be ok if the temp warmup is not done by the end of the training

src/lightly_train/_methods/dino/dino.py
```diff
@@ -112,19 +122,27 @@ def resolve_auto(
...
-                    value_start=0,
-                    value_end=30,
-                )
-            )
+        if self.warmup_teacher_temp_steps == "auto":
+            if self.warmup_teacher_temp_epochs is None:
+                # Default DINO settings are 30 epochs warmup on ImageNet with 1.28M images
+                # at batch size 8192. This is 30 * 1280000 / 8192 ~= 4700 steps.
+                # We don't want to warmup for a fixed number of epochs because that would
+                # be too long for large datasets that are only trained for a few epochs.
```
stegmuel, line 148: Same as above.

src/lightly_train/_methods/dino/dino.py
```diff
@@ -112,19 +122,27 @@ def resolve_auto(
...
+                self.warmup_teacher_temp_steps = None
+
+        if self.student_freeze_last_layer_steps == "auto":
+            if self.student_freeze_last_layer_epochs is None:
+                # Default DINO settings are 1 epoch freeze on ImageNet with 1.28M images
+                # at batch size 8192. This is 1 * 1280000 / 8192 ~= 160 steps.
+                # We don't want to freeze for a fixed number of epochs because that would
+                # be too long for large datasets that are only trained for a few epochs.
+                # So we set a fixed number of steps.
+                self.student_freeze_last_layer_steps = 160
```
stegmuel: Would be in favor of using 1250 (see [here](https://github.com/facebookresearch/dinov2/blob/b194f00db6136677fc8a4cc2ef2168f7699dfba2/dinov2/configs/ssl_default_config.yaml#L68) and [here](https://github.com/facebookresearch/dinov2/blob/b194f00db6136677fc8a4cc2ef2168f7699dfba2/dinov2/configs/ssl_default_config.yaml#L106)).

src/lightly_train/_methods/dino/dino.py
```diff
@@ -301,10 +405,60 @@ def on_before_optimizer_step(self, optimizer: Optimizer, *args: Any) -> None:
...
+        updates = [{"name": "params", "weight_decay": weight_decay}]
+
+        freeze_last_layer_steps = no_auto(
+            self.method_args.student_freeze_last_layer_steps
         )
+        if freeze_last_layer_steps is None:
+            if self.method_args.student_freeze_last_layer_epochs is None:
+                raise ValueError(
+                    "Either student_freeze_last_layer_epochs or "
+                    "student_freeze_last_layer_steps must be set."
```
stegmuel, line 456: Conversion seems incorrect.

stegmuel, line 456: In DINOv1 they cancel the gradients, which is slightly different from the above.
But I think setting lr to 0 is probably better so it should be fine.

src/lightly_train/_methods/dino/dino.py
```diff
@@ -301,10 +405,60 @@ def on_before_optimizer_step(self, optimizer: Optimizer, *args: Any) -> None:
...
+
+        optim.update_param_groups(optimizer, updates=updates)
 
     @staticmethod
     def transform_cls() -> type[MethodTransform]:
         return DINOTransform
+
+
+def _teacher_temp_schedule(
+    temp: float,
```
stegmuel, line 484: Both are already int. Is the casting needed?

src/lightly_train/_methods/dino/dino.py
```diff
@@ -292,7 +331,72 @@ def configure_gradient_clipping(
             gradient_clip_val=3.0,
             gradient_clip_algorithm="norm",
         )
-        self.student_projection_head.cancel_last_layer_gradients(self.current_epoch)
+
+    # Ignore the return type, because pytorch-lightning types it wrongly.
+    # See https://github.com/Lightning-AI/pytorch-lightning/issues/20106
+    def configure_optimizers(self) -> OptimizerLRScheduler:
+        # Scale the learning rate based on the global batch size.
+        lr_scale: float = self.global_batch_size / self.method_args.reference_batch_size
```
stegmuel, line 377: Are `reference_batch_size` and `lr_scale_method` defined?
FYI use default should be 256 and "linear" for DINOv1

src/lightly_train/_methods/dino/dino.py
guarin: Yes I took it over from DINOv2

src/lightly_train/_methods/dino/dino.py
guarin: Yes they are defined on the base class

src/lightly_train/_methods/dino/dino.py
```diff
@@ -292,7 +332,74 @@ def configure_gradient_clipping(
...
+        if self.method_args.lr_scale_method == "sqrt":
+            lr_scale = math.sqrt(lr_scale)
+
+        self.optimizer_args.lr *= lr_scale  # type: ignore[attr-defined]
```
yutong-xiang-97: From Codex with I think is valid:

"When `global_batch_size` differs from `reference_batch_size`, `lr_scale` should only be applied once. Here `self.optimizer_args.lr` is multiplied by `lr_scale` and then `get_optimizer(..., lr_scale=lr_scale)` applies the scaling again. This results in an effective LR of `base_lr * lr_scale^`2, which can significantly destabilize training and will compound if `configure_optimizers` runs again (e.g., resume). Consider removing the in-place scaling and rely on `get_optimizer`’s `lr_scale` parameter.
`

src/lightly_train/_methods/dino/dino.py
yutong-xiang-97: From Codex:

The new freeze logic sets `lr` to `0` for `params_last_layer`, but gradients are still computed and optimizer state (momentum/Adam moments) continues to update while frozen. With SGD/AdamW (the defaults) and a non-zero `student_freeze_last_layer_steps`, this accumulates state and can cause a sudden jump when unfreezing, which differs from the previous `cancel_last_layer_gradients` behavior. To preserve the intended freeze semantics, you likely also need to zero/skip grads or otherwise block optimizer-state updates for the last layer during the freeze window.

src/lightly_train/_methods/dino/dino.py
```diff
@@ -48,13 +50,16 @@ class DINOArgs(MethodArgs):
     hidden_dim: int = 2048
     bottleneck_dim: int = 256
     output_dim: int | Literal["auto"] = "auto"
-    student_freeze_last_layer_epochs: int = 1
+    student_freeze_last_layer_epochs: int | None = None
```
yutong-xiang-97: What should we actually do if both epoch and step params are set? For now the logic is to simply use the steps and ignore the epochs, but this message is not clearly conveyed to the user. 

Or, we actually want to deprecate the epoch-based settings and leave it here for backward compatibility? If so, we may need to add an explicit deprecation warning.

src/lightly_train/_methods/dino/dino.py
```diff
@@ -240,13 +257,36 @@ def training_step_impl(self, batch: Batch, batch_idx: int) -> TrainingStepResult
             # Process only global views
             x_student = self._forward_student(global_views)
 
+        if self.trainer.max_epochs is None:
```
yutong-xiang-97, line 297: This could be moved to `configure_optimizers` to fail earlier.

src/lightly_train/_methods/dino/dino.py
```diff
@@ -240,13 +257,36 @@ def training_step_impl(self, batch: Batch, batch_idx: int) -> TrainingStepResult
...
+
+        teacher_temp = _teacher_temp_schedule(
+            temp=no_auto(self.method_args.teacher_temp),
+            warmup_temp=no_auto(self.method_args.warmup_teacher_temp),
+            warmup_epochs=self.method_args.warmup_teacher_temp_epochs,
+            warmup_steps=no_auto(self.method_args.warmup_teacher_temp_steps),
+            warmup_max_steps_fraction=self.method_args.warmup_teacher_temp_max_steps_fraction,
+            step=self.trainer.global_step,
+            max_steps=int(self.trainer.estimated_stepping_batches),
+            steps_per_epoch=int(
```
yutong-xiang-97: Not to be fixed for now but round down with `int` can lead to a zero with very small datasets. Maybe consider ceiling it instead.

src/lightly_train/_methods/dino/dino.py
guarin: Not relevant. Grads are zeroed after each step. This is same behavior as in DINOv2 which is preferred over setting grads to None which has undesired side effects.

src/lightly_train/_methods/dino/dino.py
guarin: Good point, updated!

src/lightly_train/_methods/dino/dino.py
guarin: We anyways have to check that it is not None for mypy

src/lightly_train/_methods/dino/dino.py
guarin: Updated

src/lightly_train/_methods/dino/dino.py
guarin: Good point, updated!

src/lightly_train/_methods/dino/dino.py
```diff
@@ -268,7 +305,7 @@ def training_step_impl(self, batch: Batch, batch_idx: int) -> TrainingStepResult
             warmup_max_steps_fraction=self.method_args.warmup_teacher_temp_max_steps_fraction,
             step=self.trainer.global_step,
             max_steps=int(self.trainer.estimated_stepping_batches),
-            steps_per_epoch=int(
+            steps_per_epoch=math.ceil(
```
yutong-xiang-97, line 308: On line 461 there is the same problem. Sorry for missing this

src/lightly_train/_methods/dino/dino.py
yutong-xiang-97: OK I will leave it as is if it does not cause the effect mentioned. In the full agent review md there is indeed a test results the agent used a script to generate. That's why I am asking

src/lightly_train/_methods/dino/dino.py
yutong-xiang-97, line 483: From AI review:

`warmup_teacher_temp_max_steps_fraction` is only applied when `warmup_steps` is `None`. For the default path (`warmup_teacher_temp_steps="auto"`), `resolve_auto()` sets a fixed step count (e.g., 37500), so the max-fraction cap is skipped and warmup can exceed the configured fraction of total steps (including lasting the entire run for small trainings). This contradicts the arg comment (“Max 30% of total steps”) and likely defeats the intended safeguard. Consider clamping `warmup_steps` regardless of whether it came from epochs or steps.

guarin: Yes this is intended as is

PR #591, opened 2026-01-30
Fix dynamic batch sizes for onnx export

simonschoelly:
## What has changed and why?

The previous code did not handle dynamic batch sizes correctly, this PR intends to change that. In particular:

- If the batch size is dyamic, we should always run it with a dummy tensor of size > 1 as otherwise PyTorch might incorrectly think the batch size is not dynamic, even if we told the model so.
- We should ensure that one can always avoid dynamic batch sizes, as non-dynamic models tend to be faster.

Furthermore:
- Added the `onnx_helpers.precalculate_for_onnx_export` function in some places as this helps to avoid a bug where positional encodings differ when exporting to ONNX.
- I discovered some issues with the ONNX export in `DINOv3EoMTSemanticSegmentation` - I've added some failing tests that we should resolve in a future issue.

## How has it been tested?

Added a lot of tests as coverage was lacking before.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_task_models/dinov2_eomt_panoptic_segmentation/task_model.py
+++ b/src/lightly_train/_task_models/dinov2_eomt_panoptic_segmentation/task_model.py
@@ -23,7 +23,7 @@ from torchvision.transforms.v2 import functional as transforms_functional
 
 from lightly_train import _logging, _torch_helpers, _torch_testing
 from lightly_train._data import file_helpers
-from lightly_train._export import tensorrt_helpers
+from lightly_train._export import onnx_helpers, tensorrt_helpers
 from lightly_train._models import package_helpers
 from lightly_train._models.dinov2_vit.dinov2_vit_package import DINOV2_VIT_PACKAGE
 from lightly_train._models.dinov2_vit.dinov2_vit_src.layers.attention import Attention
@@ -1032,6 +1032,10 @@ class DINOv2EoMTPanopticSegmentation(TaskModel):
             device=model_device,
             dtype=dtype,
         )
+
+        # Precalculate interpolated positional encoding for ONNX export.
+        with onnx_helpers.precalculate_for_onnx_export():
+            self(dummy_input)
 
         input_names = ["images"]
         output_names = ["masks", "segment_ids", "scores"]
--- a/src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/task_model.py
+++ b/src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/task_model.py
@@ -697,6 +697,7 @@ class DINOv2EoMTSemanticSegmentation(TaskModel):
         *,
         precision: Literal["auto", "fp32", "fp16"] = "auto",
         batch_size: int = 1,
+        dynamic_batch_size: bool = True,
         height: int | None = None,
         width: int | None = None,
         opset_version: int | None = None,
@@ -706,10 +707,11 @@ class DINOv2EoMTSemanticSegmentation(TaskModel):
     ) -> None:
         """Exports the model to ONNX for inference.
 
-        The export uses a dummy input of shape (batch_size, C, H, W) where C is inferred
-        from the first model parameter and (H, W) come from `self.image_size`.
-        The ONNX graph uses dynamic batch size for both inputs and produces
-        two outputs: masks and logits.
+        The export uses a dummy input of shape (batch_size, C, H, W) where C is
+        inferred from the first model parameter and (H, W) come from
+        `self.image_size`. If `dynamic_batch_size` is True, the ONNX graph will
+        have a dynamic batch dimension for the input. The graph produces two
+        outputs: masks and logits.
 
         Optionally simplifies the exported model in-place using onnxslim and
         verifies numerical closeness against a float32 CPU reference via
@@ -723,6 +725,9 @@ class DINOv2EoMTSemanticSegmentation(TaskModel):
                 uses the model's current precision.
             batch_size:
                 Batch size for the ONNX input.
+            dynamic_batch_size:
+                If True, the ONNX graph will have a dynamic batch dimension for the
+                input. If False, the batch dimension is fixed to `batch_size`.
             height:
                 Height of the ONNX input. If None, will be taken from `self.image_size`.
             width:
@@ -768,6 +773,10 @@ class DINOv2EoMTSemanticSegmentation(TaskModel):
         width = self.image_size[1] if width is None else width
         num_channels = len(self.image_normalize["mean"])
 
+        if dynamic_batch_size:
+            batch_size = 2
+        dynamic_axes = {"images": {0: "N"}} if dynamic_batch_size else None
+
         dummy_input = torch.randn(
             batch_size,
             num_channels,
@@ -793,7 +802,7 @@ class DINOv2EoMTSemanticSegmentation(TaskModel):
             output_names=output_names,
             opset_version=opset_version,
             dynamo=False,
-            dynamic_axes={"images": {0: "N"}},
+            dynamic_axes=dynamic_axes,
             **(format_args or {}),
         )
 
--- a/src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/task_model.py
+++ b/src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/task_model.py
@@ -242,6 +242,7 @@ class DINOv2LinearSemanticSegmentation(TaskModel):
 
     def forward(self, x: Tensor) -> tuple[Tensor, Tensor]:
         # Function used for ONNX export
+        # TODO (Simon, 05/26) This class does not seem to have an export_onnx function
         logits = self._forward_logits(x)  # (B, K|K+1, H, W), K=num_classes
         if self.class_ignore_index is not None:
             # Restrict logits to known classes only.
--- a/src/lightly_train/_task_models/image_classification/task_model.py
+++ b/src/lightly_train/_task_models/image_classification/task_model.py
@@ -366,6 +366,7 @@ class ImageClassification(TaskModel):
         *,
         precision: Literal["auto", "fp32", "fp16"] = "auto",
         batch_size: int = 1,
+        dynamic_batch_size: bool = True,
         height: int | None = None,
         width: int | None = None,
         opset_version: int | None = None,
@@ -375,10 +376,11 @@ class ImageClassification(TaskModel):
     ) -> None:
         """Exports the model to ONNX for inference.
 
-        The export uses a dummy input of shape (batch_size, C, H, W) where C is inferred
-        from the first model parameter and (H, W) come from `self.image_size`.
-        The ONNX graph uses dynamic batch size for both inputs and produces
-        two outputs: labels and scores.
+        The export uses a dummy input of shape (batch_size, C, H, W) where C is
+        inferred from the first model parameter and (H, W) come from
+        `self.image_size`. If `dynamic_batch_size` is True, the ONNX graph will
+        have a dynamic batch dimension for the input. The graph produces two
+        outputs: labels and scores.
 
         Optionally simplifies the exported model in-place using onnxslim and
         verifies numerical closeness against a float32 CPU reference via
@@ -392,6 +394,9 @@ class ImageClassification(TaskModel):
                 uses the model's current precision.
             batch_size:
                 Batch size for the ONNX input.
+            dynamic_batch_size:
+                If True, the ONNX graph will have a dynamic batch dimension for the
+                input. If False, the batch dimension is fixed to `batch_size`.
             height:
                 Height of the ONNX input. If None, will be taken from `self.image_size`.
             width:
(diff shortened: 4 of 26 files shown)


src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/task_model.py
```diff
@@ -703,6 +704,11 @@ def export_onnx(
         width = self.image_size[1] if width is None else width
         num_channels = len(self.image_normalize["mean"])
 
+        if dynamic_batch_size:
+            batch_size = 2
```
guarin, line 777: This will not work for most model as the code is not implemented for batch size > 1. Please check models and only enable for models that actually support it.

src/lightly_train/_task_models/dinov2_eomt_semantic_segmentation/task_model.py
```diff
@@ -634,6 +634,7 @@ def export_onnx(
         batch_size: int = 1,
         height: int | None = None,
         width: int | None = None,
+        dynamic_batch_size: bool = True,
```
guarin: Please update docstring in all models

guarin (dismissed):

JiwaniZakir (commented):
The `dynamic_batch_size=True` default silently overrides the caller-provided `batch_size` argument in `dinov2_eomt_semantic_segmentation`, `dinov3_eomt_instance_segmentation`, and `dinov3_eomt_semantic_segmentation` — if a user passes `batch_size=4` with the default `dynamic_batch_size=True`, it gets quietly replaced with `2`. A less surprising design would preserve the user's value and only enforce the minimum when needed (e.g., `batch_size = max(batch_size, 2) if dynamic_batch_size else batch_size`), or at minimum document the override in the docstring.

The `dinov3_eomt_panoptic_segmentation` change is inconsistent with the rest of the PR: it simply removes `"images": {0: "batch_size"}` from `dynamic_axes` without adding a `dynamic_batch_size` parameter or overriding the dummy input size to `2`. If the intent is the same fix, this file should receive the same treatment; if images intentionally have a fixed batch in panoptic segmentation, that reasoning should be explained in a comment.

In `dinov3_ltdetr_object_detection` and `picodet_object_detection`, the dummy input batch size is hardcoded to `2` with no corresponding `dynamic_batch_size` parameter, meaning there's no way for callers to opt out of dynamic batch sizing for these models, unlike the three files that do expose the parameter.

src/lightly_train/_task_models/picodet_object_detection/task_model.py
```diff
@@ -606,8 +613,12 @@ def export_onnx(
                         "num_channels must be provided for ONNX export if it cannot be inferred."
                     )
 
+        if dynamic_batch_size:
+            batch_size = 2
+        dynamic_axes = {"images": {0: "N"}} if dynamic_batch_size else None
```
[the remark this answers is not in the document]
simonschoelly: I think having a different batch size for verification is out of the scope for now.

src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/task_model.py
```diff
@@ -242,6 +242,7 @@ class IDs as defined in the `classes` argument of your dataset. These
 
     def forward(self, x: Tensor) -> tuple[Tensor, Tensor]:
         # Function used for ONNX export
+        # TODO (Simon, 05/26) This class does not seem to have an export_onnx function
```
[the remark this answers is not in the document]
simonschoelly: There is no agreed on TODO format.

tests/_task_models/picodet_object_detection/test_task_model.py
```diff
@@ -83,14 +83,72 @@ def test_export_onnx_has_no_nms(tmp_path: Path) -> None:
...
     op_types = {node.op_type for node in onnx_model.graph.node}
     assert "NonMaxSuppression" not in op_types
     assert "If" not in op_types
 
 
+@pytest.mark.skipif(not RequirementCache("onnx"), reason="onnx not installed")
+@pytest.mark.skipif(
+    not RequirementCache("onnxruntime"), reason="onnxruntime not installed"
+)
+def test_export_onnx__dynamic_batch_size(tmp_path: Path) -> None:
```
gabrielfruet, line 101: This test seems a bit heavy. How can we handle that? I'm not sure if we can mark this to run on the CI with GPUs instead of the GitHub one.

tests/_task_models/picodet_object_detection/test_task_model.py
```diff
@@ -83,14 +83,72 @@ def test_export_onnx_has_no_nms(tmp_path: Path) -> None:
...
+    assert close_boxes.float().mean() > 0.95
+    close_scores = torch.isclose(onnx_scores, scores, atol=2e-2, rtol=1e-1)
+    assert close_scores.float().mean() > 0.95
+
+
+@pytest.mark.skipif(not RequirementCache("onnx"), reason="onnx not installed")
+@pytest.mark.skipif(
+    not RequirementCache("onnxruntime"), reason="onnxruntime not installed"
+)
+def test_export_onnx__static_batch_size(tmp_path: Path) -> None:
```
gabrielfruet, line 141: Shouldn't we do the same checks as in `test_export_onnx__dynamic_batch_size`?

tests/_task_models/image_classification/test_task_model.py
```diff
@@ -0,0 +1,91 @@
...
+        backbone_freeze=False,
+        load_weights=False,
+    )
+
+
+@pytest.mark.skipif(not RequirementCache("onnx"), reason="onnx not installed")
+@pytest.mark.skipif(
+    not RequirementCache("onnxruntime"), reason="onnxruntime not installed"
+)
+def test_export_onnx__dynamic_batch_size(
```
gabrielfruet, line 37: Same questions and concerns as in PicoDet.

tests/_task_models/dinov3_ltdetr_object_detection/test_task_model.py
```diff
@@ -170,3 +174,63 @@ def test_get_optimizer__linear_warns_when_warmup_exceeds_training(
         train_model.get_optimizer(total_steps=1000, global_batch_size=16)
 
     assert "the schedule will not complete as intended" in caplog.text
+
+
+@pytest.mark.skipif(not RequirementCache("onnx"), reason="onnx not installed")
+@pytest.mark.skipif(
+    not RequirementCache("onnxruntime"), reason="onnxruntime not installed"
+)
+def test_export_onnx__dynamic_batch_size(tmp_path: Path) -> None:
```
gabrielfruet, line 183: Same questions and concerns as in PicoDet.

tests/_task_models/dinov2_eomt_instance_segmentation/test_task_model.py
```diff
@@ -0,0 +1,78 @@
...
+        num_joint_blocks=1,
+        load_weights=False,
+    )
+
+
+@pytest.mark.skipif(not RequirementCache("onnx"), reason="onnx not installed")
+@pytest.mark.skipif(
+    not RequirementCache("onnxruntime"), reason="onnxruntime not installed"
+)
+def test_export_onnx(model: DINOv2EoMTInstanceSegmentation, tmp_path: Path) -> None:
```
gabrielfruet: The name of the test here can be misleading. It tests the export when the `dynamic_batch_size` is `True`

src/lightly_train/_task_models/picodet_object_detection/task_model.py
```diff
@@ -528,6 +528,8 @@ def export_onnx(
         out: PathLike,
         *,
         precision: Literal["auto", "fp32", "fp16"] = "auto",
+        batch_size: int = 1,
```
gabrielfruet, line 531: Introducing `dynamic_batch_size` as default can cause breaking changes. I think backward compatibility is probably prefferable here.

src/lightly_train/_task_models/image_classification/task_model.py
```diff
@@ -366,6 +366,7 @@ def export_onnx(
         *,
         precision: Literal["auto", "fp32", "fp16"] = "auto",
         batch_size: int = 1,
+        dynamic_batch_size: bool = True,
```
gabrielfruet, line 369: Same for breaking changes.

src/lightly_train/_task_models/dinov2_eomt_panoptic_segmentation/task_model.py
```diff
@@ -1033,6 +1033,10 @@ def export_onnx(
             dtype=dtype,
         )
 
+        # Precalculate interpolated positional encoding for ONNX export.
```
gabrielfruet, line 1036: In this file, why we don't have the ability of exporting with static batch size?

tests/_task_models/dinov3_eomt_semantic_segmentation/test_task_model.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+@pytest.mark.xfail(
+    strict=True, reason="dinov3 ONNX export shape mismatch with static batch size"
+)
+@pytest.mark.skipif(not RequirementCache("onnx"), reason="onnx not installed")
+@pytest.mark.skipif(
+    not RequirementCache("onnxruntime"), reason="onnxruntime not installed"
+)
+def test_export_onnx__static_batch_size(
```
gabrielfruet, line 78: Do we an issue for these xfails tests?

gabrielfruet (changes requested):

tests/_task_models/picodet_object_detection/test_task_model.py
simonschoelly: This is already handled by the `verify=True` argument. We only need the additional checks for the dynamic batch sizes, because `verify=True` used the same batch size as the dummy tensor - so in that case I also want to test with a different batch size.

But it's a good point we should eventually improve the verify argument so it also checks with a different batch size.

src/lightly_train/_task_models/picodet_object_detection/task_model.py
simonschoelly: No, this model had dynamic batch size by default already, so it is not a breaking change.

I think we should make all exports non-dynamic though - the exported models are much better - but it is a breaking change and I can bring that up somewhen in the future.

src/lightly_train/_task_models/image_classification/task_model.py
simonschoelly: See above.

src/lightly_train/_task_models/dinov2_eomt_panoptic_segmentation/task_model.py
simonschoelly: We did not have that option before - I was mainly focused on fixing the existing functions. I have an find out issue this sprint to look closer what it is missing.

It could also be, that dynamic batches did not work for this model - PyTorch often struggles with dynamic dimensions.

tests/_task_models/dinov3_eomt_semantic_segmentation/test_task_model.py
simonschoelly: Yes, I created an issue for the current sprint.

tests/_task_models/picodet_object_detection/test_task_model.py
simonschoelly: I think the way to handle this is to have the tests run on our internal machine. There is a way to do that, but then the tests get executed only once a day or so.

All the tests in this file together run in < 4s on my laptop so I don't think it is an issue in this case.

tests/_task_models/image_classification/test_task_model.py
simonschoelly: See above

tests/_task_models/dinov3_ltdetr_object_detection/test_task_model.py
simonschoelly: See above

tests/_task_models/dinov2_eomt_instance_segmentation/test_task_model.py
simonschoelly: Renamed

src/lightly_train/_task_models/picodet_object_detection/task_model.py
gabrielfruet: Oh sorry, I forgot to look on the minus diff

gabrielfruet (approved):

tests/_task_models/picodet_object_detection/test_task_model.py
liopeer: We use this here: https://github.com/lightly-ai/lightly-train/blob/a19976e1c3c2716ede62689326898da1eda76c26/tests/_commands/test_train_task.py#L43

I think it's fine to leave them on the normal CI as well if they're <4s. But for the future we can put above line into the `conftest.py` and use it from there.

PR #597, opened 2026-02-03
No weight decay for bias/norm/token/etc.

guarin:
## What has changed and why?

* Skip weight decay for bias/norm/token/etc.
* Reduce number of logged lr/wd values

This PR updates all models except LTDETR for which I'll create a follow-up PR.

## How has it been tested?

* Unit tests
* Manual tests

- [x] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_optim/optimizer_helpers.py
+++ b/src/lightly_train/_optim/optimizer_helpers.py
@@ -115,7 +115,7 @@ def get_weight_decay_parameters(
         for name, param in module.named_parameters():
             param_id_to_info[id(param)] = {"name": name, "param": param, "module": None}
         for mod in module.modules():
-            for param in mod.parameters():
+            for param in mod.parameters(recurse=False):
                 param_id_to_info[id(param)]["module"] = mod
 
     params = []


src/lightly_train/_commands/train_task.py
```diff
@@ -941,10 +941,11 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
                     task=config.task,
                 )
                 for group in optimizer.param_groups:
-                    train_log_dict[f"learning_rate/{group['name']}"] = group["lr"]
-                    train_log_dict[f"weight_decay/{group['name']}"] = group[
-                        "weight_decay"
-                    ]
+                    if group.get("log", True):
```
guarin, line 947: Allows skipping lr/wd logs for models with many optimizer param groups

guarin:
> I was wondering whether we could put some of the helper into LightlySSL?

The weight decay / no weight decay helper function is in LightlySSL although not with all the functionality. We could definitely also move it there.

guarin:
@codex review

PR #603, opened 2026-02-06
CI: Remove python 3.12

guarin:
## What has changed and why?

* Remove python 3.12 from CI

## How has it been tested?

* CI

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
+++ b/src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
@@ -9,7 +9,6 @@ from __future__ import annotations
 
 import copy
 import math
-import re
 from typing import Any, ClassVar, Literal
 
 import torch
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -210,9 +210,8 @@ EXTRAS_PY38 := [dev,dicom,mlflow,notebook,onnx,super-gradients,tensorboard,timm,
 
 # SuperGradients is not compatible with Python>=3.10. It is also not easy to install
 # on MacOS. Therefore we exclude it from the default extras.
-# RFDETR has installation issues because of onnxsim dependency on CI with Python 3.12.
+# RFDETR has installation issues because of onnxsim dependency on CI with Python 3.13.
```
yutong-xiang-97, line 213: Not to be fixed by this PR, but is it still the case for Py3.13? Also I would assume we could update the RFDETR version to remove this altogether.

guarin: RFDETR version was already updated. Might not be relevant anymore

yutong-xiang-97 (approved):

PR #608, opened 2026-02-10
Add Image Classification Implementation

guarin:
## What has changed and why?

* Add multiclass image classification
* Add multilabel image classification

Adds support for fine-tuning any model supported by LightlyTrain for classification.

Will make a follow-up PR with docs and changelog updates. The current PR is for testing. 

## How has it been tested?

* Unit tests
* Manually

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_task_models/image_classification/task_model.py
+++ b/src/lightly_train/_task_models/image_classification/task_model.py
@@ -155,7 +155,14 @@ class ImageClassification(TaskModel):
 
     @classmethod
     def is_supported_model(cls, model: str) -> bool:
-        # This class works with any model.
+        try:
+            package_name, _ = package_helpers.parse_model_name(model=model)
+        except ValueError:
+            return False
+        try:
+            package_helpers.get_package(package_name)
+        except ValueError:
+            return False
         return True
 
     @classmethod


yutong-xiang-97:
@codex review

src/lightly_train/_task_models/image_classification/task_model.py
```diff
@@ -0,0 +1,672 @@
...
+        )
+
+    @classmethod
+    def list_model_names(cls) -> list[str]:
+        return [
+            f"{name}-{cls.model_suffix}" for name in package_helpers.list_model_names()
+        ]
+
+    @classmethod
+    def is_supported_model(cls, model: str) -> bool:
```
yutong-xiang-97, line 157: From Codex:

ImageClassification.is_supported_model() currently returns True for every string, which prevents load_checkpoint() from treating model as a checkpoint path when users pass [exported_last.pt"](https://file+.vscode-resource.vscode-cdn.net/Users/yutong/.vscode/extensions/openai.chatgpt-0.4.74-darwin-arm64/webview/#) (a documented workflow). In that case the code skips checkpoint loading and later tries to parse the path as a package/model identifier, causing training startup to fail instead of fine-tuning from the checkpoint.

Issue #510, opened 2025-12-16
[FEAT] Support for OBB detection


swdmike:
Hi team,

First of all, thank you for open-sourcing such an amazing framework. LightlyTrain is by far the best framework I’ve used in terms of both training speed and performance.

I’m working on an object detection task where oriented bounding boxes (OBB) are essential, and I would like to request native support for OBB detection.

Thanks again!

stegmuel:
Thank you for the kind words.
We've added this feature to our list and will keep you updated once we are getting started.

roboyul:
Would love to see OBB asap. Most real world use cases will benefit from this highly.

gabrielfruet:
@guarin @stegmuel 

Do we expect any specific OBB Head to be implemented (RFCOS, RTMDet, YOLO26...) ? Or integrating from another library (e.g. transformers)

I plan myself to create a PR for that.

guarin:
We didn't decide on the head architecture yet. We also have to look into data loading first:
* Which OBB data format to support
* Augmentations support for OBB in albumentations

gabrielfruet:
https://github.com/albumentations-team/AlbumentationsX

AlbumentationsX apparently supports OBB

Some examples here:

https://albumentations.ai/docs/api-reference/albumentations/augmentations/geometric/functional/#bboxes_rot90

gabrielfruet:
I think 4 corners, like Ultralytics framework, is the best.

Representing internally as xywhr is okay, and the standard i think. But in the txts it can cause ambiguity for 90° shifts

XYWHR (0,0,10,5,0°) == (0,0,5,10,90°)

So i suggest XYXYXYXY format.

swdmike:
"class_index x1 y1 x2 y2 x3 y3 x4 y4" is great.

yobar08:
Great to see that OBB support is planned! Format like YOLO-OBB is easy to use."

Scott143-dot:
Really happy to see OBB support—thanks so much for all the hard work!

guarin:
> AlbumentationsX apparently supports OBB

I don't think we can use AlbumentationsX because of the commercial license. Will investigate.

> So i suggest XYXYXYXY format.

👍🏼

gabrielfruet:
[torchvision rotated transforms](https://docs.pytorch.org/vision/main/auto_examples/transforms/plot_rotated_box_transforms.html)

I think albumentations don't support OBB, but torchvision does.

[...]

PR #616, opened 2026-02-20
Oriented Object Detection Tranforms, collation and types.

gabrielfruet:
## What has changed and why?

This pull request introduces support for oriented object detection transforms in the data processing pipeline, to address #510. It adds a new transform class for handling oriented bounding boxes (OBBs), utilities to dispatch torchvision transforms for OBBs, updates the dataset item types to include OBBs, and provides comprehensive tests for the new functionality.

The most important changes are:

### Oriented Object Detection Transform

* Added `OrientedObjectDetectionTransform`, its input/output types, and argument class to handle images and oriented bounding boxes, supporting various augmentations and normalization steps.

### Torchvision Dispatcher for Oriented Bounding Boxes

* Introduced new utilities and dispatcher classes in `torchvision_dispatcher.py` to convert between numpy and torchvision tensor formats for images and OBBs, and to apply torchvision transforms (e.g., rotation, flipping, resizing) to oriented bounding boxes.

### Type Definitions

* Added `OrientedObjectDetectionDatasetItem` to the types module to represent dataset items with oriented bounding boxes and class labels.

## Unsupported operations

* Some operations are currently not supported by `torchvision`  for OBB. An example is RandomIoUCrop, since `torchvision` is starting to support rotated iou now (funny, rotated iou got merged 12h ago https://github.com/pytorch/vision/pull/9379 haha)

## Notes about dispatching over migrating to `torchvision` completely

I choose dispatching augmentations in a `albumentations` compatible API to keep a friendly API for the `TransformsArgs` (e.g. `RandomZoomOutArgs`). 

Since the `torchvision` API is a bit different and has different naming as the `albumentations` API, i preferred to create a dispatcher, such that we can add custom logic, warnings and etc for each augmentation.

# Testing

* Added comprehensive tests for the new oriented object detection transform, covering all combinations of augmentations and verifying collation logic.
* Reused same testing pattern as the `ObjectDetectionTransform`.
* Added tests for the `TorchVisionDispatcher` mechanism.

- [ ] Yes

- [ ] Yes


guarin:
/review


Thanks a lot for the PR! Sorry that I didn't already provide feedback in the issue. I think what we should try with this is to write the transform and collate only using torchvision and completely replace albumentations for OBB. This way the code becomes simpler and we don't have to worry about switching between torch and numpy all the time. It will also give us a good indication of the state of torchvision transforms. When we started out torchvision transforms v2 was still quite recent and didn't support all the operations we needed. Now it seems more mature and it would be a good moment to check if anything is missing. If not it would make sense to eventually move the augmentations for all tasks to torchvision.

src/lightly_train/_data/task_batch_collation.py
```diff
@@ -219,3 +236,83 @@ def __call__(self, batch: list[ObjectDetectionDatasetItem]) -> ObjectDetectionBa
...
+            )
+
+    def __call__(
+        self, batch: list[OrientedObjectDetectionDatasetItem]
+    ) -> OrientedObjectDetectionBatch:
+        if self.scale_jitter is not None:
+            batch_np = [
+                {
+                    "image_path": item["image_path"],
+                    "image": item["image"].permute(1, 2, 0).numpy(),
```
guarin: Would prefer to keep everything as torch Tensors if we are using torchvision transforms. This will simplify the code and make things faster.

src/lightly_train/_transforms/torchvision_dispatcher.py
```diff
@@ -0,0 +1,307 @@
...
+
+CV2_INTERPOLATION_MODE = int
+
+CV2_TO_TV_INTERPOLATION: Mapping[CV2_INTERPOLATION_MODE, InterpolationMode] = {
+    cv2.INTER_NEAREST: InterpolationMode.NEAREST,
+    cv2.INTER_LINEAR: InterpolationMode.BILINEAR,
+    cv2.INTER_CUBIC: InterpolationMode.BICUBIC,
+    cv2.INTER_AREA: InterpolationMode.BOX,
+    cv2.INTER_LANCZOS4: InterpolationMode.LANCZOS,
+}
```
guarin: Caveat: We use `cv2.INTER_AREA` by default as replacement for `InterpolationMode.BILINEAR` as it gives better results.

src/lightly_train/_transforms/oriented_object_detection_transform.py
```diff
@@ -0,0 +1,218 @@
...
+
+class OrientedObjectDetectionTransformArgs(ObjectDetectionTransformArgs):
+    pass
+
+
+class OrientedObjectDetectionTransform(TaskTransform):
+    transform_args_cls: type[OrientedObjectDetectionTransformArgs] = (
+        OrientedObjectDetectionTransformArgs
+    )
+
```
guarin, line 51: Is it possible to rewrite this transform to only use torchvision transforms without albumentations and dispatching? I think this would be preferred as it keeps a clear separation between torchvision and albumentations and makes the code simpler.

Long term we might move away from albumentations anyways due to the license change.

In the dataset class we can use `file_helpers.open_image_tensor` instead of `file_helpers.open_image_numpy` to directly read the image as a tensor. This will keep the whole pipeline in tensors and make it fast. With dispatch we have to constantly switch between numpy/tensors which creates unnecessary overhead.

guarin (changes requested):

src/lightly_train/_transforms/torchvision_dispatcher.py
gabrielfruet: Unfortunately v2 Rotate don't support INTER_AREA :(

guarin: BILINEAR is preferred :) The default is usually BILINEAR but because it is not supported in cv2 we use INTER_AREA instead of it because we saw superior results with it.

src/lightly_train/_transforms/oriented_object_detection_transform.py
gabrielfruet: That's fair and i agree with you. Dispatching introduce many overheads. I only opted for it because i didn't know if relying 100% on `torchvision` was an option, also wanted to keep a compatible API with the transform args. 

Will rewrite it using only `torchvision`. Might take 1-2 days

src/lightly_train/_transforms/torchvision_dispatcher.py
gabrielfruet: I'll implement your suggestion! I think I should also convert INTER_AREA to BILINEAR by default internally for the oriented case. Since the current default is INTER_AREA, it would otherwise cause a crash in the current implementation, this change will ensure a more seamless experience for the end user.

src/lightly_train/_data/task_batch_collation.py
gabrielfruet: That's fair and desirable! I'll address this also.

src/lightly_train/_data/task_batch_collation.py
```diff
@@ -219,3 +236,43 @@ def __call__(self, batch: list[ObjectDetectionDatasetItem]) -> ObjectDetectionBa
...
+                scale_range=scale_range,
+            )
+        else:
+            self.scale_jitter = None
+
+    def __call__(
+        self, batch: list[OrientedObjectDetectionDatasetItem]
+    ) -> OrientedObjectDetectionBatch:
+        if self.scale_jitter is not None:
+            batch = self.scale_jitter(batch)
```
gabrielfruet, line 265: This syntax is a very nice feature in favor of `torchvision`.

`torchvision.transforms.v2` uses pytree flattening, so we can pass lists, dicts etc, and it will use the same parameter across all of the tranforms. 

The dispatch mechanism is based on the type of the variables, `tv_tensors.BoundingBoxes, tv_tensors.Image`.

This is a significant improvement over `albumentations`, since we don't need manual seed and state manipulation to achieve this behavior.

src/lightly_train/_data/task_batch_collation.py
```diff
@@ -114,24 +122,33 @@ def __call__(
         return out
 
 
+def _get_scale_range_from_transform_args(
```
gabrielfruet, line 125: Extracted this since it's basically the same logic for both Oriented and Axis aligned case.

src/lightly_train/_transforms/oriented_object_detection_transform.py
```diff
@@ -0,0 +1,195 @@
...
+from lightly_train._transforms.random_rotate_90 import RandomRotate90
+from lightly_train._transforms.task_transform import (
+    TaskTransform,
+    TaskTransformInput,
+    TaskTransformOutput,
+)
+
+
+class OrientedObjectDetectionTransformInput(TaskTransformInput):
+    image: tv_tensors.Image  # CHW
```
gabrielfruet, line 30: Per Guarin's suggestion, we rely solely on torch tensors, since `torchvision` don't rely on numpy.

src/lightly_train/_transforms/scale_jitter.py
```diff
@@ -12,10 +12,50 @@
 import numpy as np
 from albumentations import Resize
 from albumentations.core.transforms_interface import DualTransform
+from torchvision.transforms import v2
 
 from lightly_train.types import NDArrayBBoxes, NDArrayImage, NDArrayMask
 
 
+def generate_discrete_sizes(
```
gabrielfruet, line 20: Extracted this functions since we have the same behavior in both scale jitter transforms.

src/lightly_train/_transforms/transform.py
```diff
@@ -190,7 +191,7 @@ class ScaleJitterArgs(PydanticConfig):
 
 class StopPolicyArgs(PydanticConfig):
     stop_step: int
-    ops: Set[type[BasicTransform]]
+    ops: Set[type[BasicTransform] | type[v2.Transform]]
```
gabrielfruet, line 194: Given the support for `v2.Transforms`, the stop policy should also feature that.

src/lightly_train/_transforms/channel_drop.py
```diff
@@ -98,3 +101,75 @@ def apply(self, img: NDArrayImage, **params: dict[str, Any]) -> NDArrayImage:
...
+        channels_to_keep = torch.sort(all_channels[~mask])[0]
+        return {"channels_to_keep": channels_to_keep}
+
+    def transform(self, inpt: Any, params: dict[str, Any]) -> Any:
+        if isinstance(inpt, tv_tensors.Image):
+            if inpt.ndim < 3:
+                return inpt
+            channels_to_keep = params["channels_to_keep"]
+            if channels_to_keep is None:
+                return inpt
```
guarin: This case is impossible

src/lightly_train/_transforms/channel_drop.py
```diff
@@ -98,3 +101,75 @@ def apply(self, img: NDArrayImage, **params: dict[str, Any]) -> NDArrayImage:
...
+            if inpt.shape[0] != len(self.weight_drop):
+                return inpt
```
guarin: This must raise an error as in the albumentations version. Please use the same errors as there.

src/lightly_train/_transforms/oriented_object_detection_transform.py
```diff
@@ -0,0 +1,184 @@
...
+    bboxes: tv_tensors.BoundingBoxes  # N x 5, where each bbox is (cx, cy, w, h, angle)
+    class_labels: Tensor  # N
```
guarin: Bboxes and classes must be `NotRequired`

src/lightly_train/_transforms/oriented_object_detection_transform.py
```diff
@@ -0,0 +1,184 @@
...
+                    side_range=transform_args.random_zoom_out.side_range,
+                    p=transform_args.random_zoom_out.prob,
+                )
+            )
+
+        if transform_args.random_iou_crop is not None:
+            raise NotImplementedError(
+                "RandomIoUCrop is not implemented yet for OrientedObjectDetectionTransform."
+                "torchvision does not support it for now."
+            )
```
guarin, line 103: It exists in torchvision: https://docs.pytorch.org/vision/main/generated/torchvision.transforms.v2.RandomIoUCrop.html

Is the problem that it doesn't support OBB?

src/lightly_train/_transforms/random_rotate_90.py
```diff
@@ -0,0 +1,40 @@
...
+                        v2.RandomRotation(degrees=[270, 270]),
+                        v2.Identity(),
+                    ]
+                )
+            ],
+            p=p,
+        )
+
+    def transform(self, inpt: Any, params: dict[str, Any]) -> Any:
+        return self._transform(inpt)
```
guarin, line 40: Don't you have to pass the params also to `self._transform`?

src/lightly_train/_transforms/scale_jitter.py
```diff
@@ -106,3 +121,42 @@ def apply_to_bboxes(
...
+
+        sizes_list = generate_discrete_sizes(
+            sizes=sizes,
+            target_size=target_size,
+            scale_range=scale_range,
+            num_scales=num_scales,
+            divisible_by=divisible_by,
+        )
+
+        transforms = [v2.Resize(size=(h, w), antialias=True) for h, w in sizes_list]
```
guarin: Please use bicubic interpolation for the resize.

tests/_transforms/test_oriented_object_detection_transform.py
```diff
@@ -0,0 +1,251 @@
...
+    # TODO: Lionel (09/25) Add StopPolicyArgs test cases.
+    [None, _get_scale_jitter_args()],
+    [None, _get_resize_args()],
+    [None, _get_normalize_args()],
+)
+
+possible_tuples = list(itertools.product(*PossibleArgsTuple))
+
+
+class TestObjectDetectionTransform:
```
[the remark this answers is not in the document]
guarin: Please update

src/lightly_train/_transforms/oriented_object_detection_transform.py
gabrielfruet: Yep, they convert to axis aligned and then it crashes because we cant convert back to OBB.

I'm willing to create a PR to fix that in `torchvision` when i manage to get some free time.

gabrielfruet:
Done!

guarin:
> Done!


Sorry, I wasn't clear. Could you answer the open comments? Not resolve them. In general we treat comments as "owned" by the person who created the comment. It is their responsibility to resolve them. Leaving them for the person to resolve also allows the person to easily check whether their comment was addressed.

gabrielfruet:
@guarin I don't see any open comments other than some comments i made myself to clarify some parts of the code. Could you please point out what you're referring to?

src/lightly_train/_transforms/oriented_object_detection_transform.py
guarin: @gabrielfruet

src/lightly_train/_transforms/random_rotate_90.py
guarin: @gabrielfruet

guarin:
Pinged you in the relevant comments:

Screenshot 2026-02-27 at 09 36 17

src/lightly_train/_transforms/random_rotate_90.py
gabrielfruet: The params are generated by the transform class responsible for transforming, so each transform class has its own parameters. In the case of this class, we don't generate any params in the class itself, but the transforms under _transforms generate their own params and apply them. In fact, params here is an empty dict :)

Basically, no, we don't need.

gabrielfruet:
@guarin Oh, that's weird :) I replied to those on the same day you commented.

For some reason my answer was "pending" (?) according to the GH UI. Anyway, just copied and pasted the "pending" answer and it worked!

image

guarin:
Thanks! I can't see the answer for the first question

gabrielfruet:
Basically, torchvision doesn't support RandomIoUCrop for oriented bounding boxes.

I'm willing to open a PR to address this on `torchvision` once i get some free time :)

src/lightly_train/_transforms/scale_jitter.py
```diff
@@ -106,3 +121,47 @@ def apply_to_bboxes(
...
+            v2.Resize(
+                size=(h, w), interpolation=InterpolationMode.BICUBIC, antialias=True
+            )
+            for h, w in sizes_list
+        ]
+
+        self._transform = v2.RandomChoice(transforms)
+
+    def forward(self, *inputs: Any) -> Any:
+        return self._transform(*inputs)
```
gabrielfruet: Crucial fix:

We were using the transform method, it was applying a different choice for each input. The right way to do is overriding the forward, to leverage torchvision pytree flattening and unflattening.

guarin:
@gabrielfruet is this ready for review? I am asking because tests are failing

gabrielfruet:
@guarin some questions:

1. I think that the pinned torchvision version for Python 3.8 don't support Oriented Bounding Box 

https://github.com/lightly-ai/lightly-train/blob/451b0016a944c89fd8da446e478f88e4868f3e1c/Makefile#L267

2. There are some type hints (like set subscription) that are not supported in Python 3.8 apparently.

> This one is easy to solve. The one that may be a problem is the `torchvision` support for oriented bounding boxes.

I have to address these issues, i'm having some trouble to install Lightly-train with python 38 on my machine due to the onnxruntime dependency. Tried a slim install, but still failed on some pre-commit dependencies.

I'm trying to install it on my machine with python 3.8, but while i can't, i cannot debug some of these issues.

Are we still supporting python 3.8 although its EOL?

gabrielfruet:
This is the error message i'm facing.

```
❯ make install-dev
uv pip install -e ".[dev,dicom,mlflow,notebook,onnx,onnxruntime,onnxslim,rfdetr,tensorboard,timm,ultralytics,wandb]"
  × No solution found when resolving dependencies:
  ╰─▶ Because only the following versions of onnxruntime are available:
          onnxruntime<=1.21.0
          onnxruntime==1.21.1
          onnxruntime==1.22.0
          onnxruntime==1.22.1
          onnxruntime==1.23.0
          onnxruntime==1.23.1
          onnxruntime==1.23.2
          onnxruntime==1.24.1
          onnxruntime==1.24.2
      and onnxruntime>=1.21.0 has no wheels with a matching Python implementation tag (e.g., `cp38`), we can conclude that onnxruntime>=1.21.0 cannot be used.
      And because lightly-train[onnxruntime]==0.14.2 depends on onnxruntime>=1.21.0, we can conclude that lightly-train[onnxruntime]==0.14.2 cannot be used.
      And because only lightly-train[onnxruntime]==0.14.2 is available and you require lightly-train[onnxruntime], we can conclude that your requirements are unsatisfiable.

      hint: Pre-releases are available for `onnxruntime` in the requested range (e.g., 1.24.0.dev20251031003), but pre-releases weren't enabled (try: `--prerelease=allow`)

      hint: You require CPython 3.8 (`cp38`), but we only found wheels for `onnxruntime` (v1.24.2) with the following Python implementation tags: `cp311`, `cp312`, `cp313`, `cp314`
make: *** [Makefile:292: install-dev] Error 1

```

gabrielfruet:
1. torchvision at the current py38 version was in BETA. there are some conflicts with the current API. 

For example:

They apparently used `_transform` as the method that should be overwritten instead of the `transform`.

The pinned `torchvision` version for py38 definetely don't support OBB. Already checked it. So i think it's not worth spending time modifying the API to be BC with this `torchvision` version.

Apparently `torchvision` starts to support OBB at [torchvision 0.23](https://github.com/pytorch/vision/releases/tag/v0.23.0)


We should also update the 313 version, since it's 0.22 now.
https://github.com/lightly-ai/lightly-train/blob/451b0016a944c89fd8da446e478f88e4868f3e1c/Makefile#L276

Should we update this dependency?

guarin:
Ah yes Python 3.8 support is a pain. We still allow it for backwards compatibility but only for pretraining. So for fine-tuning you can assume Python > 3.8.

What you can do is add a check for the torchvision version and raise an error with a message that at least version X of torchvision must be installed for OBB support. The simplest way to do this is to check torchvision version with `RequirementCache` (you can search the codebase for example usage) and raise the error in the `__init__` of the transform.

If you have tests you can skip them with:
```
@pytest.mark.skipif(not RequirementCache("torchvision>=X"), "torchvision too old for OBB support")
def test_xyz()
    ...
```

Hope that helps. Otherwise happy to take over the PR and handle it.

gabrielfruet:
I can handle it! Would help me to get in touch with the codebase.

One quick question: Should i put the default `torchvision` version for 0.23 for Py313 or not? Since it's the minimum version that supports OBB?

guarin:
> One quick question: Should i put the default torchvision version for 0.23 for Py313 or not? Since it's the minimum version that supports OBB?

Would prefer to not change it. We can make a follow-up PR to update dependencies for tests. There are always some unexpected things breaking when updating deps :)

gabrielfruet:
@guarin Tests are green here both for python 3.8 and for python 3.13. tested on both.

i'm now skipping tests for OBB based on torchvision version. i also fixed some type hints that were not compatible with python 3.8.

tests/_transforms/test_channel_drop.py
```diff
@@ -82,6 +83,10 @@ def test__call__(
             )
 
 
+@pytest.mark.skipif(
+    not RequirementCache("torchvision>=0.23"),
+    reason="torchvision too old for oriented bounding box suport",
```
guarin: Should we change the reason here? Because this doesn't test OBB

tests/_transforms/test_random_rotate_90.py
```diff
@@ -19,6 +20,10 @@ def dummy_tv_image() -> tv_tensors.Image:
     return tv_tensors.Image(torch.rand(3, 64, 64))
 
 
+@pytest.mark.skipif(
+    not RequirementCache("torchvision>=0.23"),
+    reason="torchvision too old for oriented bounding box suport",
```
guarin: Same as above

gabrielfruet:
> LGTM! There are still some typing errors but you can add `# type: ignore` to suppress them.

1. Added `# type: ignore` at the class declaration of the transforms
2. Removed the OBB related part of the error message, leaving only that torchvision is too old.

guarin (approved):

Issue #615, opened 2026-02-20
[FEAT] Frozen Backbone Finetuning


Jaster1999:
Is it possible to train models (e.g. ltdetr) with dinov3 backbones but keep the features frozen for the backbone (and turn off gradients). In the Dinov3 paper they report good accuracy with frozen features and it would help with keeping vram requirements low for larger models.

### 🧰 Describe the solution you'd like
lightly_train.train_object_detection(
        out="out/my_experiment",
        model="dinov3/vitt16-ltdetr-coco",
        forzen_backbone=True,                        <-------- add something like
        data={
            "path": "my_data_dir",
            "train": "images/train",
            "val": "images/val",
            "names": {
                0: "person",
                1: "bicycle",
                2: "car",
                # ...
            },
        },
    )

guarin:
Hi, thanks for the suggestion. Happy to add this! It should be relatively straight forward as we already have all the logic in place for some other models.

That being said, this will probably only work well when fine-tuning models that were already trained on an object detection task (e.g. COCO).

Jaster1999:
My specific interest here is frozen finetuning the dinov3 backbones, due to their great dense features I was figuring that it may still work well (they talk about this in the dino paper when they finetune a plain detr on top of the frozen dino encoder they get very good results (albeit is the Dinov3 7B params model).

I'm also wondering whether the performance may be worse than a full finetune but the generalization to new slightly out of domain data would be better. Anyways awesome to hear it can be implemented!

guarin:
This was released on 0.14.2

PR #618, opened 2026-02-20
Object Detection backbone freezing option

gabrielfruet:
## What has changed and why?

#615

This pull request adds support for freezing the backbone weights during training for both DINOv2 and DINOv3 object detection models. The change is controlled via a new `backbone_freeze` argument, and includes updates to the training logic and new tests to ensure correct behavior.

**Backbone freezing functionality:**

* Added a `freeze_backbone` method to both `DINOv2LTDETRObjectDetectionTaskModel`, `DINOv3LTDETRObjectDetectionTaskModel` and `PicoDetObjectDetectionTaskModel` which sets the backbone to evaluation mode and disables gradient computation for its parameters.
* Introduced the `backbone_freeze` argument to `DINOv2LTDETRObjectDetectionTrainArgs`, `DINOv3LTDETRObjectDetectionTrainArgs` and `PicoDetObjectDetectionTrainArgs` allowing users to specify whether the backbone should be frozen during training. 
* Modified the `set_train_mode` method in both train model classes to call `freeze_backbone` if `backbone_freeze` is set to `True`. 

**Code quality improvements**

* Standardized the backbone freezing for all models that supports backbone freezing using PyTorch current recommended approach, using Module.requires_grad_

## How has it been tested?

* Added tests in `test_task_model.py` for both DINOv2 and DINOv3 to verify that the backbone freezing logic works as expected, including parameterization for both frozen and unfrozen cases.

- [x] Yes

- [x] Yes

### DIFF
--- a/src/lightly_train/_task_models/picodet_object_detection/task_model.py
+++ b/src/lightly_train/_task_models/picodet_object_detection/task_model.py
@@ -178,7 +178,7 @@ class PicoDetObjectDetection(TaskModel):
         """Check if a model name is supported."""
         return model in _MODEL_CONFIGS
 
-    def freeze_backbone(self):
+    def freeze_backbone(self) -> None:
         self.backbone.eval()
         self.backbone.requires_grad_(False)
 


src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/task_model.py
```diff
@@ -395,5 +395,4 @@ def load_train_state_dict(self, state_dict: dict[str, Any]) -> None:
 
     def freeze_backbone(self) -> None:
         self.backbone.eval()
-        for param in self.backbone.parameters():
-            param.requires_grad = False
+        self.backbone.requires_grad_(False)
```
guarin, line 397: Thanks for the improvement!

gabrielfruet:
Sure! I can do it.

Just one question that i would like to know your opinion.

Is it desirable to replicate the freezing behavior in two places? shouldn't we keep this either on the TrainModel or TaskModel? Other than BC, I don't know what value we get from adding this to the TaskModel. 

It makes more sense to me the train model being responsible for this, since freezing or unfreezing is related to the training, not task.

I can do either, just let me know what do you think!

src/lightly_train/_task_models/dinov2_linear_semantic_segmentation/task_model.py
gabrielfruet: Thank you!

gabrielfruet:
@liopeer I saw your reaction to my message. Did you mean we should go forward with adding freezing to each TaskModel, or should we leave it out?

guarin:
> @liopeer I saw your reaction to my message. Did you mean we should go forward with adding freezing to each TaskModel, or should we leave it out?

I would prefer to keep freezing for each task model individually. There are always exceptions where the default will not work well and keeping it in the individual task model forces you to adjust it when implementing the model. As an example, we could also add backbone freezing to EoMT, but EoMT injects query parameters in the middle of the backbone. So backbone freezing for EoMT should actually only freeze the layers before the query injection.

> Is it desirable to replicate the freezing behavior in two places? shouldn't we keep this either on the TrainModel or TaskModel? Other than BC, I don't know what value we get from adding this to the TaskModel.

Yes this is desirable, the idea is that the TaskModel controls the model structure and the TrainModel controls the training (model + metrics + loss + train/eval). Both have to know about freezing. Also, people can fine-tune the TaskModel in their own pipeline so it makes sense to expose backbone freezing from the TaskModel as well.

gabrielfruet:
Clear now, thanks for the patience! I’ll get this implemented too.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/task_model.py
```diff
@@ -320,6 +320,7 @@ def __init__(
         classes: dict[int, str],
         image_size: tuple[int, int],
         image_normalize: dict[str, Any] | None = None,
+        backbone_freeze: bool = True,
```
guarin: This should be False by default because until now it was not possible to freeze the backbone

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/task_model.py
```diff
@@ -730,6 +739,7 @@ def __init__(
...
```
guarin: Same here

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
```diff
@@ -451,6 +451,7 @@ def __init__(
         classes: dict[int, str],
         image_size: tuple[int, int],
         image_normalize: dict[str, Any] | None = None,
+        backbone_freeze: bool = True,
```
guarin: Same here

src/lightly_train/_task_models/semantic_segmentation_multihead/task_model.py
```diff
@@ -55,15 +55,15 @@ def __init__(
...
+                 List of head names. One segmentation head is created for each name.
                 Head names should follow the format `head_lr{value}` where value is the
                 learning rate formatted without trailing zeros and dots replaced with
                 underscores (e.g., "head_lr0_001" for lr=0.001).
             class_ignore_index:
                 The class index to ignore during training. If provided, an additional
                 class output is created for this ignored class.
             image_size:
-                The size of the input images as (height, width). Default (518, 518).
+                The size of the input images.
```
guarin: Not sure why this changed?

guarin (changes requested):

src/lightly_train/_task_models/semantic_segmentation_multihead/task_model.py
gabrielfruet: I think it was an erroneous tab-autocomplete

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/task_model.py
gabrielfruet: Great catch! leaving that `True` was a slip here.

src/lightly_train/_task_models/image_classification_multihead/train_model.py
```diff
@@ -455,6 +455,7 @@ def validation_step(
 
     def set_train_mode(self) -> None:
         self.train()
+        # backbone is always frozen for multihead training
```
gabrielfruet: Added a comment to clarify why multi-head is frozen by default.

guarin:
PR looks good! If you can resolve the merge conflicts we can go ahead and merge it

gabrielfruet:
@guarin Done!

straint1:
Hi, I've been training an object detection model with LT-DETR and DinoV3 backend with the freeze_backbone option. It appears to have broken the validation functionality, giving small MAP scores (0.001). The MAP scores are fine when finetuning the full model. It could very well be a mistake I've made but thought I would sense check this with you.

gabrielfruet:
Hi @straint1 , can you provide the code snippet where you're facing this issue? I'll take a look. Also, if you're using a public dataset, let me know!

straint1:
> Hi @straint1 , can you provide the code snippet where you're facing this issue? I'll take a look. Also, if you're using a public dataset, let me know!

I am running the below - unfortunately the dataset is not publically available, but works fine with ultralytics and with lightly-train before I reinstalled this afternoon following this PR. 

```
names = {int(num) - 1: class_name for (num, class_name) in class_mapping.items()}

    # Fine-tune DINOv3 with DETR for object detection.
    lightly_train.train_object_detection(
        out=out_path,
        overwrite=True,
        model="dinov3/vits16-ltdetr-coco",
        data={
            "path": "/home/ec2-user/data/Maritime_Dataset",
            "train": "images/train",
            "val": "images/val",
            "names": names,
        },
        batch_size=16,
        steps=steps,
        model_args={"backbone_freeze": True},
    )
```

Screenshot 2026-02-24 195518

Here is what is outputted, the training loss decreases as before, (much smoother with the backend frozen btw) but the validation loss is huge (not shown) and low MAP scores. The validation and training sets are from the same dataset so shouldn't diverge by an order of magnitude.

guarin:
Interesting, I haven't observed this. I just tested it with a `dinov3/vits16-ltdetr` model (not pretrained on COCO) and it converged quite nicely:

Screenshot 2026-02-25 at 08 30 25

PR #627, opened 2026-02-24
Gradient Accumulation for Task Training

liopeer:
## What has changed and why?
- adds gradient accumulation to task training (includes automatic LR scaling when non-divisible)
- still missing documentation

## How has it been tested?
- passes existing unit tests (no additional needed)

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/train_task.py
+++ b/src/lightly_train/_commands/train_task.py
@@ -1549,6 +1549,7 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
                             log_dict=val_log_dict,
                             split="val",
                             global_batch_size=config.batch_size,
+                            gradient_accumulation_steps=config.gradient_accumulation_steps,
                         )
                         fabric.log_dict(val_log_dict, step=step)
                         helpers.reset_metrics(val_result.log_dict)


src/lightly_train/_commands/train_task.py
```diff
@@ -189,6 +190,12 @@ def train_image_classification(
         save_checkpoint_args:
             Arguments to configure the saving of checkpoints. The checkpoint frequency
             can be set with ``save_checkpoint_args={"save_every_num_steps": 100}``.
+        gradient_accumulation_steps:
+            Number of gradient accumulation steps. 'auto' automatically enables
+            gradient accumulation when batch_size is smaller than the model's default
+            batch size, using ``max(1, default_batch_size // batch_size)`` steps to
+            keep the effective batch size and learning rate close to the model defaults.
+            Set to 1 to disable gradient accumulation.
```
liopeer:
```suggestion
            Set to 1 to explicitly disable gradient accumulation.
```
Also add to the other docstrings.

src/lightly_train/_commands/train_task.py
```diff
@@ -1232,7 +1291,7 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
         train_model.set_train_mode()
         optimizer, scheduler = train_model.get_optimizer(
             total_steps=config.steps,
-            global_batch_size=config.batch_size,
+            global_batch_size=effective_global_batch_size,
```
liopeer, line 1294: This will use the auto-scaling we already have inside the models.

src/lightly_train/_commands/train_task.py
```diff
@@ -1349,18 +1412,23 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
...
+                timer.end_step("train_dataload")
 
-            # Training backward pass, optimizer step, and scheduler step.
-            fabric.backward(train_result.loss)
+                with fabric.no_backward_sync(train_model, enabled=is_accumulating):
+                    train_result = train_model.training_step(
+                        fabric=fabric, batch=batch, step=step
+                    )
+                    fabric.backward(
+                        train_result.loss / config.gradient_accumulation_steps  # type: ignore[operator]
```
liopeer: Still have to check if they don't somehow already automatically scale the loss.

src/lightly_train/_commands/train_task.py
liopeer, line 1447: TBD if we want to have the step timing inside or outside the gradient accumulation. But I think the way it is now makes more sense.

src/lightly_train/_commands/train_task.py
liopeer: They don't track how many iterations we do, so scaling is appropriate. I was just confused because they don't do it in the examples in their docs.

liopeer:
@codex review

src/lightly_train/_commands/train_task.py
guarin: Yes I agree that the current implementation makes sense 👍🏼

src/lightly_train/_commands/train_task.py
```diff
@@ -1383,13 +1453,13 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
                     log_dict=train_log_dict,
                     task=config.task,
                     timer_agg=timer_agg,
-                    global_batch_size=config.batch_size,
+                    global_batch_size=effective_global_batch_size,
```
guarin, line 1461: I believe there is now a small calculation error in `log_step`, `add_timer_logs`, and `log_training_summary`. The data time is calculated as `total_data_time / data_steps` but we make `gradient_accumulation_steps` many data steps in each effective training step. So I think we should report the final per step time as `total_data_time / data_steps * gradient_accumulation_steps`.

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -755,6 +787,7 @@ def log_training_summary(
     timer_agg: TimerAggregateMetrics,
     fabric: Fabric,
     global_batch_size: int,
+    gradient_accumulation_steps: int = 1,
```
guarin: Do not use default value otherwise it is easy to forget to pass that value from `train_task.py`

liopeer: Totally correct. Removed it.

src/lightly_train/_commands/train_task.py
```diff
@@ -1349,18 +1412,25 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
...
-            )
+                timer.start_step("train_dataload")
+                batch = next(infinite_train_dataloader)
+                timer.end_step("train_dataload")
+
+                # Type ignore is needed because `train_model` is not recognized as an
+                # instance of `_FabricModule`
+                with fabric.no_backward_sync(train_model, enabled=is_accumulating):  # type: ignore[arg-type]
+                    train_result = train_model.training_step(
+                        fabric=fabric, batch=batch, step=step
```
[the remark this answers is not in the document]
liopeer: Hmmm. This is indeed an issue (the curves will likely look noisier). Will address.

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -658,6 +658,38 @@ def get_steps(steps: int | Literal["auto"], default_steps: int) -> int:
...
+        default_batch_size: The model's default batch size.
+
+    Returns:
+        The number of gradient accumulation steps.
+
+    Raises:
+        ValueError: If gradient_accumulation_steps < 1.
+    """
+    if gradient_accumulation_steps == "auto":
+        return max(1, default_batch_size // global_batch_size)
```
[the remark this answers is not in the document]
liopeer: This is a deliberate choice. We use floor operation and only scale the learning rate downwards. At least for fine-tuning we never have super large batch sizes, so the current implementation is good enough imo. For pre-training we should maybe use the above suggested option.

src/lightly_train/_commands/train_task.py
liopeer: Yes, I think that makes indeed more sense. Then we have everything calculated on the "full" batch, basically ignoring gradient accumulation:
1. Average time needed to do a single optimization step (i.e. one or many accumulation calls).
2. Average waiting time to accumulate a single "full" batch (i.e. one or multiple accumulation batches).
3. Average throughput based on the "full" batch.
Will change accordingly.

liopeer: Done.

src/lightly_train/_commands/train_task.py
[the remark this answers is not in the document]
liopeer: Addressed.

src/lightly_train/_commands/train_task.py
```diff
@@ -1429,6 +1506,8 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
                 timer.reset_gpu_max_memory("val")
 
                 val_dataloader_iter = iter(val_dataloader)
+                # TODO (Lionel, 02/26): Average metrics during validation instead of
```
liopeer, line 1509: I deliberately do not accumulate/average metrics for now, since I figured this will be part of your metrics refactoring anyway @guarin.

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -697,6 +729,7 @@ def log_step(
     task: str,
     timer_agg: TimerAggregateMetrics,
     global_batch_size: int,
+    gradient_accumulation_steps: int = 1,
 ) -> None:
```
guarin, line 733:
```suggestion
    gradient_accumulation_steps: int,
) -> None:
```

PR #640, opened 2026-03-09
Add torch.compile support for image classification

guarin:
## What has changed and why?

* Compile image classification models by default
* Add `torch_compile_args` for all tasks

Compilation is by default disabled for all models except image classification. For image classification I got ~25% training speed increase on a single RTX3090.

I also tested it for object detection but this requires more work due to model architectures and variable batch shapes/contents.

## How has it been tested?

* Manually
* CI

- [x] Yes

- [x] Yes

### DIFF
--- a/.github/workflows/check_code_format.yml
+++ b/.github/workflows/check_code_format.yml
@@ -43,15 +43,6 @@ jobs:
       matrix:
         python: ["3.8", "3.13"]
     steps:
-      - name: Setup cmake
-        # TODO(04/25, Guarin): Check if this is still needed on GitHub hosted runners.
-        # TODO(04/04, Yutong): cmake executable is installed as a requirement for RFDETR support
-        # this setup is included only as a last resort because
-        # including cmake in the dependencies does not add cmake exec to PATH
-        # We should remove it once RFDETR improves their dependency spec since it causes a lot of overhead on the CI.
-        uses: jwlawson/actions-setup-cmake@v2
-        with:
-          cmake-version: '3.31'
       - name: Checkout Code
         uses: actions/checkout@v4
       - name: Set Up uv
--- a/.github/workflows/test_unit.yml
+++ b/.github/workflows/test_unit.yml
@@ -44,15 +44,6 @@ jobs:
         runners: ["ubuntu-latest", "windows-latest"]
     runs-on: ${{ matrix.runners }}
     steps:
-      - name: Setup cmake
-        # TODO(04/25, Guarin): Check if this is still needed on GitHub hosted runners.
-        # TODO(04/04, Yutong): cmake executable is installed as a requirement for RFDETR support
-        # this setup is included only as a last resort because
-        # including cmake in the dependencies does not add cmake exec to PATH
-        # We should remove it once RFDETR improves their dependency spec since it causes a lot of overhead on the CI.
-        uses: jwlawson/actions-setup-cmake@v2
-        with:
-          cmake-version: '3.31'
       - name: Checkout Code
         uses: actions/checkout@v4
       - name: Set Up uv


src/lightly_train/_torch_compile.py
```diff
@@ -0,0 +1,90 @@
...
+
+def try_compile(fn: _T, name: str, torch_compile_args: TorchCompileArgs) -> _T:
+    """Tries to compile a torch model or function and falls back to the uncompiled
+    version if any error occurs."""
+    if torch_compile_args.disable:
+        return fn
+    if not hasattr(torch, "compile"):
+        return fn
+    if sys.platform.startswith("win"):
+        # Compilation is currently unstable on Windows.
```
yutong-xiang-97, line 77: Maybe add a note in the readme/docs/changelog that compile is currently disabled on Windows

src/lightly_train/_torch_compile.py
```diff
@@ -0,0 +1,90 @@
...
+        return fn
+
+    logger.info(f"Compiling {name} with torch.compile")
+    start_time = time.perf_counter()
+    try:
```
yutong-xiang-97, line 82: From Codex:

The fallback here only protects the `torch.compile(...)` call itself, but PyTorch compilation is lazy: backend/operator errors usually surface on the first invocation of the returned callable, not when the wrapper is created. With image classification now opting into compilation by default, unsupported environments or ops will still crash on the first train/val batch instead of reverting to eager mode.

The practical fix is to move the fallback to call time, not compile-registration time. `torch.compile(...)` often just returns a wrapped callable; the real compilation and many backend/operator failures happen on the first execution. So `try_compile()` should:

- Keep the original eager function.
- Build the compiled wrapper once.
- Return a small wrapper that calls the compiled version.
- If that call raises a compiler-specific exception, log once, permanently switch to eager, and rerun eagerly.
- Do not catch arbitrary runtime errors, or you’ll hide real model bugs.

src/lightly_train/_torch_compile.py
```diff
@@ -0,0 +1,90 @@
...
+        fn = torch.compile(fn, **torch_compile_args.model_dump())  # type: ignore
+    except Exception as ex:
+        logger.warning(
+            f"Compilation failed, falling back to uncompiled version. Error: {ex}"
+        )
+        return fn
+    total_time = time.perf_counter() - start_time
```
yutong-xiang-97, line 89: From Codex

Lazy compilation also affects compile timing. The current "Compilation completed in X seconds" log around `torch.compile(...)` is misleading, because that mostly measures wrapper creation. PyTorch’s docs say compiled regions are compiled when executed and cached for future use, and their tutorial notes the first few iterations are slower because of compile overhead. Dynamic-shape guard failures can also trigger later recompilations, so there usually isn’t a single one-time “compile time”.

So I’d change the timing behavior too:

Remove or rename the current log.
If you want a useful metric, time the first successful compiled invocation instead.
Expect additional compile time later if shapes/guards change.

src/lightly_train/_torch_compile.py
```diff
@@ -0,0 +1,90 @@
...
+    if not hasattr(torch, "compile"):
+        return fn
+    if sys.platform.startswith("win"):
+        # Compilation is currently unstable on Windows.
+        return fn
+
+    logger.info(f"Compiling {name} with torch.compile")
+    start_time = time.perf_counter()
+    try:
+        fn = torch.compile(fn, **torch_compile_args.model_dump())  # type: ignore
```
yutong-xiang-97, line 83: From Claude Code:

`model_dump()` includes disable key — `torch_compile_args.model_dump()` will pass `disable=True/False` as a kwarg to `torch.compile(**...)`. This works because `torch.compile` happens to accept a disable parameter. If the field were named something else, it would break.

src/lightly_train/_task_models/image_classification/train_model.py
```diff
@@ -162,12 +168,15 @@ def __init__(
     def get_task_model(self) -> ImageClassification:
         return self.model
 
+    def forward(self, images: Tensor) -> Tensor:
```
yutong-xiang-97, line 171: From Codex:

The fragile part is that the current code in `train_task.py` (line 1413) mutates the forward attribute on one specific instance:

```py
train_model.forward = _torch_compile.try_compile(train_model.forward, ...)
```

That has a few consequences.

First, forward stops being the normal method defined on the class in `image_classification/train_model.py` (line 171). It becomes an instance-level callable that shadows the class method. `self(images)` still works today because `nn.Module.__call__` just looks up `self.forward` and calls it. But this is now “monkey-patched instance state”, not a regular method anymore.

Second, you install that patched callable before `fabric.setup()` in `train_task.py` (line 1413). So the compiled callable is created from the pre-wrap object. If later framework code expects to wrap, inspect, replace, or reason about `forward` after setup, you are depending on that patched callable still matching the final runtime object correctly. That may still work, but it is harder to reason about and easier to break.

Third, it makes the contract fuzzy. The class says “`forward()` is implemented here”, but at runtime the real logic is swapped out elsewhere. That is exactly the kind of thing that tends to confuse tooling, wrappers, debugging, and future refactors.

So the problem is mostly:

- not incorrect binding,
- but brittle runtime mutation of a core method.

The cleaner solution is to keep `forward()` stable and make it delegate to a private implementation callable.

Example shape:
```py
class ImageClassificationTrain(TrainModel):
    def __init__(self, ...):
        ...
        self._forward_impl = self.model.forward_train

    def configure_torch_compile(self, torch_compile_args):
        self._forward_impl = _torch_compile.try_compile(
            self.model.forward_train,
            "model",
            torch_compile_args=torch_compile_args,
        )

    def forward(self, images):
        return self._forward_impl(images)
```

Then `train_task.py` (line 1413) calls `train_model.configure_torch_compile(...)` instead of overwriting forward.

Why this is better:

- `forward` remains a normal method.
- `self(images)` and `self.forward(images)` both stay conventional.
- compilation is explicit model configuration, not method replacement.
- easier to debug, test, and maintain.
- less likely to interact badly with Fabric or future wrappers.

So the short version is: the current implementation probably works, but it works by mutating a fundamental method on one object instance. The safer design is to keep the method fixed and swap only the internal callable it uses.

src/lightly_train/_torch_compile.py
guarin: Wouldn't add it to the docs yet as compilation is not a documented feature. It is more in preview mode. This is also the reason why it is disabled by default.

src/lightly_train/_torch_compile.py
guarin: I'm aware of this. The "proper" way to do this is to pass some images to the model at compile time but this is much more work and needs some refactoring. The potential for errors is why compilation is disabled by default for all tasks except classification where I know it works well.

src/lightly_train/_torch_compile.py
guarin: Aware of that, see comment above.

src/lightly_train/_torch_compile.py
guarin: This is expected.

src/lightly_train/_torch_compile.py
yutong-xiang-97: So will there be follow-ups for this?

src/lightly_train/_task_models/image_classification/train_model.py
guarin: Codex got this mostly wrong. The reason it is implemented like this is because `torch.compile(train_model)` would result in `forward`, `training_step`, and `validation_step` being compiled which we don't want because `training_step` and `validation_step` are hard to make compile-friendly as they contain metrics and lots of for loops. Finally, they also get `batch` as input which contains filenames which triggers a recompile in every step.  I updated the code to include comments that explain this.

src/lightly_train/_torch_compile.py
guarin: Right now no because it is too much effort for too little gain. I mostly tried compilation out to speed up object detection training but didn't get significant improvements. Might be worthwhile to revisit later.

PR #642, opened 2026-03-13
Unify DINOv3 fine-tuning backbone loading

guarin:
## What has changed and why?

* Unify DINOv3 fine-tuning backbone loading
* Fix bug when loading `dinov3/vitt16-ltdetr` which loaded `dinov3/vitt16-notpretrained-ltdetr` instead

## How has it been tested?

* Manually

- [x] Yes

- [ ] Yes

### DIFF
--- a/.github/workflows/check_code_format.yml
+++ b/.github/workflows/check_code_format.yml
@@ -43,15 +43,6 @@ jobs:
       matrix:
         python: ["3.8", "3.13"]
     steps:
-      - name: Setup cmake
-        # TODO(04/25, Guarin): Check if this is still needed on GitHub hosted runners.
-        # TODO(04/04, Yutong): cmake executable is installed as a requirement for RFDETR support
-        # this setup is included only as a last resort because
-        # including cmake in the dependencies does not add cmake exec to PATH
-        # We should remove it once RFDETR improves their dependency spec since it causes a lot of overhead on the CI.
-        uses: jwlawson/actions-setup-cmake@v2
-        with:
-          cmake-version: '3.31'
       - name: Checkout Code
         uses: actions/checkout@v4
       - name: Set Up uv
--- a/.github/workflows/test_unit.yml
+++ b/.github/workflows/test_unit.yml
@@ -44,15 +44,6 @@ jobs:
         runners: ["ubuntu-latest", "windows-latest"]
     runs-on: ${{ matrix.runners }}
     steps:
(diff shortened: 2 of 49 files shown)


src/lightly_train/_models/dinov3/dinov3_package.py
```diff
@@ -246,6 +246,8 @@ def get_model(
         if not load_weights:
             args["weights"] = None
             args["pretrained"] = False
+        if load_weights and args.get("weights") is not None:
```
yutong-xiang-97, line 249: From Codex (explained by CC):

```text
  1. Task model sets backbone_model_args["weights"] = str(backbone_weights) (a local path)
  2. get_model() at line 249-250 forces args["pretrained"] = True because load_weights=True and args.get("weights") is not None
  3. Builder (_make_dinov3_vit) receives pretrained=True, weights="/path/run_a/exported_best.pt"
  4. Since type(weights) is str (not Weights enum), hits the else branch at line 152-153:
  url = convert_path_or_url_to_url(weights)
  # -> "file:///path/run_a/exported_best.pt"
  5. torch.hub.load_state_dict_from_url(url, ...) at line 155 computes:
  filename = os.path.basename(urlparse(url).path)  # -> "exported_best.pt"
  cached_file = os.path.join(hub_dir, "checkpoints", filename)
  if not os.path.exists(cached_file):  # skips if already cached!
      download_url_to_file(url, cached_file, ...)

  So if two different files share the same basename (e.g., run_a/exported_best.pt and run_b/exported_best.pt), the second call silently loads the first's cached copy from
  ~/.cache/torch/hub/checkpoints/exported_best.pt. The cache persists across process runs.

  Nuance: this is a regression only for the segmentation task models

  - Old segmentation code used the now-removed load_backbone_weights() which called torch.load(path, ...) directly — no caching, no issue.
  - Old LTDETR code already passed backbone_weights as backbone_args["weights"] to the builder, so it already went through torch.hub.load_state_dict_from_url. The caching issue
  pre-existed for LTDETR.
  - New code routes all four task models through the builder path, introducing the caching issue for the three segmentation models where it didn't exist before.

  Since exported_best.pt and exported_last.pt are the standard exported backbone filenames, basename collisions are likely in practice.
```

guarin: Oh wow, crazy bug. Updated!

PR #647, opened 2026-03-18
Remove detector weight decay

guarin:
## What has changed and why?

* Remove `detector_weight_decay` argument from LTDETR
* Remove LTDETR default optimizer param group
* Raise error if params require grad but are not in any optimizer group

This argument is not really needed and was a small bug. The default in RTDETRv2 is to apply the same weight decay to the backbone and the detector with the exception of the norm and bias layers which do not use weight decay. This PR fixes an issue where we mistakenly applied no weight decay to all layers in the detector.

## How has it been tested?

* Manually

- [ ] Yes

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,9 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- Fix a bug when loading DINOv3 LTDETR checkpoints that were not pretrained on COCO
+  which resulted in backbone weights not being loaded.
+
 ### Security
 
 ## [0.14.2] - 2026-02-24
--- a/src/lightly_train/_models/dinov3/dinov3_package.py
+++ b/src/lightly_train/_models/dinov3/dinov3_package.py
@@ -246,6 +246,8 @@ class DINOv3Package(Package):
         if not load_weights:
             args["weights"] = None
             args["pretrained"] = False
+        if load_weights and args.get("weights") is not None:
+            args["pretrained"] = True
         model = model_builder(**args)
         assert isinstance(model, (DinoVisionTransformer, ConvNeXt))
         return model
--- a/src/lightly_train/_models/dinov3/dinov3_src/hub/backbones.py
+++ b/src/lightly_train/_models/dinov3/dinov3_src/hub/backbones.py
@@ -9,6 +9,7 @@
 # - Add small test models
 from __future__ import annotations
 
+import logging
 import os
 from enum import Enum
 from pathlib import Path
@@ -18,6 +19,8 @@ from urllib.parse import urlparse
 import torch
 
 from .utils import DINOV3_BASE_URL
+
+logger = logging.getLogger(__name__)
 
 
 class Weights(Enum):
@@ -148,9 +151,15 @@ def _make_dinov3_vit(
             )
         else:
             url = convert_path_or_url_to_url(weights)
-        state_dict = torch.hub.load_state_dict_from_url(
-            url, map_location="cpu", check_hash=check_hash
-        )
+
+        if isinstance(weights, (str, Path)) and Path(weights).is_file():
+            logger.info(f"Loading DINOv3 checkpoint from '{weights}'")
+            state_dict = torch.load(weights, map_location="cpu", weights_only=False)
+        else:
+            logger.info(f"Loading DINOv3 checkpoint from '{url}'")
+            state_dict = torch.hub.load_state_dict_from_url(
+                url, map_location="cpu", check_hash=check_hash
+            )
 
         # Re-sample the projection weights before loading the statedict.
         key = "patch_embed.proj.weight"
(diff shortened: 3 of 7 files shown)


src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
```diff
@@ -114,8 +114,6 @@ class DINOv2LTDETRObjectDetectionTrainArgs(TrainModelArgs):
     backbone_lr_factor: float = 1e-2
     backbone_weight_decay: float | None = None  # Use default if None
```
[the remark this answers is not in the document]
guarin: This is ok

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -1394,3 +1395,43 @@ def finetune_from_checkpoint(
...
+
+    # Log optimizer parameter groups
+    param_id_to_name = {id(param): name for name, param in model.named_parameters()}
+    logger.debug("Optimizer parameter groups")
+    for group in optimizer.param_groups:
+        group_info = group.copy()
+        group_info.pop("params", None)
+        logger.debug(f"  {group_info}")
+        for param in group["params"]:
+            param_name = param_id_to_name[id(param)]
```
yutong-xiang-97, line 1421: Is there a potential to get a `KeyError` here?

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
```diff
@@ -114,8 +114,6 @@ class DINOv2LTDETRObjectDetectionTrainArgs(TrainModelArgs):
...
 
-    detector_weight_decay: float = 0.0
```
yutong-xiang-97, line 118: This is a breaking change for anyone with an existing YAML config (from Codex):

Removing this public model arg turns previously valid LT-DETR training configs into hard failures. `TrainModelArgs` inherits the repo-wide      
strict config behavior `(extra="forbid")`, so any saved YAML/CLI config that still sets `model_args.detector_weight_decay` (often explicitly to 0.0, the old default) will now fail validation before training starts.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -113,8 +113,6 @@ class DINOv3LTDETRObjectDetectionTrainArgs(TrainModelArgs):
     backbone_lr_factor: float = 1e-2
     backbone_weight_decay: float | None = None  # Use default if None
 
-    detector_weight_decay: float = 0.0
```
yutong-xiang-97, line 117: Same here

yutong-xiang-97 (commented):
Left some comments.

A clarifying question: by "The default in RTDETRv2 is to apply the same weight decay to the backbone and the detector with the exception of the norm and bias layers which do not use weight decay. ", do you mean that both the backbone and detector use the identical weight decay (minus norm and bias layers), or all the backbone / detector parameters uses the same weight decay, but weight decay values can be differenr for backbone and detector?

src/lightly_train/_commands/train_task_helpers.py
guarin: Theoretically no. `model` contains all possible parameters that should be in the optimizer param groups. In practice, if there is some other parameter in there we definitely want it to raise a KeyError

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
guarin: Yes and this is fine as this parameter should really not be used.

guarin:
> A clarifying question: by "The default in RTDETRv2 is to apply the same weight decay to the backbone and the detector with the exception of the norm and bias layers which do not use weight decay. ", do you mean that both the backbone and detector use the identical weight decay (minus norm and bias layers), or all the backbone / detector parameters uses the same weight decay, but weight decay values can be differenr for backbone and detector?

They use the same weight decay values

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
yutong-xiang-97: Then should we give an heads-up in the Changelog or something? My understanding is that users have their own YAML files, and this would give an error once they update the package.

yutong-xiang-97:
> > A clarifying question: by "The default in RTDETRv2 is to apply the same weight decay to the backbone and the detector with the exception of the norm and bias layers which do not use weight decay. ", do you mean that both the backbone and detector use the identical weight decay (minus norm and bias layers), or all the backbone / detector parameters uses the same weight decay, but weight decay values can be differenr for backbone and detector?
> 
> They use the same weight decay values

OK but now backbone uses `backbone_weight_decay` if there is a `self.model_args.backbone_weight_decay` and fall back to `self.model_args.weight_decay` while detector uses `self.model_args.weight_decay`. Is this intended?

guarin:
> OK but now backbone uses backbone_weight_decay if there is a self.model_args.backbone_weight_decay and fall back to self.model_args.weight_decay while detector uses self.model_args.weight_decay. Is this intended?

Ah yes, great point. I'll also remove the `backbone_weight_decay`

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
guarin: Updated the changelog. I don't expect that anyone used these arguments (`detector_weight_decay` and `backbone_weight_decy`)

PR #648, opened 2026-03-18
Remove cmake install in github actions

guarin:
## What has changed and why?

* Remove unneeded cmake install

## How has it been tested?

* CI

- [ ] Yes

- [ ] Yes


guarin:
> Approved but also consider all mentions of cmake in Makefile and docs: https://github.com/search?q=repo%3Alightly-ai%2Flightly-train%20cmake&type=code

I think the other mentions still apply for the old RFDETR versions

PR #650, opened 2026-03-19
Add YOLOOrientedObjectDetectionDataset for oriented object detection

gabrielfruet:
## What has changed and why?

- Add `YOLOOrientedObjectDetectionDataset` for loading YOLO oriented object detection datasets
- Add `oriented_bbox_from_corners()` helper function to convert 4-corner format to (cx, cy, w, h, angle)

The new dataset class enables training models on rotated/object detection tasks where bounding boxes have angles (OBB format).

## How has it been tested?

- Unit tests (`pytest tests/_data/test_yolo_helpers.py tests/_data/test_yolo_oriented_object_detection_dataset.py`)
- All 11 tests pass

- [x] Yes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -233,17 +233,20 @@ endif
 # 
 # This problem is fixed in pywinpty>2.0.14 but these versions are not compatible with
 # Python 3.8.
-EXTRAS_PY38 := [dev,dicom,mlflow,onnx,super-gradients,tensorboard,timm,ultralytics,wandb]
+#
+# SuperGradients is excluded as it is outdated and causes issues in CI.
+EXTRAS_PY38 := [dev,dicom,mlflow,onnx,tensorboard,timm,ultralytics,wandb]
 
-# SuperGradients is not compatible with Python>=3.10. It is also not easy to install
-# on MacOS. Therefore we exclude it from the default extras.
 # RFDETR has installation issues because of onnxsim dependency on CI with Python 3.13.
 # Onnx dependencies in RFDETR should become optional in RFDETR >1.1.0.
+#
+# SuperGradients is excluded as it is not compatible with Python>=3.10.
 EXTRAS_PY313 := [dev,dicom,mlflow,notebook,onnx,onnxruntime,onnxslim,tensorboard,timm,ultralytics,wandb]
 
 # RF-DETR is not always installable for Python>=3.12, therefore we remove it from the
-# default development dependencies. And SuperGradients is not compatible with
-# Python>=3.10, therefore we also remove it from the default development dependencies.
+# default development dependencies.
+
+# SuperGradients is excluded as it is not compatible with Python>=3.10.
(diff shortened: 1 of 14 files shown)


src/lightly_train/_data/yolo_helpers.py
```diff
@@ -131,3 +137,55 @@ def binary_masks_from_polygons(
         return np.stack(binary_masks)
     else:
         return np.zeros((0, height, width), dtype=np.bool_)
+
+
+def oriented_bbox_from_corners(corners: NDArray4Corners) -> NDArrayOBBoxes:
+    """Convert 4-corner format to (cx, cy, w, h, angle) format.
+
+    Uses geometric approach: the center is the midpoint of opposite
```
gabrielfruet: We assume here that the user is providing clockwise/counterclockwise corners. We don't check, but if necessary, would be a defensive addition.

src/lightly_train/_data/yolo_oriented_object_detection_dataset.py
```diff
@@ -0,0 +1,245 @@
...
+        else:
+            bboxes_np = np.zeros((0, 5), dtype=np.float64)
+
+        # Convert numpy arrays to tv_tensors for torchvision transforms.
+        # Image: (H, W, C) -> (C, H, W) for tv_tensors.Image
+        image_tensor = torch.from_numpy(image_np).permute(2, 0, 1)
+        tv_image = tv_tensors.Image(image_tensor)
+
+        # Bboxes: normalized (cx, cy, w, h, angle) with canvas size (h, w)
+        if bboxes_np.shape[0] > 0:
```
gabrielfruet: This could be simply reduced to one `tv_tensors.BoundingBoxes` call with a ternary operator on the bboxes argument.

guarin: If I understand correctly you don't need the if/else at all because `bboxes_np` is always a numpy array with dim `(N, 5)`. So `torch.from_numpy(bboxes_np)` will return the same as `torch.zeros((0, 5), dtype=torch.float64)` in the empty bbox case.

guarin (approved):
LGTM! Thanks for the PR! Left one suggestion but this already looks good for merging as is.

src/lightly_train/_data/yolo_oriented_object_detection_dataset.py
gabrielfruet: That's true. I updated the `ndmin` when reading the array and forgot to update it here. I'll simplify a bit things. Thank you for this insight!

gabrielfruet:
@guarin 

* Removed the unnecessary if statement
* Made the label loading more linear (branching can hurt a bit the code readability) 
* I took an opportunity to simplify the data loading (OBB and non-OBB). It will be significantly faster due to using `numpy` IO operations.
    * If you wish i could do a benchmark on that.
    * We cannot do the same to instance segmentation labels because they can have a variable amount of elements per line.

gabrielfruet:
Tests are failing due to the torchvision version. I'll handle that with RequirementsCache

gabrielfruet:
removed the file loading logic update, will add in a follow-up pr.

PR #655, opened 2026-03-20
Log ci container

guarin:
## What has changed and why?

* Logs whether LightlyTrain is running in CI or a container

## How has it been tested?

* Unit tests

- [ ] Yes

- [x] Yes


tests/_events/test_tracker.py
```diff
@@ -348,3 +350,105 @@ def test_track_inference_started__without_batch_size(
     assert props["task_type"] == "embedding"
     assert props["model_name"] == "EmbeddingModel"
     assert "batch_size" not in props
+
+
+def test__is_ci__true(mocker: MockerFixture) -> None:
```
yutong-xiang-97, line 355: From CC:

No test for `_is_ci()` with empty string value: `CI="" `should still return `True` per the implementation. Worth a quick parametrized test to document this behavior explicitly.

tests/_events/test_tracker.py
```diff
@@ -348,3 +350,105 @@ def test_track_inference_started__without_batch_size(
...
+
+    assert tracker._is_ci() is True
+
+
+def test__is_ci__false(mocker: MockerFixture) -> None:
+    """Test that _is_ci returns False when CI environment variable is not set."""
+    mocker.patch.dict(os.environ, {}, clear=True)
+    # Ensure CI is not set by deleting it if it exists.
+    if "CI" in os.environ:
+        del os.environ["CI"]
```
yutong-xiang-97: From CC:

`test__is_ci__false` redundant cleanup: After `mocker.patch.dict(os.environ, {}, clear=True)`, manually checking and deleting CI is redundant — `clear=True `already removed it. The safety net is harmless but adds dead code.

tests/_events/test_tracker.py
```diff
@@ -348,3 +350,105 @@ def test_track_inference_started__without_batch_size(
...
+    mocker.patch.dict(os.environ, {"CI": "true"})
+
+    assert tracker._is_ci() is True
+
+
+def test__is_ci__false(mocker: MockerFixture) -> None:
+    """Test that _is_ci returns False when CI environment variable is not set."""
+    mocker.patch.dict(os.environ, {}, clear=True)
+    # Ensure CI is not set by deleting it if it exists.
+    if "CI" in os.environ:
```
yutong-xiang-97: From CC:

No test for the exception path in `_is_container()`: The `False` case (`test__is_container__false`) mocks open to return non-matching content. There are no tests for:
- `FileNotFoundError` when `/proc/1/cgroup` doesn't exist (the common macOS/Windows case)
- `PermissionError` when the file is unreadable.

tests/_events/test_tracker.py
```diff
@@ -348,3 +350,105 @@ def test_track_inference_started__without_batch_size(
...
+        side_effect=lambda path: path == "/run/.containerenv",
+    )
+
+    assert tracker._is_container() is True
+
+
+def test__is_container__singularity(mocker: MockerFixture) -> None:
+    """Test that _is_container returns True when SINGULARITY_CONTAINER is set."""
+    mocker.patch("os.path.isfile", return_value=False)
+    mocker.patch.dict(os.environ, {"SINGULARITY_CONTAINER": "1"})
```
yutong-xiang-97: From CC:

Singularity/Apptainer tests don't clear env: Unlike the cgroup tests that use `clear=True`, these tests only mock `isfile` and add the specific env var. If the test runner itself sets e.g. `SINGULARITY_CONTAINER`, both singularity and apptainer tests would pass for the wrong reason. Consider adding `mocker.patch.dict(os.environ, {"SINGULARITY_CONTAINER": "1"}, clear=True)` for isolation.

src/lightly_train/_events/tracker.py
```diff
@@ -141,6 +143,51 @@ def _get_device_count(devices: int | str | list[int]) -> int:
...
+
+    # Check for Singularity/Apptainer.
+    if os.environ.get("SINGULARITY_CONTAINER") is not None:
+        return True
+    if os.environ.get("APPTAINER_CONTAINER") is not None:
+        return True
+
+    # Check for Docker-in-K8s or containerd via cgroup.
+    try:
+        with open("/proc/1/cgroup", encoding="utf-8") as f:
```
yutong-xiang-97: From Codex and CC:

When LightlyTrain runs inside a container that shares the host PID namespace, `/proc/1/cgroup` describes the host init process rather than the training process, so this branch reports `is_container=False` even though the current run is containerized. That makes the newly added telemetry inaccurate for those deployments; reading the current process cgroup (`/proc/self/cgroup`) avoids this false negative.

The false negative would only occur when all of these are true simultaneously:
1. Shared host PID namespace, AND
2. No `/.dockerenv` or `/run/.containerenv` (e.g., containerd-only Kubernetes), AND
3. No Singularity/Apptainer env vars

This happens mainly with Kubernetes pods using hostPID: true on a containerd runtime — uncommon for ML training, but not impossible.

The fix (`/proc/self/cgroup`) is strictly better and costs nothing, so it's worth adopting. It's also what most container-detection libraries use (e.g., Python's `is-docker`, Go's `containerized`).

PR #658, opened 2026-03-23
Fix ScaleJitter Config

yutong-xiang-97:
## What has changed and why?

Fix the wrong config of ScaleJitter sizes of LT-DETR: `(640,640)` should be repeated 20 times by default.

## How has it been tested?

N/A.

- [x] Yes

- [ ] Yes


guarin:
Could you also update it for DINOv2 LTDETR?

PR #666, opened 2026-03-25
Disable torch.compile by default

guarin:
## What has changed and why?

* Disable torch.compile for classification as it isn't compatible with older GPUs

## How has it been tested?

* Manually

- [x] Yes

- [ ] Yes


guarin:
> Any plan to add them back later?

Yes once we get them to work reliably (or handle older GPU versions). But not highest prio right now.

PR #670, opened 2026-03-26
update EMA

yutong-xiang-97:
## What has changed and why?

Update EMA calculation with `_foreach_mul_`

## How has it been tested?

Unit tests

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_methods/dinov2/dinov2_loss.py
+++ b/src/lightly_train/_methods/dinov2/dinov2_loss.py
@@ -153,8 +153,9 @@ class DINOLoss(nn.Module):
                 self.reduce_handle.wait()
             _t = self.async_batch_center / (self.len_teacher_output * world_size)
 
-            self.center.mul_(self.center_momentum)
-            self.center.add_(_t, alpha=1 - self.center_momentum)
+            self.center = self.center * self.center_momentum + _t * (
+                1 - self.center_momentum
+            )
 
             self.updated = True
 
@@ -289,7 +290,8 @@ class IBOTPatchLoss(nn.Module):
                 self.reduce_handle.wait()
             _t = self.async_batch_center / (self.len_teacher_patch_tokens * world_size)
 
-            self.center.mul_(self.center_momentum)
-            self.center.add_(_t, alpha=1 - self.center_momentum)
+            self.center = self.center * self.center_momentum + _t * (
+                1 - self.center_momentum
+            )
 
             self.updated = True
--- a/src/lightly_train/_torch_helpers.py
+++ b/src/lightly_train/_torch_helpers.py
@@ -9,7 +9,6 @@ from __future__ import annotations
 
 import contextlib
 import os
-from collections import defaultdict
 from typing import Any, Callable, Generator
 
 import torch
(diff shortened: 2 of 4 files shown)


src/lightly_train/_methods/dinov2/dinov2_loss.py
```diff
@@ -153,9 +153,8 @@ def apply_center_update(self) -> None:
                 self.reduce_handle.wait()
             _t = self.async_batch_center / (self.len_teacher_output * world_size)
 
-            self.center = self.center * self.center_momentum + _t * (
-                1 - self.center_momentum
-            )
+            self.center.mul_(self.center_momentum)
+            self.center.add_(_t, alpha=1 - self.center_momentum)
```
guarin: Doesn't need a change but this probably doesn't give a speedup as it is just a single operation and makes the code slightly harder to understand.

src/lightly_train/_torch_helpers.py
```diff
@@ -68,3 +70,36 @@ def set_warn_on_accumulate_grad_stream_mismatch(value: bool) -> None:
...
+    """Updates tensors with an exponential moving average using foreach ops."""
+    if not ema_tensors:
+        return
+
+    grouped_ema_tensors = defaultdict(list)
+    grouped_tensors = defaultdict(list)
+    for ema_tensor, tensor in zip(ema_tensors, tensors):
+        key = (ema_tensor.device.type, ema_tensor.device.index, ema_tensor.dtype)
+        grouped_ema_tensors[key].append(ema_tensor)
+        grouped_tensors[key].append(tensor)
```
guarin: Why is the grouping needed?

src/lightly_train/_torch_helpers.py
```diff
@@ -68,3 +70,36 @@ def set_warn_on_accumulate_grad_stream_mismatch(value: bool) -> None:
     # suppress this warning.
     if hasattr(torch.autograd.graph, "set_warn_on_accumulate_grad_stream_mismatch"):
         torch.autograd.graph.set_warn_on_accumulate_grad_stream_mismatch(value)  # type: ignore
+
+
+@torch.no_grad()
+def update_ema_tensors(
+    ema_tensors: list[Tensor],
+    tensors: list[Tensor],
+    decay: float,
```
guarin: I think it would be nice to have the same argument order as in `update_momentum`. Also same naming with `tensors`, `tensors_ema`, `m`.

src/lightly_train/_torch_helpers.py
```diff
@@ -68,3 +70,36 @@ def set_warn_on_accumulate_grad_stream_mismatch(value: bool) -> None:
...
+    grouped_tensors = defaultdict(list)
+    for ema_tensor, tensor in zip(ema_tensors, tensors):
+        key = (ema_tensor.device.type, ema_tensor.device.index, ema_tensor.dtype)
+        grouped_ema_tensors[key].append(ema_tensor)
+        grouped_tensors[key].append(tensor)
+
+    for key, grouped_ema in grouped_ema_tensors.items():
+        grouped_tensor = grouped_tensors[key]
+        torch._foreach_mul_(grouped_ema, decay)
+        torch._foreach_add_(grouped_ema, grouped_tensor, alpha=1.0 - decay)
```
guarin:
```suggestion
        torch._foreach_mul_(grouped_ema, m)
        torch._foreach_add_(grouped_ema, grouped_tensor, alpha=1.0 - m)
```

Following https://github.com/lightly-ai/lightly/pull/1899/changes

Note that momentum and decay are not the same thing. decay is 1 - momentum.

src/lightly_train/_torch_helpers.py
```diff
@@ -68,3 +70,36 @@ def set_warn_on_accumulate_grad_stream_mismatch(value: bool) -> None:
...
+
+
+@torch.no_grad()
+def update_momentum(model: Module, model_ema: Module, m: float) -> None:
+    """Updates parameters of `model_ema` with the EMA of `model`."""
+    update_ema_tensors(
+        ema_tensors=list(model_ema.parameters()),
+        tensors=list(model.parameters()),
+        decay=m,
```
guarin: Decay is not the same thing as momentum

src/lightly_train/_torch_helpers.py
yutong-xiang-97: I checked and it was due to a false statement of Codex that using tensors with different data types would break stuff. I removed it now.

PR #672, opened 2026-03-26
refactor: use np.loadtxt for YOLO label loading

gabrielfruet:
## Summary

Replace Python loop-based YOLO label parsing with `np.loadtxt` for improved performance on fixed-column formats, while preserving existing duplicate line handling.

### Changes

- `open_yolo_object_detection_label_numpy`: Now uses `np.loadtxt` + `np.unique` for parsing
- `open_yolo_oriented_object_detection_label_numpy`: Same pattern for 9-column oriented bboxes

### Benefits

- **Performance:** `np.loadtxt` is ~10-100x faster than Python loops for large files
- **Simplicity:** Fewer lines of code, clearer intent  
- **Order preservation:** Unlike `set()`, `np.unique(return_index=True)` preserves insertion order

### Tests

- 130 tests passed, 2 skipped
- Format checks pass

### DIFF
--- a/src/lightly_train/_data/file_helpers.py
+++ b/src/lightly_train/_data/file_helpers.py
@@ -392,60 +392,50 @@ def _open_mask_numpy__with_pil(
 def open_yolo_oriented_object_detection_label_numpy(
     label_path: Path,
 ) -> tuple[NDArray4Corners, NDArrayClasses]:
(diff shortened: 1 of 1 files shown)


gabrielfruet:
Using correct template...

PR #673, opened 2026-03-26
Use np.loadtxt for YOLO label loading

gabrielfruet:
## What has changed and why?

Replace Python loop-based YOLO label parsing with `np.loadtxt` for improved performance on fixed-column formats, while preserving existing duplicate line handling. Improved also the documentation to include shapes for the arrays.

- `open_yolo_object_detection_label_numpy`: Now uses `np.loadtxt` + `np.unique` for parsing
- `open_yolo_oriented_object_detection_label_numpy`: Same pattern for 9-column oriented bboxes

## How has it been tested?

- 130 tests passed, 2 skipped
- Format checks pass (ruff, mdformat, pre-commit)
- Order preservation verified (duplicates are removed while keeping first occurrence)

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_data/file_helpers.py
+++ b/src/lightly_train/_data/file_helpers.py
@@ -392,60 +392,50 @@ def _open_mask_numpy__with_pil(
 def open_yolo_oriented_object_detection_label_numpy(
     label_path: Path,
 ) -> tuple[NDArray4Corners, NDArrayClasses]:
-    """Open a YOLO label file and return the oriented bounding boxes and classes as numpy arrays.
+    """Open a YOLO label file and return oriented bounding boxes and classes.
 
     Returns:
-        (bboxes, classes) tuple. All values are in normalized coordinates
-        between [0, 1]. Bboxes are formatted as (x1, y1, x2, y2, x3, y3, x4, y4).
+        (bboxes, classes):
+        - bboxes: shape (n, 8), normalized (x1, y1, x2, y2, x3, y3, x4, y4)
+        - classes: shape (n,)
     """
-    oriented_bboxes = []
-    classes = []
-    for line in _iter_yolo_label_lines(label_path=label_path):
-        parts = [float(x) for x in line.split()]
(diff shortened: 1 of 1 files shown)


src/lightly_train/_data/file_helpers.py
```diff
@@ -395,33 +395,23 @@ def open_yolo_oriented_object_detection_label_numpy(
...
-        else np.zeros((0, 8), dtype=np.float64)
-    )
-    classes_np = np.array(classes, dtype=np.int64)
-    return oriented_bboxes_np, classes_np
+    if data.size == 0:
+        return np.zeros((0, 8), dtype=np.float64), np.zeros((0,), dtype=np.int64)
+
+    # Remove duplicate lines (preserve order by keeping first occurrence)
+    _, unique_idx = np.unique(data, axis=0, return_index=True)
+    unique_idx = np.sort(unique_idx)  # Sort to preserve original order
```
gabrielfruet, line 409: This `np.sort` line is here only because `np.unique` returns a sorted array, and our current testing suite doesn't handle the case where the input lines order are different than the output lines order. 

https://github.com/lightly-ai/lightly-train/blob/5eb67a1e7e97808e2807e02dad751e4543320fd7/tests/_data/test_file_helpers.py#L463-L484

An alternative would be to modify this behavior to accept a shuffled output. I think this would not be a problem, only if some component rely on the order of the lines in the label file.

simonschoelly:
Hi Gabriel thanks for this! Unfortunately we just merged a [PR](https://github.com/lightly-ai/lightly-train/pull/674) that changes the same code that you are touching here.

There might be a few other changes coming up in the next days as we adding the COCO format for object detection datasets, so I would wait and then see if this is still useful.

gabrielfruet:
No problem @simonschoelly. Please let me know if i should continue working on this PR.

simonschoelly:
Unfortunately we did some major changes while this PR was in process, so I am closing this.

PR #674, opened 2026-03-27
Parse labels in YoloObjectDetectionDatasetArgs.list_image_info

simonschoelly:
## What has changed and why?

This PR makes it so that `YOLOObjectDetectionDatasetArgs.list_image_info` parses the label files and returns the parsed results instead of returning the paths to the label files.

## How has it been tested?

I relayed on the existing tests which had to be slightly changed.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/tests/_data/test_yolo_object_detection_dataset.py
+++ b/tests/_data/test_yolo_object_detection_dataset.py
@@ -76,31 +76,13 @@ class TestYoloObjectDetectionDataset:
         train_dataset = YOLOObjectDetectionDataset(
             dataset_args=train_args,
             transform=ObjectDetectionTransform(DummyTransformArgs()),
-            image_info=[
-                {
-                    "image_path": str(tmp_path / "train/images/0.png"),
-                    "label_path": str(tmp_path / "train/labels/0.txt"),
-                },
-                {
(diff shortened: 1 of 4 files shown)


tests/_data/test_yolo_object_detection_dataset.py
```diff
@@ -164,32 +128,3 @@ def test__split_last(self, tmp_path: Path) -> None:
...
-
-    def test__get_item__internal_class_ids(self, tmp_path: Path) -> None:
-        create_yolo_object_detection_dataset(tmp_path=tmp_path, split_first=True)
-
-        args = YOLOObjectDetectionDataArgs(
-            path=tmp_path,
-            train="train/images",
-            val="val/images",
-            names={0: "class_0", 2: "class_2"},
-        )
```
guarin, line 176: Is there a new test that tests the class mapping in `list_image_info`?

simonschoelly: There is not currently - I removed the test because it is not a member of the class anymore. Perhaps I can check how I test this when I add the COCO datasets - there should be some shared code.

PR #675, opened 2026-03-29
Add step-aware LTDETR scale jitter control

yutong-xiang-97:
## What has been changed and why?

Enable step-based augmentation control for LTDETR object detection with the following methods:

"Keep all LTDETR augmentations inside the regular transform and collate path, and
propagate the current training step by updating the main-process transform/collate
objects before each batch. When a worker-visible schedule change crosses an augmentation boundary, restart the dataloader iterator so the worker processes receive fresh copies of the transform and collate objects with the new step value."

As a start, this PR applies this approach to the existing `ScaleJitter` augmentation. We use a `stop_step` to control where this augmentations should be stopped. It is achieved by:

1. In the respective `ScaleJitterArgs`, making LTDETR `scale_jitter` stop-aware via `stop_step`
2. adding shared step-control hooks to collate functions
  - `set_step`: sets the current optimizer step
  - `uses_step_dependent_worker_state` determine whether the step-based control is enabled
  - `requires_dataloader_reinitialization` flags if dataloader reinit is needed
3. Rewrite the above hooks in `ObjectDetectionCollateFunction` to determine if dataloader reset is needed.
4. adding `InfiniteCycleIterator.reset()` and restart dataloader iteration when step-dependent worker state changes
5. resetting the dataloader when `stop_step` is reached in the main loop.

Note that:
1. We force `persistent_workers=False` for step-aware train dataloaders even if the users specify `True` and emit both a warning and log message when overriding user input.
2. We don't reset `cycles` in the `InfiniteCycleIterator`.
3. The hooks are also added to `TaskTransform` as a placeholder for `MosaicTransform`.
4. Resize should be always on and applied before collate to match the behavior with original LTDETR.

## How has it been tested?

- Unit tests.
- Manually.

- [x] Yes

- [ ] Yes

### DIFF
--- a/tests/_commands/test_train.py
+++ b/tests/_commands/test_train.py
@@ -320,7 +320,7 @@ def test_train_from_dictconfig(tmp_path: Path) -> None:
         ),
         (
             "distillationv3",
-            lambda: DummyCustomModel(),
+            lambda: DummyCustomModel(stride=16),
         ),
     ],
 )
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -137,10 +137,12 @@ class DummyMethod(Method):
 
 
 class DummyCustomModel(Module, ModelWrapper):
-    def __init__(self, feature_dim: int = 2):
+    def __init__(self, feature_dim: int = 2, stride: int = 1) -> None:
         super().__init__()
         self._feature_dim = feature_dim
-        self.conv = Conv2d(in_channels=3, out_channels=feature_dim, kernel_size=2)
+        self.conv = Conv2d(
+            in_channels=3, out_channels=feature_dim, kernel_size=2, stride=stride
+        )
         self.global_pool = AdaptiveAvgPool2d(output_size=(1, 1))
 
     def feature_dim(self) -> int:


src/lightly_train/_commands/train_task.py
```diff
@@ -1488,6 +1489,11 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
 
         # TODO(Guarin, 07/25): Replace with infinite batch sampler instead to avoid
         # reloading dataloader after every epoch? Is this preferred over persistent workers?
+        train_transform = train_dataset.transform
+        # DataLoader erases the concrete TaskCollateFunction type of collate_fn.
```
guarin: `train_transform` is already created further above

src/lightly_train/_commands/train_task.py
```diff
@@ -1488,6 +1489,11 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
...
+        train_collate_fn = train_dataloader.collate_fn
+        assert isinstance(train_collate_fn, TaskCollateFunction)  # for mypy
```
guarin: Very minor:

```suggestion
        train_collate_fn: TaskCollateFunction = train_dataloader.collate_fn # type: ignore
```

src/lightly_train/_commands/train_task.py
```diff
@@ -1537,6 +1543,18 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
...
+            train_collate_fn.set_step(step)
+
+            needs_reinit = (
+                train_transform.requires_dataloader_reinitialization()
+                or train_collate_fn.requires_dataloader_reinitialization()
+            )
+            if config.num_workers > 0 and needs_reinit:
+                timer.start_step("train_dataloader_restart")
+                infinite_train_dataloader.reset()
+                timer.end_step("train_dataloader_restart")
```
guarin:
```suggestion
                infinite_train_dataloader.reset()
```

No need to crate a new timer step. If `reset()` only deletes the current dataloader then the dataloader will only be restarted in the `next(inifite_train_dataloader)` step further down which is the time-consuming part. This will be correctly attributed to the `train_dataload` step

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -651,10 +656,25 @@ def get_train_dataloader(
...
+            warning_message = (
+                "Step-aware transform/collate requires "
+                "persistent_workers=False. Overriding user-provided "
+                "persistent_workers=True."
+            )
+            logger.warning(warning_message)
+            warnings.warn(
+                warning_message,
+                stacklevel=2,
+            )
```
guarin: We usually only log with `logger.warning` and not with `warnings.warn` (which will also end up in the logger)

src/lightly_train/_data/infinite_cycle_iterator.py
```diff
@@ -21,6 +21,9 @@ def __init__(self, iterable: Iterable[_T]):
         self.cycles = 0
         self._iter: Iterator[_T] | None = None
 
+    def reset(self) -> None:
+        self._iter = iter(self.iterable)
```
guarin:
```suggestion
    def reset(self) -> None:
        self._iter = None
```

This will defer new worker creation to the next `__next__` call. See https://github.com/lightly-ai/lightly-train/pull/675/changes#r3008760898

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
```diff
@@ -135,8 +136,10 @@ class DINOv2LTDETRObjectDetectionTrainTransformArgs(ObjectDetectionTransformArgs
     image_size: ImageSizeTuple | Literal["auto"] = "auto"
     # TODO: Lionel (09/25): Remove None, once the stop policy is implemented.
     stop_policy: StopPolicyArgs | None = None
-    resize: ResizeArgs | None = None
-    scale_jitter: ScaleJitterArgs | None = Field(
```
guarin, line 139: Small note: Do you plan to remove `stop_policy` in a follow-up PR?

src/lightly_train/_transforms/object_detection_transform.py
```diff
@@ -293,13 +298,40 @@ def __init__(
...
+        return self._scale_jitter_stop_step is not None
+
+    def requires_dataloader_reinitialization(self) -> bool:
+        is_scale_jitter_applied_at_step = self._is_scale_jitter_applied_at_step(
+            self._step
+        )
+        if is_scale_jitter_applied_at_step == self._is_scale_jitter_applied_in_workers:
+            return False
+        self._is_scale_jitter_applied_in_workers = is_scale_jitter_applied_at_step
+        return True
```
guarin, line 331: I don't fully follow the logic here. Should this not be:

```suggestion
        return self._step == self._scale_jitter_stop_step
```

Because we only want to reset the dataloader a single time when the we change from scale jitter to no scale jitter. Once we disabled it we shouldn't reset again because resetting is very time consuming.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/transforms.py
```diff
@@ -69,6 +69,7 @@ class DINOv3LTDETRObjectDetectionRandomFlipArgs(RandomFlipArgs):
 
 
 class DINOv3LTDETRObjectDetectionScaleJitterArgs(ScaleJitterArgs):
+    stop_step: int | None = None
```
guarin: Does `stop_step=None` mean that it is never stopped? If yes you might want to add this as a comment for easier understanding.

src/lightly_train/_transforms/task_transform.py
```diff
@@ -48,6 +48,15 @@ def __init__(
...
         self.transform_args = transform_args
 
+    def set_step(self, step: int) -> None:
+        pass
+
+    def uses_step_dependent_worker_state(self) -> bool:
+        return False
+
+    def requires_dataloader_reinitialization(self) -> bool:
+        return False
```
guarin, line 61: Would add short docstrings explaining why those three methods are needed and what they achieve.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
```diff
@@ -68,6 +68,7 @@ class DINOv2LTDETRObjectDetectionRandomFlipArgs(RandomFlipArgs):
 
 
 class DINOv2LTDETRObjectDetectionScaleJitterArgs(ScaleJitterArgs):
+    stop_step: int | None = None
```
guarin: Would consider naming this `step_stop` instead. We usually follow the naming pattern `<name>_start`, `<name>_stop` etc. Where `<name>` is first and the property later. This has the benefit that the values are next to each other in configs. For example:

https://github.com/lightly-ai/lightly-train/blob/764ae1f4a40a2443cbd2dae18aa0b2dad6f177d8/src/lightly_train/_methods/dinov2/dinov2.py#L106-L107

or

https://github.com/lightly-ai/lightly-train/blob/764ae1f4a40a2443cbd2dae18aa0b2dad6f177d8/src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py#L74-L77

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
yutong-xiang-97: Will consider removing them once all the augmentations are done.

src/lightly_train/_transforms/object_detection_transform.py
yutong-xiang-97: This is more of a concern when we add the `MixUp` and `Copyblend` where there will be a `start_step` and `stop_step` so two resets. Will look into how to optimize this in the following PRs with those augmentations added.

src/lightly_train/_commands/train_task.py
```diff
@@ -1488,6 +1489,9 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
 
         # TODO(Guarin, 07/25): Replace with infinite batch sampler instead to avoid
         # reloading dataloader after every epoch? Is this preferred over persistent workers?
+        # DataLoader erases the concrete TaskCollateFunction type of collate_fn.
```
guarin, line 1490: The comment above should stay right above the `inifite_train_dataloader` definition:
```
# TODO(Guarin, 07/25): Replace with infinite batch sampler instead to avoid
# reloading dataloader after every epoch? Is this preferred over persistent workers?
infinite_train_dataloader = InfiniteCycleIterator(iterable=train_dataloader)
```

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/transforms.py
```diff
@@ -69,6 +69,8 @@ class DINOv3LTDETRObjectDetectionRandomFlipArgs(RandomFlipArgs):
 
 
 class DINOv3LTDETRObjectDetectionScaleJitterArgs(ScaleJitterArgs):
+    # None means scale jitter is always on.
+    step_stop: int | None = None
```
guarin: Could you add `step_stop` also to the parent `ScaleJitterArgs`? Then you don't need to use `gettatr` further down:
```
self._scale_jitter_step_stop = getattr(
    transform_args.scale_jitter, "step_stop", None
)
``` 

I think it is ok if `step_stop` is unused for the segmentation transforms. Alternatively you can create an `ObjectDetectionScaleJitterArgs` class that has `step_stop` defined while the others don't.

src/lightly_train/_transforms/object_detection_transform.py
guarin: Ooh I actually didn't understand the logic. Suggestion: Could you drop the `_is_scale_jitter_applied_in_workers` attribute in a follow-up PR and make the `requires_dataloader_reinitialization` method not set any state? I think it would make the logic easier to follow.

PR #677, opened 2026-03-30
Add coco object detection dataset

simonschoelly:
## What has changed and why?

- Added COCO format input for object detection
- Added some tests for `get_class_id_to_internal_class_id_mapping` because that method was not covered anymore after this [PR](https://github.com/lightly-ai/lightly-train/pull/674).
- Had to disable the `test_file_locking_concurrent_increments` test on Python 3.8 because it kept failing even after restarting the tests.

## How has it been tested?

- Wrote some tests to cover the new dataset
- Wrote a test that runs the train loop with a COCO dataset.

- [x] Yes

Coming in a follow up PR

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -472,7 +472,7 @@ def create_normalized_yolo_object_detection_labels(
     missing_label_indices: list[int] | None = None,
     empty_label_indices: list[int] | None = None,
 ) -> None:
-    # TODO (Simon, 03/26): Ideally we would also create test instances with less or more more than one bounding box.
+    # TODO (Simon, 03/26): Ideally we would also create test instances with less or more than one bounding box.
 
     if missing_label_indices is None:
         missing_label_indices = []
--- a/tests/_data/test_coco_object_detection_dataset.py
+++ b/tests/_data/test_coco_object_detection_dataset.py
@@ -42,10 +42,7 @@ class TestCOCOObjectDetectionDatasetArgs:
         )
 
         # class 3 -> internal 0, class 7 -> internal 1
-        for split, dataset_args in [
-            ("train", args.get_train_args()),
-            ("val", args.get_val_args()),
-        ]:
+        for dataset_args in [args.get_train_args(), args.get_val_args()]:
             image_info = list(dataset_args.list_image_info())
             assert len(image_info) == 3
             assert json.loads(image_info[0]["bboxes"]) == []


src/lightly_train/_data/object_detection_dataset.py
```diff
@@ -0,0 +1,75 @@
...
+        if not image_path.exists():
+            raise FileNotFoundError(f"Image file {image_path} does not exist.")
+
+        image_np = file_helpers.open_image_numpy(image_path)
+        h, w, _ = image_np.shape
+
+        bboxes = json.loads(image_info["bboxes"])
+        class_labels = json.loads(image_info["class_labels"])
+        # TODO (simon, 03/26) do we need this assert?
+        assert len(bboxes) == len(class_labels)
```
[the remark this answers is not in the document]
simonschoelly: This was just copied over from previous code. And it's a case that should never occur.

src/lightly_train/_data/coco_object_detection_dataset.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def train_imgs_path(self) -> Path:
+        # TODO (simon 03/26): We currently only need this to calculate a hash for the mmmap file for the dataset.
+        #  This might not be the best idea as the contents of the file might change.
+        return self.train_labels.resolve()
+
+    def val_imgs_path(self) -> Path:
+        # TODO (simon 03/26): We currently only need this to calculate a hash for the mmmap file for the dataset.
+        #  This might not be the best idea as the contents of the file might change.
+        return self.val_labels.resolve()
```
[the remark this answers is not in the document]
simonschoelly: Fixed the typo, but as discussed in our sync the mmap logic will be handled another time.

src/lightly_train/_commands/train_task.py
```diff
@@ -704,6 +708,10 @@ def train_object_detection(
         devices=devices,
         steps=steps,
     )
+    # This is just here as the format field was recently introduced so that we don't break existing code.
+    # It is easier to set this here than to use a Pydantic validator as that one creates issues with YAML parsing.
+    if isinstance(data, dict) and "format" not in data:
```
liopeer: This would indeed be nicer to handle inside Pydantic ...

tests/_commands/test_train_task.py
```diff
@@ -224,6 +224,53 @@ def test_train_object_detection(tmp_path: Path) -> None:
     assert results["labels"].ndim == 1
 
 
+def test_train_object_detection_coco(tmp_path: Path) -> None:
```
liopeer: I actually don't think we should have this integration test. I think our tests are already taking long enough to run and the integration test is already covered well enough by the test with the YOLO dataset. I'd rather focus on unit tests whereever possible.

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- Add support for COCO object detection dataset format.
```
liopeer, line 15: Will you document the newly supported format in a 2nd PR? Also because the recommended way is to now add `"format"` to the data arguments (otherwise it will fall back to YOLO format).

src/lightly_train/_data/yolo_object_detection_dataset.py
```diff
@@ -195,5 +147,5 @@ def list_image_info(self) -> Iterable[dict[str, str]]:
             }
 
     @staticmethod
-    def get_dataset_cls() -> type[YOLOObjectDetectionDataset]:
-        return YOLOObjectDetectionDataset
+    def get_dataset_cls() -> type[ObjectDetectionDataset]:
```
liopeer, line 150: It's now a bit confusing that the actual dataset is in the `object_detection_dataset.py` file and the arguments for it are distributed over 2 different files. Would not address it right now, but maybe in a quick follow-up you could move everything to a single file? Can also create a separate issue for it.

src/lightly_train/_commands/train_task.py
liopeer: What is the issue with having something like this additionally:
```python
class ObjectDetectionTrainTaskConfig(TrainTaskConfig):
    data: Annotated[
        Union[YOLOObjectDetectionDataArgs, COCOObjectDetectionDataArgs],
        Field(discriminator="format"),
    ]
    task: Literal["object_detection"] = "object_detection"

    @model_validator(mode="before")
    @classmethod
    def choose_union_member(cls, values):
        data = values.get("data")

        if isinstance(data, dict) and "format" not in data:
            # update the dict
        return values
```

simonschoelly: I had that before - then I got a comment from the copilot reviewer that this would break YAML parsing: https://github.com/lightly-ai/lightly-train/pull/677#discussion_r3017713957-permalink

I tried to vibe code a fix for that in the YAML parsing logic, but it looked extremely complicated.

CHANGELOG.md
simonschoelly: Yes, that's the plan.

src/lightly_train/_data/yolo_object_detection_dataset.py
simonschoelly: I don't mind doing that in a separate issue.

src/lightly_train/_commands/train_task.py
liopeer: Mmmh okay. Maybe you can re-name the validator in the original `TrainTaskConfig` and extend it here through an additional `super()` call. But if it's too complicated I'd also leave it the way it is for now, but add a clear `TODO` with the note that this should be in a validator somewhere.

simonschoelly: I already spend a few hours trying to fix that with the YAML parser - so for now I just added a TODO.

I think it is alright; the ability to ommit `format` is just a convenience to not break existing user code. I will leave that out from the documentation and in the future we might even remove it.

tests/_commands/test_train_task.py
simonschoelly: Not sure if I agree - but removed it for now. I can always come back here when I understand the code base better.

src/lightly_train/_data/coco_object_detection_dataset.py
```diff
@@ -0,0 +1,180 @@
...
+    to the annotation file's parent directory, optionally under ``train_data_dir``
+    or ``val_data_dir``.
+    """
+
+    # TODO: (Lionel, 08/25): Handle test set.
+    format: Literal["coco"] = "coco"
+    train_labels: Path
+    train_data_dir: Path | None = None
+    val_labels: Path
+    val_data_dir: Path | None = None
```
guarin: This should be nested as in our other configurations:
```
data={
    "train": {
        "images": "...",
        "annotations": "...",
    },
    "val": {
        "images": "...",
        "annotations": "...",
    },
}
```

See https://github.com/lightly-ai/lightly-train/blob/0b649e141f539ea5f238cd6273c284efe96e0eda/src/lightly_train/_data/mask_panoptic_segmentation_dataset.py#L304-L313 which follows also the COCO format.

src/lightly_train/_commands/train_task.py
guarin: I believe the "proper" way to do this is:
```
def infer_format(data: dict) -> dict:
    if isinstance(data, dict) and "format" not in data:
        data = {**data, "format": "yolo"}
    return data

class ObjectDetectionTrainTaskConfig(TrainTaskConfig):
    data: Annotated[
        Union[YOLOObjectDetectionDataArgs, COCOObjectDetectionDataArgs],
        Field(discriminator="format"),
        BeforeValidator(infer_format)
    ]
    task: Literal["object_detection"] = "object_detection"
```

Then you don't have to inject anything here and don't need the extra `choose_union_member` validator

guarin: This might be an even cleaner solution: https://github.com/pydantic/pydantic/discussions/7442#discussioncomment-10834375

simonschoelly: The issue is still this line of code: https://github.com/lightly-ai/lightly-train/blob/9697d618f7962a15593521be93793a215a58cd07/src/lightly_train/_commands/train_task.py#L1825

But I had incorrectly assumed by that we had this covered by tests - but that is not the case and I now verified that it would actually crash if we would load a yaml config, so I need to find another workaround.

simonschoelly: Done. I also added some logic to the YAML parsing, it is not ideal, but the best I could come up with.

src/lightly_train/_data/coco_object_detection_dataset.py
simonschoelly: Done

tests/helpers.py
```diff
@@ -713,6 +715,64 @@ def create_yolo_oriented_object_detection_dataset(
...
+            images.append(
+                {
+                    "id": idx,
+                    "file_name": img_path.name,
+                    "width": width,
+                    "height": height,
+                }
+            )
+            for ann in annotations_per_image[idx]:
+                annotations.append({"id": ann_id, "image_id": idx, **ann})
```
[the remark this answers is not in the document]
simonschoelly: That should be ok

src/lightly_train/_data/coco_object_detection_dataset.py
```diff
@@ -0,0 +1,184 @@
...
+
+        image_dir = self.labels.resolve().parent
+        if self.data_dir is not None:
+            image_dir /= self.data_dir
+
+        for image in labels_dict["images"]:
+            image_width_pixel = image["width"]
+            image_height_pixel = image["height"]
+            file_name = image["file_name"]
+            image_id = image["id"]
```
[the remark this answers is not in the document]
simonschoelly: I think is ok for now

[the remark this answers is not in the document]
guarin: ^feel free to ignore copilot. It is super nitpicky

PR #678, opened 2026-03-31
Distillationv3: Add convolutional detection heuristic

liopeer:
## What has changed and why?
- add simple heuristic to determine the type of backbone and adjust weight decay accordingly
- heuristic can be overriden with a method arg
- I actually don't like this approach:
    - I think the `ModelWrapper` classes should have their own heuristic/attribute (much safer, since we know the models to be expected)
    - the current heuristic could be used to not require all custom models to implement that attribute instead (i.e. only as a fallback in very few use cases)

## How has it been tested?
- adjusted existing unit tests

- [ ] Yes

- [ ] Yes

Will happen with single PR that updates entire documentation.

### DIFF
--- a/src/lightly_train/_methods/distillationv3/distillationv3.py
+++ b/src/lightly_train/_methods/distillationv3/distillationv3.py
@@ -489,7 +489,8 @@ def _is_probably_conv(wrapped_model: ModelWrapper) -> bool:
     logger.warning(
         "Could not determine whether the student model is convolutional or "
         "transformer-based. Defaulting to non-convolutional (transformer-style "
-        "training recipe). Set `student_is_conv` explicitly to suppress this warning."
+        "training recipe). Implement the `architecture_info` method in your wrapper "
+        "around the student model for accurate detection and better defaults."
     )
     return False
 


src/lightly_train/_methods/distillationv3/distillationv3.py
```diff
@@ -153,11 +208,20 @@ class DistillationV3LARSArgs(LARSArgs):
     trust_coefficient: float = 0.001
     eps: float = 1e-8
 
+    def resolve_auto(self, wrapped_model: ModelWrapper) -> None:
```
liopeer, line 156: Not sure if we want here 4e-2 in case of Transformers.

src/lightly_train/_methods/distillationv3/distillationv3.py
```diff
@@ -153,11 +208,20 @@ class DistillationV3LARSArgs(LARSArgs):
...
 
 class DistillationV3AdamWArgs(AdamWArgs):
     lr: float = 0.0005
     weight_decay: float = 0.04
 
+    def resolve_auto(self, wrapped_model: ModelWrapper) -> None:
+        if _is_probably_conv(wrapped_model=wrapped_model):
+            # Use a smaller weight decay for convolutional models, which typically
+            # benefit from a more regularized training recipe.
+            self.weight_decay = 1e-6
```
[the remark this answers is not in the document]
liopeer: Indeed an issue.

tests/_methods/distillationv3/test_distillationv3.py
```diff
@@ -0,0 +1,52 @@
...
+        ("timm/resnet50", True),
+        # Transformers: False
+        (dummy_dinov2_vit_model, False),
+        (dummy_dinov3_vit_model, False),
+        ("timm/vit_tiny_patch16_224", False),
+        # ConvNeXt-style (LayerNorm, transformer training recipe): False
+        (dummy_dinov3_convnext_model, False),
+        ("torchvision/convnext_tiny", False),
+        ("torchvision/convnext_small", False),
+        ("timm/convnext_tiny", False),
```
guarin, line 58: ConvNeXt are convolutional models and should return `True`

liopeer: Outdated.

src/lightly_train/_methods/distillationv3/distillationv3.py
guarin: Let's leave it for now, I would actually even consider removing LARS for DistillationV3 and only keep AdamW.

src/lightly_train/_methods/distillationv3/distillationv3.py
```diff
@@ -153,10 +213,19 @@ class DistillationV3LARSArgs(LARSArgs):
...
     eps: float = 1e-8
 
+    def resolve_auto(self, wrapped_model: ModelWrapper) -> None:
+        pass
+
 
 class DistillationV3AdamWArgs(AdamWArgs):
     lr: float = 0.0005
     weight_decay: float = 0.04
+    _student_is_conv: bool | Literal["auto"] = "auto"
```
guarin:
```suggestion
weight_decay: float | Literal["auto"] = "auto"
```

We usually set the value that is automatically set to "auto". I wouldn't add `_student_is_conv` as attribute to the `AdamWArgs` because it will be passed as argument to the `AdamW` optimizer on initialization.

src/lightly_train/_methods/distillationv3/distillationv3.py
```diff
@@ -144,6 +144,66 @@ def resolve_auto(
...
+    logger.warning(
+        "Could not determine whether the student model is convolutional or "
+        "transformer-based. Defaulting to non-convolutional (transformer-style "
+        "training recipe). Set `student_is_conv` explicitly to suppress this warning."
+    )
+    return False
+
+
+def _get_student_is_conv(wrapped_model: ModelWrapper) -> bool:
+    """Determine whether the student model uses a convolutional (batchnorm-based) training
```
guarin: We usually put helper functions at the bottom of the file. Would also consider renaming this to `_use_conv_weight_decay` to make it more descriptive.

src/lightly_train/_methods/distillationv3/distillationv3.py
liopeer: Fair point, that makes a lot more sense.

liopeer: Changed

src/lightly_train/_methods/distillationv3/distillationv3.py
liopeer: Done

src/lightly_train/_commands/train_helpers.py
```diff
@@ -290,13 +290,16 @@ def get_optimizer_args(
...
+    wrapped_model: ModelWrapper,
 ) -> OptimizerArgs:
     if isinstance(optim_args, OptimizerArgs):
         return optim_args
     optim_args = {} if optim_args is None else optim_args
     optim_args_cls = method_cls.optimizer_args_cls(optim_type=optim_type)
     logger.debug(f"Using optimizer '{optim_args_cls.type()}'.")
-    return validate.pydantic_model_validate(optim_args_cls, optim_args)
+    args = validate.pydantic_model_validate(optim_args_cls, optim_args)
+    args.resolve_auto(wrapped_model=wrapped_model)
```
guarin, line 301: Didn't notice this before but ideally we call `.resolve_auto()` also when `optim_args` is already an `OptimizerArgs` instance. I also saw this problem in other places in the `train_helpers.py` and `train_task_helpers.py`. Might be a good follow-up PR.

src/lightly_train/_methods/distillationv3/distillationv3.py
```diff
@@ -446,3 +455,63 @@ def load_state_dict(
...
+        return False
+    # Traditional CNNs (ResNet, ShuffleNet, YOLO, …) use convolutions and BatchNorm.
+    if has_convolution:
+        return True
+    # Nothing detected – log a warning and default to non-convolutional.
+    logger.warning(
+        "Could not determine whether the student model is convolutional or "
+        "transformer-based. Defaulting to non-convolutional (transformer-style "
+        "training recipe). Set `student_is_conv` explicitly to suppress this warning."
+    )
```
guarin, line 494: Missed this one, the setting is outdated. Would advice to set weight decay manually.

liopeer: Thx, well spotted!

src/lightly_train/_commands/train_helpers.py
liopeer: You're right. It's very unlikely to happen (people usually don't pass an `OptimizerArgs` instance), but we should indeed take care of that case. I'll create an issue for it.

liopeer: Done.

src/lightly_train/_methods/distillationv3/distillationv3.py
liopeer: Changed.

PR #679, opened 2026-03-31
Add Step-aware MixUp Augmentation

yutong-xiang-97:
## What has changed and why?

Add step-aware `MixUp` augmentation based on LTDETR setups.

Under the hood, this augmentation is implemented as a batch operation in the collate function with NumPy backend. It is however still interfaced as a transform class in `_transforms/mixup.py`.

The default values reflect the ones in the config file [here](http://github.com/lightly-ai/LT-DETR/blob/main/rtdetrv2_pytorch/configs/rtdetrv2/rtdetrv2_vitt16_1x_coco.yml). Epochs are converted to steps referencing our COCO training regime (6250 steps/epoch). The `beta_range` value is hard-coded to be `(0.45, 0.55)` as done in LTDETR.

Note that:

- We also fixed the "`requires_dataloader_reinitialization` setting state" issue (see [comments](https://github.com/lightly-ai/lightly-train/pull/675#discussion_r3010347714)). Specifically, we split the public interface to two, so that `requires_dataloader_reinitialization` only does the boolean comparison, while the additional `update_transform_active_status` operates the state changes.

## How has it been tested?

- Unit tests
- Manually

- [x] Yes

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,7 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
-- Add `MixUp` augmentation for LT-DETR object detection training. This is done by adding
-  `mixup` train transform args to DINOv2 and DINOv3 LT-DETR object detection models with
-  defaults `prob=0.5`, `step_start=15000`, and `step_stop=30000`.
+- Add `MixUp` augmentation for LT-DETR object detection training.
 
 ### Changed
 
--- a/src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
+++ b/src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
@@ -150,9 +150,7 @@ class DINOv2LTDETRObjectDetectionTrainTransformArgs(ObjectDetectionTransformArgs
     scale_jitter: DINOv2LTDETRObjectDetectionScaleJitterArgs | None = Field(
         default_factory=DINOv2LTDETRObjectDetectionScaleJitterArgs
     )
-    mixup: DINOv2LTDETRObjectDetectionMixUpArgs | None = Field(
-        default_factory=DINOv2LTDETRObjectDetectionMixUpArgs
-    )
+    mixup: DINOv2LTDETRObjectDetectionMixUpArgs | None = None
     # We use the YOLO format internally for now.
     bbox_params: BboxParams = Field(
         default_factory=lambda: BboxParams(
--- a/src/lightly_train/_task_models/dinov3_ltdetr_object_detection/transforms.py
+++ b/src/lightly_train/_task_models/dinov3_ltdetr_object_detection/transforms.py
@@ -148,9 +148,7 @@ class DINOv3LTDETRObjectDetectionTrainTransformArgs(ObjectDetectionTransformArgs
     scale_jitter: DINOv3LTDETRObjectDetectionScaleJitterArgs | None = Field(
         default_factory=DINOv3LTDETRObjectDetectionScaleJitterArgs
     )
-    mixup: DINOv3LTDETRObjectDetectionMixUpArgs | None = Field(
-        default_factory=DINOv3LTDETRObjectDetectionMixUpArgs
-    )
+    mixup: DINOv3LTDETRObjectDetectionMixUpArgs | None = None
     # We use the YOLO format internally for now.
     bbox_params: BboxParams = Field(
         default_factory=lambda: BboxParams(


src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -429,23 +431,49 @@ def get_metric_args(
     train_model_cls: type[TrainModel],
     metric_args: dict[str, Any] | TaskMetricArgs | None,
     data_args: TaskDataArgs,
+    train_transform_args: TaskTransformArgs,
```
guarin: I would actually skip this change and still compute train metrics even if mixup or any of the other augmentations are enabled. Train metrics are anyways always computed on images with stronger augmentations than validation metrics. Mixup etc. don't change that.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
```diff
@@ -110,6 +111,12 @@ class DINOv2LTDETRObjectDetectionScaleJitterArgs(ScaleJitterArgs):
     divisible_by: int | None = None
 
 
+class DINOv2LTDETRObjectDetectionMixUpArgs(MixUpArgs):
+    prob: float = 0.5
+    step_start: int = 15000
+    step_stop: int = 30000
```
guarin: I would suggest to disable them for now by default until they are tested. Also we should think how we ant to calculate the start/stop steps.

The default number of steps is here: https://github.com/lightly-ai/lightly-train/blob/22be9afbc7f46e9e6e35b0dd55ebc0ad9f00ba6b/src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py#L70-L72

However the user might ofc modify the number of training steps. So I guess ideally we would scale the number of mixup/copyblend/mosaic steps based on the user configured number of steps. Maybe best to do this with `resolve_auto` on the transform? 

Also, 15000 steps with batch size 16 (which is the default in LightlyTrain) is only 2 epochs. It should be 4 epochs. The difference comes from the fact that LT-DETR uses batch size 32.

src/lightly_train/_transforms/object_detection_transform.py
```diff
@@ -273,32 +276,35 @@ def __init__(
     ):
         super().__init__(split, transform_args)
         self.scale_jitter: BatchReplayCompose | None = None
+        self.mixup: MixUp | None = None
+        self._scale_jitter_args = transform_args.scale_jitter
+        self._mixup_args = transform_args.mixup
```
guarin: FYI you don't have to store these in the class, you can always access them with `self.transform_args.scale_jitter` and `self.transform_args.mixup`

src/lightly_train/_transforms/object_detection_transform.py
```diff
@@ -340,7 +378,10 @@ def __call__(self, batch: list[ObjectDetectionDatasetItem]) -> ObjectDetectionBa
             for item in batch
         ]
 
-        if self.scale_jitter is not None and self._is_scale_jitter_applied_at_step(
+        if self.mixup is not None and self._is_mixup_active_at_step(self._step):
```
guarin, line 385: Very minor: 
```suggestion
        if self._is_mixup_active_at_step(self._step):
```

becuase `_is_mixup_active_at_step` already checks for `if self.mixup is None`. Same for scale jitter.

src/lightly_train/_transforms/task_transform.py
```diff
@@ -82,3 +86,7 @@ def uses_step_dependent_worker_state(self) -> bool:
     def requires_dataloader_reinitialization(self) -> bool:
         """Signal when worker-visible collate state changed and the loader must reset."""
         return False
+
+    def update_transform_active_status(self) -> None:
+        """Update internal active-status state for step-aware transforms after a dataloader refresh."""
+        pass
```
guarin: Would call this `reinitialized_dataloader` to make it more obvious that it is linked to `requires_dataloader_reinitialization`.

src/lightly_train/_transforms/transform.py
```diff
@@ -197,6 +197,18 @@ def scale_range(self) -> tuple[float, float] | None:
         return None
 
 
+class MixUpArgs(PydanticConfig):
+    prob: float = Field(ge=0.0, le=1.0)
+    step_start: int = Field(ge=0)
+    step_stop: int = Field(gt=0)
```
guarin, line 203: Would allow `None` to disable stop

src/lightly_train/_transforms/object_detection_transform.py
yutong-xiang-97: This is mainly for correct typing of `self.mixup` and `self.scale_jitter`.

src/lightly_train/_transforms/task_transform.py
yutong-xiang-97: I would use a more specified name `sync_after_dataloader_reinitialization`. `reinitialized_dataloader` sounds like an attribute.

guarin: Agree, `mark_dataloader_as_reinitialized` could be another option.

src/lightly_train/_transforms/transform.py
yutong-xiang-97: Will fix in a follow-up issue.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
yutong-xiang-97: Will fix in a follow-up PR.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
```diff
@@ -110,6 +111,14 @@ class DINOv2LTDETRObjectDetectionScaleJitterArgs(ScaleJitterArgs):
     divisible_by: int | None = None
 
 
+class DINOv2LTDETRObjectDetectionMixUpArgs(MixUpArgs):
+    prob: float = 0.5
```
guarin, line 115: As suggested before, I would leave these transforms deactivated by default until we have benchmarked them. Otherwise we might accidentally change the training recipe if we release a new version for distillationv3.

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- Add `MixUp` augmentation for LT-DETR object detection training. This is done by adding
+  `mixup` train transform args to DINOv2 and DINOv3 LT-DETR object detection models with
+  defaults `prob=0.5`, `step_start=15000`, and `step_stop=30000`.
```
guarin: The steps here are incorrect. IMO it is also enough to just write that we added mixup augmentation. No need to describe the steps in detail.

PR #686, opened 2026-04-07
Simon yolo instance segmentation list image info parses labels

simonschoelly:
## What has changed and why?

This PR extracts the labels for YOLO instance segmentation datasets already when creating `image_info`. This is in preparation for COCO instance segmentation datasets support.

## How has it been tested?

Slightly refactored the tests.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/train_task.py
+++ b/src/lightly_train/_commands/train_task.py
@@ -43,6 +43,9 @@ from lightly_train._data.image_classification_dataset import (
     ImageClassificationMultilabelDataArgs,
 )
 from lightly_train._data.infinite_cycle_iterator import InfiniteCycleIterator
+from lightly_train._data.instance_segmentation_dataset import (
+    YOLOInstanceSegmentationDataArgs,
+)
 from lightly_train._data.mask_panoptic_segmentation_dataset import (
     MaskPanopticSegmentationDataArgs,
 )
@@ -51,9 +54,6 @@ from lightly_train._data.mask_semantic_segmentation_dataset import (
 )
 from lightly_train._data.task_data_args import TaskDataArgs
 from lightly_train._data.task_dataset import TaskDataset
-from lightly_train._data.yolo_instance_segmentation_dataset import (
-    YOLOInstanceSegmentationDataArgs,
-)
 from lightly_train._data.yolo_object_detection_dataset import (
     YOLOObjectDetectionDataArgs,
 )
--- a/src/lightly_train/_task_models/dinov2_eomt_instance_segmentation/train_model.py
+++ b/src/lightly_train/_task_models/dinov2_eomt_instance_segmentation/train_model.py
@@ -21,10 +21,10 @@ from torch.optim.optimizer import Optimizer
 
 from lightly_train import _torch_helpers
 from lightly_train._configs.validate import no_auto
-from lightly_train._data.task_data_args import TaskDataArgs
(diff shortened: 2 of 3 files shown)


src/lightly_train/_data/file_helpers.py
```diff
@@ -465,28 +462,25 @@ def open_yolo_instance_segmentation_label_numpy(
...
         polygons.append(polygon)
         bboxes.append(_bbox_from_polygon(polygon))
-    classes_np = np.array(classes, dtype=np.int64)
-    bboxes_np = np.stack(bboxes) if bboxes else np.zeros((0, 4), dtype=np.float64)
-    return polygons, bboxes_np, classes_np
+    return polygons, bboxes, classes
 
 
-def _bbox_from_polygon(polygon: NDArrayPolygon) -> NDArrayBBoxes:
+def _bbox_from_polygon(polygon: list[float]) -> list[float]:
```
yutong-xiang-97, line 472: Same as above.

src/lightly_train/_data/file_helpers.py
```diff
@@ -447,16 +445,15 @@ def open_yolo_object_detection_label(
     return bboxes, classes
 
 
-def open_yolo_instance_segmentation_label_numpy(
+def open_yolo_instance_segmentation_label(
```
yutong-xiang-97, line 448: Just want to get more context: is there a reason why we changed the I/O typing and removed NumPy backend?

tests/_data/test_yolo_instance_segmentation_dataset.py
yutong-xiang-97, line 1: From Copilot:

Reusing an existing task mmap file can now break instance-segmentation training with `KeyError: 'polygons'`. The dataset now assumes every `image_info` row already contains JSON-encoded `polygons`/`bboxes`/`class_labels` at instance_segmentation_dataset.py, but the mmap filename is still derived only from the resolved `data`/`out` paths at train_task_helpers.py, and existing files are reopened without any schema/version check at train_task_helpers.py. That means any mmap produced by the previous code path, where rows contained `label_path` instead, will be reused and crash as soon as `__getitem__` runs. I reproduced this locally with an old-format mmap row and got exactly `KeyError: 'polygons'`. This needs either cache invalidation/versioning or a backward-compatible fallback.

The removed end-to-end dataset test means the new `__getitem__` path is only indirectly covered now, so this kind of schema-compatibility regression is easy to miss.

src/lightly_train/_data/file_helpers.py
simonschoelly: This is the same as we did for object detection. In order to be able to use COCO and YOLO and to have a single `Dataset` class, we parse the label files when creating `image_info`.

See these previous PRs:
https://github.com/lightly-ai/lightly-train/pull/674
https://github.com/lightly-ai/lightly-train/pull/677

simonschoelly:
> Please validate the mmap file issue. Potential issue with logic changes should be either captured via a unit/integration test (like the one removed in this PR), or a manual debug training run (where you see people "testing manually"). You could use the compute machines with GPUs for this purpose.

The mmap issue that you get from Copilot is a schema change that would not be captured by by integratiin tests. This is a very rare edge case - given the general issues with mmap that we said we will handle at a later time, I don't think this is something we need to do now.

simonschoelly:
The integration test that I removed did avoid the mmap logig.

PR #687, opened 2026-04-07
Simon add coco instance segmentation dataset

simonschoelly:
## What has changed and why?

This PR adds support for COCO instance segmentation datasets.

Also, Copilot spotted an error where we for forgot to pass `image_mode` to `file_helpers.open_image_numpy` so I added that.

[Instance Segmentation - LightlyTrain documentation.pdf](https://github.com/user-attachments/files/26923565/Instance.Segmentation.-.LightlyTrain.documentation.pdf)

## How has it been tested?

Added a few unit tests and rewrote one of the integration tests

- [x] Yes

- [x] Yes

### DIFF
--- a/docs/source/instance_segmentation.md
+++ b/docs/source/instance_segmentation.md
@@ -56,10 +56,11 @@ if __name__ == "__main__":
         out="out/my_experiment",
         model="dinov3/vitl16-eomt-inst-coco", 
         data={
+            "format": "yolo",           # either "yolo" or "coco"
             "path": "my_data_dir",      # Path to dataset directory
             "train": "images/train",    # Path to training images
             "val": "images/val",        # Path to validation images
-            "names": {                  # Classes in the dataset                    
+            "names": {                  # Classes in the dataset
                 0: "background",
                 1: "car",
                 2: "bicycle",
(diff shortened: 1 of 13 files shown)


docs/source/instance_segmentation.md
```diff
@@ -173,29 +174,45 @@ fig.show()
...
+
+If you would like to skip specific classes during training, add their IDs to the
+optional `ignore_classes` list. The trainer omits these classes from loss computation
+and the exported model does not predict them.
+
+(instance-segmentation-data-yolo)=
+
+### YOLO format
+
+For the [YOLO](https://labelformat.com/formats/instance-segmentation/yolov8/) format,
```
yutong-xiang-97, line 204: The link is not working.

docs/source/instance_segmentation.md
```diff
@@ -284,22 +281,166 @@ If you would like to exclude images without label files from training, you can s
...
+        "skip_if_label_file_missing": True, # Skip images without label files.
+    }
+)
+```
+
+(instance-segmentation-data-coco)=
+
+### COCO format
+
+For the [COCO](https://labelformat.com/formats/instance-segmentation/coco/) format,
```
yutong-xiang-97, line 306: Same here. Link does not work.

yutong-xiang-97 (approved):
Approved for easy merge but there are two broken links.

Also could you quickly explain why we enforce the polygons to be split to `list[list[float]`?

yutong-xiang-97:
> Also, Copilot spotted an error where we for forgot to pass image_mode to file_helpers.open_image_numpy so I added that.

This is already fixed with #705.

simonschoelly:
> > Also, Copilot spotted an error where we for forgot to pass image_mode to file_helpers.open_image_numpy so I added that.
> 
> This is already fixed with #705.

Removed it during rebase.

simonschoelly:
> Approved for easy merge but there are two broken links.
> 
> Also could you quickly explain why we enforce the polygons to be split to `list[list[float]`?

So the reason for this is, that both YOLO and COCO support segments consisting of multiple polygons. But they store them in a different way:
- YOLO: Just has polygon - if there are multiple polygons then they just add a line of width zero. We already handled that case so far, by checking if a point is repeated.
- COCO: Stores multiple polygons separately.

In order to use the same code for drawing the masks, we now simply create a list of polygons 'list[list[float]]' - we still keep the logic though that one polygon there can be the YOLO style of polygon.

PR #694, opened 2026-04-09
Apply Mosaic transform in ObjectDetectionTransform

yutong-xiang-97:
## What has changed and why?

As the title suggests.

There are also updates to the integrations tests.

## How has it been tested?

- Unit test
- Manually

- [x] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_transforms/object_detection_transform.py
+++ b/src/lightly_train/_transforms/object_detection_transform.py
@@ -237,6 +237,8 @@ class ObjectDetectionTransform(TaskTransform):
         )
 
         # Mosaic setup.
+        self.individual_transforms_skip_zoomout_ioucrop: list[Any] = []
+        self.transform_skip_zoomout_ioucrop: Compose | None = None
         self.mosaic: MosaicTransform | None = None
         if transform_args.mosaic is not None:
             self.mosaic = MosaicTransform(
@@ -333,6 +335,7 @@ class ObjectDetectionTransform(TaskTransform):
 
         if self._should_apply_mosaic():
             assert self.mosaic is not None
+            assert self.transform_skip_zoomout_ioucrop is not None
             image, bboxes, class_labels = self.mosaic(image, bboxes, class_labels)
             # MosaicTransform clips boxes to the canvas but keeps degenerate boxes
             # (zero width/height). Filter them before passing to albumentations.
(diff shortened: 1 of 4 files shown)


tests/_transforms/test_object_detection_transform.py
```diff
@@ -171,9 +189,19 @@ def test___all_args_combinations(
         resize: ResizeArgs | None,
         random_iou_crop: RandomIoUCropArgs | None,
         normalize: NormalizeArgs | None,
+        mosaic: MosaicArgs | None,
     ) -> None:
         image_size = _get_image_size()
-        bbox_params = _get_bbox_params()
+        # MosaicTransform expects YOLO normalized bboxes.
+        if mosaic is not None:
```
liopeer: I would just directly use `"yolo"` in `_get_bbox_params()`. Since we don't check for actual values in the tests of this file you should not have to change anything else. It would much simplify the test.

Note: In the code we anyway use YOLO style bboxes anyway. The reason I usually went with `"pascal_voc"` in the tests is that they are integers (pixel values), which makes it easier to assert that values match, therefore a bit more convenient in tests.

src/lightly_train/_transforms/object_detection_transform.py
```diff
@@ -234,6 +236,80 @@ def __init__(
...
+                fill_value=int(transform_args.mosaic.fill_value),
+                max_cached_images=transform_args.mosaic.max_cached_images,
+                random_pop=transform_args.mosaic.random_pop,
+            )
+
+        # Build skip pipeline (without RandomZoomOut and RandomIoUCrop) for mosaic.
+        self._build_skip_pipeline()
+
+        # Step-aware state for mosaic scheduling.
+        self._step = 0
```
liopeer, line 259: I assume this is just a workaround now and will be merge with already existing `global_step` in the next PR.

liopeer (changes requested):
Overall looking good (assuming that the refactor of the "stepping" will still happen). However, I think the small issue with the tests is worth addressing.

src/lightly_train/_transforms/object_detection_transform.py
yutong-xiang-97: Yes.

tests/_transforms/test_oriented_object_detection_transform.py
liopeer: Now the values don't really make sense. I would either change here the values as well, or otherwise not change the format to `"yolo"`.

liopeer (approved):
Approved with a small comment.

PR #697, opened 2026-04-10
Refactor StopPolicy

yutong-xiang-97:
## What has changed and why?

Remove `StopPolicy` and use an `ActivationPolicy` to control the start and stop for all the sample-level and batch-level transformations, including
 - Refactoring the corresponding arguments and implementation of `ObjectDetectionTransform`
 - Adding the missing start and stop mechanism for `RandomPhotometricDistort`, `RandomZoomOut` and `RandomIoUCrop`

Also
- Make the step-aware augmentation transforms default to be used. Also fix the wrong defaults.
- Add comments to the `step_start` and `step_stop` parameter to explain the default setup.
- Update relevant tests and also refactor `test_object_detection_transfrom` accordingly.

## How has it been tested?

- Unit tests
- Manually

- [x] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_transforms/oriented_object_detection_transform.py
+++ b/src/lightly_train/_transforms/oriented_object_detection_transform.py
@@ -66,15 +66,6 @@ class OrientedObjectDetectionTransform(TaskTransform):
 
         self.transform_args: OrientedObjectDetectionTransformArgs = transform_args
 
-        self.stop_step = (
-            transform_args.stop_policy.stop_step if transform_args.stop_policy else None
-        )
-        self.global_step = 0
-        self.stop_ops = (
-            transform_args.stop_policy.ops if transform_args.stop_policy else set()
-        )
-        self.past_stop = False
-
         transforms_list: list[v2.Transform] = []
 
         if transform_args.channel_drop is not None:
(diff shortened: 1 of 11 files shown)


src/lightly_train/_transforms/transform.py
liopeer, line 196: What's the reason for not using inheritance here?

liopeer (approved):
LGTM! One small comment

src/lightly_train/_transforms/transform.py
yutong-xiang-97: ScaleJitter does not have `step_start`

src/lightly_train/_transforms/oriented_object_detection_transform.py
```diff
@@ -66,15 +66,6 @@ def __init__(
...
-        )
-        self.global_step = 0
-        self.stop_ops = (
-            transform_args.stop_policy.ops if transform_args.stop_policy else set()
-        )
-        self.past_stop = False
-
         transforms_list: list[v2.Transform] = []
 
         if transform_args.channel_drop is not None:
```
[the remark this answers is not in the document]
yutong-xiang-97: @guarin what are the configs for OBB? Might also need to create another issue for it.

src/lightly_train/_transforms/transform.py
liopeer: Ok. To me it would make more sense to still inherit and just set `step_start` to 0.

yutong-xiang-97: Yeah but I would prefer not exposing this interface to the users. I will leave a comment here.

PR #699, opened 2026-04-13
Allow specifying classes from a json file for semantic segmentation

simonschoelly:
## What has changed and why?

This PR allows one to specify the classes for semantic segmentation from a json file.

This was done in order to import semantic segmentation datasets exported from Lightly Studio.

For consistency, we might add the same functionality in the future for other tasks/datasets.

[Semantic Segmentation - LightlyTrain documentation.pdf](https://github.com/user-attachments/files/26703961/Semantic.Segmentation.-.LightlyTrain.documentation.pdf)

## How has it been tested?

Added two unit tests.

- [x] Yes

- [x] Yes

### DIFF
--- a/src/lightly_train/_data/mask_semantic_segmentation_dataset.py
+++ b/src/lightly_train/_data/mask_semantic_segmentation_dataset.py
@@ -330,8 +330,7 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
                 raise ValueError(f"Failed to read classes file '{path}': {e}") from e
             if not isinstance(data, dict):
                 raise ValueError(
-                    f"Expected '{path}' to contain a JSON object ({{...}}), "
-                    f"got {type(data).__name__}."
+                    f"Expected '{path}' to contain a JSON dict, got {type(data).__name__}."
                 )
             classes = {int(k): v for k, v in data.items()}
 
--- a/src/lightly_train/_data/mask_semantic_segmentation_dataset.py
+++ b/src/lightly_train/_data/mask_semantic_segmentation_dataset.py
@@ -7,13 +7,14 @@
 #
 from __future__ import annotations
 
+import json
 from collections.abc import Sequence
 from pathlib import Path
 from typing import Any, ClassVar, Dict, Iterable, Union
 
 import numpy as np
 import torch
-from pydantic import AliasChoices, Field, TypeAdapter, field_validator
+from pydantic import AliasChoices, Field, TypeAdapter, field_validator, model_validator
 from torch import Tensor
 
 from lightly_train._configs.config import PydanticConfig
@@ -306,8 +307,23 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
     ignore_index: ClassVar[int] = -100
     train: SplitArgs
     val: SplitArgs
-    classes: dict[int, ClassInfo]
+    classes: dict[int, ClassInfo] | None = None
+    classes_json: PathLike | None = None
     ignore_classes: set[int] | None = Field(default=None, strict=False)
+
+    @model_validator(mode="before")
+    @classmethod
+    def check_classes_xor_classes_json(cls, values: Any) -> Any:
+        has_classes = values.get("classes") is not None
+        has_classes_json = values.get("classes_json") is not None
+        if has_classes == has_classes_json:
+            raise ValueError("Exactly one of 'classes' or 'classes_json' must be set.")
+        if has_classes_json:
+            path = Path(values["classes_json"])
+            with open(path) as f:
+                data: dict[str, str] = json.load(f)
+            values["classes"] = {int(k): v for k, v in data.items()}
+        return values
 
     def train_imgs_path(self) -> Path:
         return Path(self.train.images)
--- a/src/lightly_train/_data/mask_semantic_segmentation_dataset.py
+++ b/src/lightly_train/_data/mask_semantic_segmentation_dataset.py
@@ -334,6 +334,7 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
                 )
             classes = {int(k): v for k, v in data.items()}
 
+        # We need strict=False, as otherwise the lists from JSON file cannot be coherced into a MultiChannelClassInfo.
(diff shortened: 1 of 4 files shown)


src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -306,9 +307,24 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
     ignore_index: ClassVar[int] = -100
     train: SplitArgs
     val: SplitArgs
-    classes: dict[int, ClassInfo]
+    classes: dict[int, ClassInfo] | None = None
+    classes_json: PathLike | None = None
```
guarin: In classification we support different file formats for the train/val/test directories and the way we did it is not to add fields like `train_dir` and `train_csv` but instead have a single `train` field that can either be a directory or a csv file. I would consider doing the same with the `classes` here and set them to `classes: dict[int, ClassInfo] | PathLike` and then check if the path is a `.json` file and load the classes from there.

simonschoelly: That is a good point, will do that.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -306,7 +307,7 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
     ignore_index: ClassVar[int] = -100
     train: SplitArgs
     val: SplitArgs
-    classes: dict[int, ClassInfo]
+    classes: dict[int, ClassInfo] | PathLike
```
guarin, line 310: I believe with `field_validator(mode="before")` you don't even have to allow `PathLike` for the field as it will be converted to a dict before the validator checks the type.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -317,12 +318,26 @@ def val_imgs_path(self) -> Path:
...
+            if path.suffix != ".json":
+                raise ValueError(f"'classes' path must be a .json file, got: '{path}'")
+            try:
+                with path.open(encoding="utf-8") as f:
+                    data = json.load(f)
+            except OSError as e:
+                raise ValueError(f"Failed to read classes file '{path}': {e}") from e
+            if not isinstance(data, dict):
+                raise ValueError(
+                    f"Expected '{path}' to contain a JSON object ({{...}}), "
```
guarin: Would give an actual example instead of `({{...}})` which seems wrong?

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -317,12 +318,26 @@ def val_imgs_path(self) -> Path:
...
+                    f"got {type(data).__name__}."
+                )
+            classes = {int(k): v for k, v in data.items()}
+
         classes_validated = TypeAdapter(
             Dict[int, Union[str, SingleChannelClassInfo, MultiChannelClassInfo]]
-        ).validate_python(classes)
+        ).validate_python(classes, strict=False)
```
guarin, line 340: Why is `strict=False` required?

guarin (changes requested):

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
simonschoelly: I think that could be true and would make a few things easier (these casts I have in my code), but then on the other hand, one would not realize, that one can pass in a `PathLike`, what do you prefer? I third option would have to have a separate function `get_classes` that would do the loading and validation - we can also add a `@cache` decorator on that so that won't do the calculation multiple times.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
simonschoelly: `({{...}}))` would render as `({...})`. I think putting a concrete example would make the error message a bit convoluted, so I just rephrased it without `({{...}}))`.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
simonschoelly: We need this, as otherwise it seems not be possible to correctly cohere the list of lists in multi class json file to a MultiChannelClassInfo. If you prefer we can already do that when reading  in the json file though.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
guarin: True, `classes: dict[int, ClassInfo] | PathLike` with a `get_classes` method or similar seems like the best option to me. The interface only has to define `included_classes`, see  https://github.com/lightly-ai/lightly-train/blob/main/src/lightly_train/_data/task_data_args.py

We also have some precedent for using caching here: https://github.com/lightly-ai/lightly-train/blob/5bcbbd050077b38986f2e11f01c5328710edeb82/src/lightly_train/_data/mask_panoptic_segmentation_dataset.py#L321-L323 Although that is a different situation where classes can only be loaded from the file.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
guarin: Oh ok, wasn't aware of the double `{{` syntax.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
guarin: Aah yes that makes sense. Might want to add a comment for this.

simonschoelly: Added a comment

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
simonschoelly: Let's keep it as it is. I tried @cached_property, but without the validator I run into some annoying issues with older Pydantic versions.

guarin (approved):

PR #702, opened 2026-04-15
Resolve default start and stop step based on total number of steps automatically.

yutong-xiang-97:
## What has changed and why?

As the title suggests. See the docstrings of `resolve_ltdetr_step_schedule` for more details how it is handled.

Also enabled augmentations by default.

## How has it been tested?

- Unit tests

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
+++ b/src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
@@ -16,6 +16,7 @@ from pydantic import Field
 from lightly_train._transforms.object_detection_transform import (
     ObjectDetectionTransform,
     ObjectDetectionTransformArgs,
+    resolve_ltdetr_step_schedule,
 )
 from lightly_train._transforms.transform import (
     CopyBlendArgs,
@@ -47,13 +48,11 @@ class DINOv2LTDETRObjectDetectionRandomPhotometricDistortArgs(
     saturation: tuple[float, float] = (0.5, 1.5)
     hue: tuple[float, float] = (-0.05, 0.05)
 
-    # Corresponds to the 4th epoch of the total training run.
-    # TODO (Yutong 04/26): Update step_start and step_stop based on the actual number of training steps.
-    step_start: int = 25_000
-    # Corresponds to the total_epochs - no_aug_epoch of the total training run.
+    # "auto" corresponds to the 4th epoch of the total training run.
+    step_start: int | Literal["auto"] = "auto"
+    # "auto" corresponds to the total_epochs - no_aug_epoch of the total training run.
     # None means photometric distort is always on.
-    # TODO (Yutong 04/26): Update step_start and step_stop based on the actual number of training steps.
-    step_stop: int | None = 375_000
+    step_stop: int | Literal["auto"] | None = "auto"
 
 
 class DINOv2LTDETRObjectDetectionRandomZoomOutArgs(RandomZoomOutArgs):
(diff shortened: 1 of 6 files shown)


src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
```diff
@@ -138,22 +139,22 @@ class DINOv2LTDETRObjectDetectionScaleJitterArgs(ScaleJitterArgs):
     prob: float = 1.0
     divisible_by: int | None = None
 
-    # Corresponds to the total_epochs - no_aug_epoch of the total training run.
+    # "auto" resolves to epoch total_epochs - no_aug_epoch. For shorter runs,
+    # no_aug_epoch is scaled following a certain rule. See :func:`resolve_ltdetr_step_schedule` for the full algorithm.
     # None means scale jitter is always on.
-    # TODO (Yutong 04/26): Update step_start and step_stop based on the actual number of training steps.
-    step_stop: int | None = 375_000
+    step_stop: int | Literal["auto"] | None = "auto"
```
simonschoelly, line 145: Is there a reason why we do not have `step_start` here?

yutong-xiang-97: Because in the LTDETR training recipe, there is only step_stop for `ScaleJitter` and I don't want to expose `step_start` to open the opportunity to change it.

src/lightly_train/_transforms/transform.py
```diff
@@ -65,19 +65,37 @@ class RandomFlipArgs(PydanticConfig):
...
             raise ValueError("activation policy requires step_start < step_stop.")
         return self
 
     def is_active(self, step: int) -> bool:
+        if self.step_start == "auto":
+            return False
         if step < self.step_start:
             return False
-        if self.step_stop is None:
+        if self.step_stop is None or self.step_stop == "auto":
```
simonschoelly, line 98: Can this happen? Don't we resolve this before?

src/lightly_train/_transforms/object_detection_transform.py
```diff
@@ -112,6 +114,162 @@ def resolve_incompatible(self) -> None:
...
+      ``copyblend`` use [``step_start_resolved``, ``step_stop_resolved``)
+    - ``mixup`` and ``mosaic`` use [``step_start_resolved``, ``step_flat_resolved``)
+    - ``scale_jitter`` only resolves ``step_stop_resolved``
+
+    If an augmentation's final integer window is empty, it is disabled instead
+    of clamped to a minimum length. In practice this means:
+    - ``step_stop <= step_start`` disables the corresponding augmentation field
+    - ``scale_jitter`` is disabled when its resolved auto ``step_stop <= 0``
+    """
+    steps_per_epoch = train_num_batches / gradient_accumulation_steps
```
simonschoelly, line 177: Can it not happen that `gradient_accumulation_steps > steps_per_epoch`? Is there a reason why we create a floating point value here?

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/transforms.py
```diff
@@ -222,9 +223,15 @@ class DINOv2LTDETRObjectDetectionTrainTransformArgs(ObjectDetectionTransformArgs
     scale_jitter: DINOv2LTDETRObjectDetectionScaleJitterArgs | None = Field(
         default_factory=DINOv2LTDETRObjectDetectionScaleJitterArgs
     )
-    mosaic: DINOv2LTDETRObjectDetectionMosaicArgs | None = None
-    mixup: DINOv2LTDETRObjectDetectionMixUpArgs | None = None
-    copyblend: DINOv2LTDETRObjectDetectionCopyBlendArgs | None = None
+    mosaic: DINOv2LTDETRObjectDetectionMosaicArgs | None = Field(
+        default_factory=DINOv2LTDETRObjectDetectionMosaicArgs
+    )
```
simonschoelly, line 228: If I understand this correctly, we now also enable mosaic by default?

yutong-xiang-97: Yes

src/lightly_train/_transforms/transform.py
yutong-xiang-97: It can happen exactly in the OD case where we can only resolve auto after we have `train_num_batches` info, which comes from the dataloader.

src/lightly_train/_transforms/object_detection_transform.py
yutong-xiang-97: It can happen but doesn't affect the calculation. If you mean `gradient_accumulation_steps > train_num_batches` i.e. `steps_per_epoch<1`, this (float) semantic should still hold because the conversion between step- and epoch-based schedule is fundamentally fractional. Early integer rounding would lose info and maybe get zero.

simonschoelly (approved):

PR #705, opened 2026-04-16
Cleanup augmentation update

yutong-xiang-97:
## What has changed and why?

- Replace `hasattr("resolve_step_schedule")` with an actual hook and defer creating the transform and collate function.
- Add unit tests for the step-aware augmentations.
- Remove TODOs for `scale_jitter.prob`

## How has it been tested?

- Unit tests

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/train_task.py
+++ b/src/lightly_train/_commands/train_task.py
@@ -1384,16 +1384,15 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
         )
         # Resolve step-based augmentation schedule now that dataloader length
         # and gradient accumulation steps are known.
-        if hasattr(train_transform_args, "resolve_step_schedule"):
-            train_transform_args.resolve_step_schedule(
-                total_steps=no_auto(config.steps),
-                train_num_batches=len(train_dataloader),
-                gradient_accumulation_steps=no_auto(config.gradient_accumulation_steps),
-            )
-            logger.debug(
-                f"Resolved train transform args after step schedule resolution: "
-                f"{helpers.pretty_format_args(train_transform_args.model_dump())}"
-            )
+        train_transform_args.resolve_step_schedule(
+            total_steps=no_auto(config.steps),
+            train_num_batches=len(train_dataloader),
+            gradient_accumulation_steps=no_auto(config.gradient_accumulation_steps),
+        )
+        logger.debug(
+            f"Resolved train transform args after step schedule resolution: "
+            f"{helpers.pretty_format_args(train_transform_args.model_dump())}"
+        )
 
         config.logger_args = helpers.get_logger_args(
             steps=config.steps,
--- a/src/lightly_train/_commands/train_task.py
+++ b/src/lightly_train/_commands/train_task.py
@@ -67,7 +67,6 @@ from lightly_train._train_task_state import (
     TrainTaskState,
 )
 from lightly_train._training_step_timer import CUDAUtilization, TrainingStepTimer
-from lightly_train._transforms.task_transform import TaskCollateFunction
 from lightly_train.types import PathLike
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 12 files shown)


src/lightly_train/_commands/train_task.py
```diff
@@ -1384,16 +1384,15 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
...
-            )
+        train_transform_args.resolve_step_schedule(
+            total_steps=no_auto(config.steps),
+            train_num_batches=len(train_dataloader),
+            gradient_accumulation_steps=no_auto(config.gradient_accumulation_steps),
+        )
+        logger.debug(
+            f"Resolved train transform args after step schedule resolution: "
+            f"{helpers.pretty_format_args(train_transform_args.model_dump())}"
+        )
```
guarin: Would only log transforms here and remove the logging from here: https://github.com/lightly-ai/lightly-train/blob/c382f8a87e0864bff4b8401f465045d53e5a1bc3/src/lightly_train/_commands/train_task_helpers.py#L412-L417

src/lightly_train/_commands/train_task.py
```diff
@@ -1384,16 +1384,15 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
...
-            )
-            logger.debug(
-                f"Resolved train transform args after step schedule resolution: "
-                f"{helpers.pretty_format_args(train_transform_args.model_dump())}"
-            )
+        train_transform_args.resolve_step_schedule(
+            total_steps=no_auto(config.steps),
+            train_num_batches=len(train_dataloader),
+            gradient_accumulation_steps=no_auto(config.gradient_accumulation_steps),
+        )
```
guarin: I see why this function is needed (classic ml-chicken-egg problem). However, it introduces a lot of complexity as things happen magically at a distance. This function affects the transform and the collate functions in a non-trivial way. I propose to make this behavior more evident in the future. Something like this should work:

```
train_dataset = helpers.get_dataset(
    ..., # Do NOT pass any transform
)
train_dataloader = helpers.get_train_dataloader(
    ..., # Do NOT pass any transform args and do NOT create a collate function inside
)

# This also call resolve_step_schedule (or ideally you merge it with resolve_auto)
train_transform_args, val_transform_args = helpers.get_transform_args(
    ...,
   total_steps=no_auto(config.steps),
   train_num_batches=len(train_dataloader),
   gradient_accumulation_steps=no_auto(config.gradient_accumulation_steps),
)

train_transform = ...
train_collate_fn = ...

# New methods that make it obvious that we set the transform and collate function
train_dataset.set_transform(train_transform)
train_dataloader.set_collate_fn(train_collate_fn)
```

Basically, we defer creating the transform and collate functions until we have all the information needed to initialize them. And then we use `set_transform` and `set_collate_fn` to update them which makes it obvious what the code is intended to achieve.

src/lightly_train/_data/instance_segmentation_dataset.py
```diff
@@ -50,12 +50,20 @@ def __init__(
...
         )
+        if transform is not None:
+            self._init_image_mode(transform)
 
+    def set_transform(self, transform: TaskTransform) -> None:
+        super().set_transform(transform)
+        assert isinstance(transform, InstanceSegmentationTransform)
+        self._init_image_mode(transform)
+
+    def _init_image_mode(self, transform: InstanceSegmentationTransform) -> None:
```
guarin, line 66: This method doesn't do anything

guarin (approved):

src/lightly_train/_data/instance_segmentation_dataset.py
yutong-xiang-97: Do you mean that we don't need to compute the `image_mode` for instance segmentation at all (i.e no multi-channel image support)? Or maybe we need to add `mode=self.image_mode` to

```python
image_np = file_helpers.open_image_numpy(image_path=image_path)
```

mrpositron (approved):
Approved by Guarin

PR #706, opened 2026-04-17
Better hash for data mmap hash

simonschoelly:
## What has changed and why?

This PR ensures that that the hash function that we compute caching image_info works better by

- Storing the timestamp for COCO image annotation files
- Add more fields for other dataset hash functions
- Ensure correct path resolution when computing the hashes

## How has it been tested?

Added some small tests for the COCO dataset hashes.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -489,17 +489,13 @@ def get_dataset_temp_mmap_path(
 ) -> Generator[Path, Any, Any]:
     """Generate file in temporary directory to be used for memory-mapping the dataset.
 
-    Creates a unique filename for the memory-mapped file based on the `out` or `data`
-    arguments. We use those arguments as they are consistent across all ranks on the
-    same node for the same run. Additionally, we can cache the file if required, since
-    the hash directly reflects the used config.
-
-    Use the same file on all ranks across all nodes, unless the filesystem is not shared.
+    Creates a unique filename for the memory-mapped file based on ``out`` and
+    ``data_hash``. The caller is responsible for producing a ``data_hash`` that
+    uniquely identifies the dataset configuration. The same file is used on all
+    ranks across all nodes, unless the filesystem is not shared.
     """
     if Env.LIGHTLY_TRAIN_MMAP_REUSE_FILE.value:
-        # Use data as identifier to share the mmap file across multiple runs.
-        # NOTE(Guarin, 09/25): Hash of data might be slow if data is a long list of
-        # filenames or directories.
+        # Use data_hash as identifier to share the mmap file across multiple runs.
         identifier = data_hash
     else:
         # Use out as identifier to create a unique mmap file for each run. We assume
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -484,7 +484,7 @@ def _unlink_and_ignore(path: Path) -> None:
 @contextlib.contextmanager
 def get_dataset_temp_mmap_path(
     fabric: Fabric,
-    data: PathLike,
+    data_hash: str,
     out: PathLike,
 ) -> Generator[Path, Any, Any]:
     """Generate file in temporary directory to be used for memory-mapping the dataset.
@@ -500,11 +500,11 @@ def get_dataset_temp_mmap_path(
         # Use data as identifier to share the mmap file across multiple runs.
         # NOTE(Guarin, 09/25): Hash of data might be slow if data is a long list of
         # filenames or directories.
-        identifier = str(Path(data).resolve())
+        identifier = data_hash
     else:
         # Use out as identifier to create a unique mmap file for each run. We assume
         # that only one run is using a specific out directory at a time.
-        identifier = str(Path(out).resolve()) + str(Path(data).resolve())
+        identifier = str(Path(out).resolve()) + data_hash
 
     mmap_filepath = (cache.get_data_cache_dir() / get_sha256(identifier)).with_suffix(
         ".mmap"
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -500,7 +500,7 @@ def get_dataset_temp_mmap_path(
     else:
         # Use out as identifier to create a unique mmap file for each run. We assume
         # that only one run is using a specific out directory at a time.
-        identifier = str(Path(out).resolve()) + data_hash
+        identifier = str((str(Path(out).resolve()), data_hash))
 
     mmap_filepath = (cache.get_data_cache_dir() / get_sha256(identifier)).with_suffix(
         ".mmap"
--- a/src/lightly_train/_data/image_classification_dataset.py
+++ b/src/lightly_train/_data/image_classification_dataset.py
@@ -118,6 +118,10 @@ class ImageClassificationDataArgs(TaskDataArgs):
                 self.classes,
                 self.label_delimiter,
                 sorted(self.ignore_classes) if self.ignore_classes else None,
+                self.classification_task,
+                self.csv_image_column,
+                self.csv_label_column,
+                self.csv_label_type,
             )
         )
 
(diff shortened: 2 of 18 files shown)


src/lightly_train/_data/image_classification_dataset.py
```diff
@@ -111,11 +111,25 @@ class ImageClassificationDataArgs(TaskDataArgs):
 
     classification_task: Literal["multiclass", "multilabel"]
 
-    def train_imgs_path(self) -> Path:
-        return Path(self.train)
+    def train_data_mmap_hash(self) -> str:
```
yutong-xiang-97, line 114: From Codex:

The new hash now tracks the dataset path and some label settings, but it still omits `classification_task`, `csv_image_column`, `csv_label_column`, and `csv_label_type`. Those fields directly change how `ImageClassificationDatasetArgs` parses the same CSV into cached `image_info`, so reusing the mmap file after changing any of them can produce stale labels. Add these parsing knobs to the hash tuple.

yutong-xiang-97 (commented):
Left a comment. I also think some comments from Codex make sense.

Make sure that you test this on multi-GPU setting, maybe using one of c01-c03.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -310,11 +310,25 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
...
-    def train_imgs_path(self) -> Path:
-        return Path(self.train.images)
+    def train_data_mmap_hash(self) -> str:
+        return str(
+            (
+                Path(self.train.images).resolve(),
+                Path(self.train.masks).resolve(),
+                self.classes,
+                sorted(self.ignore_classes) if self.ignore_classes else None,
+            )
```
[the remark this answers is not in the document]
simonschoelly: Fixed

src/lightly_train/_data/mask_panoptic_segmentation_dataset.py
```diff
@@ -312,11 +312,23 @@ class MaskPanopticSegmentationDataArgs(TaskDataArgs):
...
-    def val_imgs_path(self) -> Path:
-        return Path(self.val.images)
+    def val_data_mmap_hash(self) -> str:
+        return str(
+            (
+                Path(self.val.images).resolve(),
+                Path(self.val.masks).resolve(),
+                sorted(self.ignore_classes) if self.ignore_classes else None,
+            )
+        )
```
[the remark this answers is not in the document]
simonschoelly: Fixed

src/lightly_train/_data/image_classification_dataset.py
simonschoelly: Fixed

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -474,27 +474,23 @@ def _unlink_and_ignore(path: Path) -> None:
...
-        # NOTE(Guarin, 09/25): Hash of data might be slow if data is a long list of
-        # filenames or directories.
-        identifier = str(Path(data).resolve())
+        # Use data_hash as identifier to share the mmap file across multiple runs.
+        identifier = data_hash
     else:
         # Use out as identifier to create a unique mmap file for each run. We assume
         # that only one run is using a specific out directory at a time.
-        identifier = str(Path(out).resolve()) + str(Path(data).resolve())
+        identifier = str(Path(out).resolve()) + data_hash
```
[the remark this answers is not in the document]
simonschoelly: Not part of this task, but fixed it anyways.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
```diff
@@ -319,13 +319,27 @@ class MaskSemanticSegmentationDataArgs(TaskDataArgs):
...
     ignore_classes: set[int] | None = Field(default=None, strict=False)
 
+    def _deterministic_classes_str(self) -> str:
+        """Serialize classes deterministically for hashing.
+
+        ClassInfo contains set fields (e.g., labels: set[int]) whose str()
+        representation is not deterministic across processes due to hash
+        randomization. We serialize to JSON with sorted sets and keys instead.
+        """
+        classes = cast(Dict[int, ClassInfo], self.classes)
```
yutong-xiang-97: Why type casting here? If it has to be of type `Dict[int, ClassInfo]` and we need to satisfy mypy, we can use `if isinstance(xxx)` and raise or even an assertion since it is internal.

tests/_data/test_coco_object_detection_dataset.py
```diff
@@ -54,3 +55,23 @@ def test_list_image_info(self, tmp_path: Path) -> None:
...
             assert json.loads(image_info[2]["class_labels"]) == [0, 1]
+
+    def _make_args(self, tmp_path: Path) -> COCOObjectDetectionDataArgs:
+        create_coco_object_detection_dataset(tmp_path=tmp_path)
+        return COCOObjectDetectionDataArgs(
+            train=SplitArgs(annotations=tmp_path / "train.json", images=Path("train")),
+            val=SplitArgs(annotations=tmp_path / "val.json", images=Path("val")),
+        )
+
+    def test_mmap_hash_is_deterministic(self, tmp_path: Path) -> None:
```
yutong-xiang-97, line 69: These tests can be under `TestCOCOObjectDetectionMmapHash` like in other tests.

tests/_data/test_yolo_object_detection_dataset.py
```diff
@@ -83,3 +83,16 @@ def test_list_image_info__split_last(self, tmp_path: Path) -> None:
             class_labels = json.loads(info["class_labels"])
             assert bboxes == [[0.375, 0.5, 0.25, 0.5]]
             assert class_labels == [1]
+
+    def test_mmap_hash_is_deterministic(self, tmp_path: Path) -> None:
```
yutong-xiang-97, line 89: Same here. You can put under a `TestYOLOObjectDetectionMmapHash`

yutong-xiang-97 (approved):
Approved with a few comments.

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
simonschoelly: Mypy does not seem to understand the Pydantic validator. So cast (which is only used during type validation anyways) helps with this.

I don't have a strong opinion here and can also use `if isinstance(self.classes, dict):` but what would be the advantage of this?

tests/_data/test_yolo_object_detection_dataset.py
simonschoelly: Done

tests/_data/test_coco_object_detection_dataset.py
simonschoelly: Done

src/lightly_train/_data/mask_semantic_segmentation_dataset.py
yutong-xiang-97: Using `cast` merely to simplify type checks could cover underlying typing issues later on when more code is added. Therefore we typically avoid using `cast()` and use `if isinstance` checks / `#type: ignore` for this instead. For private code I am also OK with assertion since we don't want user errors there. 

This is also listed in Guarin's SKILL.md for AI reviews. Somehow I couldn't find it in the dev guide but I have clear impression that this was part of the LightlyTrain coding convention.

simonschoelly: Are you really sure? Because the implementation is this one
```
def cast(typ, val):
    """Cast a value to a type.

    This returns the value unchanged.  To the type checker this
    signals that the return value has the designated type, but at
    runtime we intentionally don't check anything (we want this
    to be as fast as possible).
    """
    return val
 ```
 
 I will try to change it though just to be sure.

yutong-xiang-97: Thank you for pointing that out. Was mixing the casting with `int()`. 

However, the main point that type casting covers future issues still holds. Using `cast` basically says that "trust me the type here must be `dict[xxx]"`, but what if at some point this assumption fails, and at runtime it could indeed be also `PathLike` and need to add logic to handle it? If the logic is overlooked, it should fail right at this point and say the typing is not correct, instead of at a later stage with a less obvious error. That is what `if-raise` or `assertion` can offer.

For this one it might be a lesser problem, but with classes for model modules like `DinoVisionTransformer` where typing is a semantic requirement it becomes more obvious.

yutong-xiang-97 (approved):

PR #707, opened 2026-04-20
Fix Training Crash Caused by Assertion Failure in `generalized_box_iou`

yutong-xiang-97:
## What has changed and why?

Fix Training Crash Caused by Assertion Failure in `generalized_box_iou` by

- Upcasting relevant decoder path
- Sanitizing the bboxes

## How has it been tested?

N/A

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_task_models/object_detection_components/box_ops.py
+++ b/src/lightly_train/_task_models/object_detection_components/box_ops.py
@@ -49,6 +49,23 @@ def box_iou(boxes1: Tensor, boxes2: Tensor):
     return iou, union
 
 
+def _sanitize_boxes_xyxy(boxes: Tensor) -> Tensor:
+    """Sanitize xyxy boxes so that they are non-degenerate and NaN-free.
+
+    Replaces NaN/inf with 0 and enforces ``x1 >= x0`` and ``y1 >= y0`` so that
+    downstream IoU computations do not produce NaN and the degenerate-box
+    assertion in :func:`generalized_box_iou` cannot trip.
+
+    This is a defensive safety net: the upstream code (decoder) should not
+    produce NaN predictions in the first place. When NaN is encountered here,
+    it indicates numerical instability upstream (e.g. fp16 overflow).
+    """
+    boxes = torch.nan_to_num(boxes, nan=0.0, posinf=0.0, neginf=0.0)
+    left_top = boxes[..., :2]
+    right_bottom = torch.max(boxes[..., 2:], left_top)
+    return torch.cat([left_top, right_bottom], dim=-1)
+
+
 def generalized_box_iou(boxes1, boxes2):
     """
     Generalized IoU from https://giou.stanford.edu/
@@ -58,10 +75,12 @@ def generalized_box_iou(boxes1, boxes2):
     Returns a [N, M] pairwise matrix, where N = len(boxes1)
     and M = len(boxes2)
     """
-    # degenerate boxes gives inf / nan results
-    # so do an early check
-    assert (boxes1[:, 2:] >= boxes1[:, :2]).all()
-    assert (boxes2[:, 2:] >= boxes2[:, :2]).all()
+    # Sanitize degenerate/NaN boxes defensively. This replaces the previous
+    # hard asserts so that transient numerical issues (e.g. fp16 overflow in
+    # predictions) don't crash training. Upstream code should still avoid
+    # producing NaN in the first place — this is a safety net, not a fix.
+    boxes1 = _sanitize_boxes_xyxy(boxes1)
+    boxes2 = _sanitize_boxes_xyxy(boxes2)
     iou, union = box_iou(boxes1, boxes2)
 
     lt = torch.min(boxes1[:, None, :2], boxes2[:, :2])
--- a/src/lightly_train/_task_models/object_detection_components/box_ops.py
+++ b/src/lightly_train/_task_models/object_detection_components/box_ops.py
@@ -15,9 +15,14 @@
 https://github.com/facebookresearch/detr/blob/main/util/box_ops.py
 """
 
+import logging
+
 import torch
 from torch import Tensor
 from torchvision.ops.boxes import box_area
+
+_logger = logging.getLogger(__name__)
+_invalid_bbox_warning_emitted = False
 
 
 def box_cxcywh_to_xyxy(x: Tensor) -> Tensor:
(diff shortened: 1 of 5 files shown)


src/lightly_train/_task_models/object_detection_components/box_ops.py
```diff
@@ -49,6 +49,23 @@ def box_iou(boxes1: Tensor, boxes2: Tensor):
     return iou, union
 
 
+def _sanitize_boxes_xyxy(boxes: Tensor) -> Tensor:
```
simonschoelly: I am just wondering if it is the right approach to sanitize these boxes. Because if there are still errors we might now silently swallow them.

src/lightly_train/_task_models/object_detection_components/box_ops.py
```diff
@@ -49,6 +49,23 @@ def box_iou(boxes1: Tensor, boxes2: Tensor):
...
+
+    Replaces NaN/inf with 0 and enforces ``x1 >= x0`` and ``y1 >= y0`` so that
+    downstream IoU computations do not produce NaN and the degenerate-box
+    assertion in :func:`generalized_box_iou` cannot trip.
+
+    This is a defensive safety net: the upstream code (decoder) should not
+    produce NaN predictions in the first place. When NaN is encountered here,
+    it indicates numerical instability upstream (e.g. fp16 overflow).
+    """
+    boxes = torch.nan_to_num(boxes, nan=0.0, posinf=0.0, neginf=0.0)
```
simonschoelly: Would it make sense to set `posinf` to a really large value and `neginf` to a really small value?

src/lightly_train/_task_models/object_detection_components/rtdetrv2_decoder.py
```diff
@@ -348,9 +348,18 @@ def forward(
...
+            # inverse_sigmoid(ref_points) before sigmoid is a known overflow
+            # hotspot in RTDETR-style decoders, especially with EMA weights
+            # that can drift into regions where the MLP output exceeds fp16's
+            # ~65K range.
+            with torch.amp.autocast(device_type=output.device.type, enabled=False):
+                output_fp32 = output.float()
+                bbox_delta = bbox_head[i](output_fp32)
+                inter_ref_bbox = F.sigmoid(
+                    bbox_delta + inverse_sigmoid(ref_points_detach.float())
+                )
```
simonschoelly, line 364: Don't you have to cast this back to the original precision here? Also for the autocast below.

src/lightly_train/_task_models/object_detection_components/box_ops.py
```diff
@@ -58,10 +75,12 @@ def generalized_box_iou(boxes1, boxes2):
...
     """
-    # degenerate boxes gives inf / nan results
-    # so do an early check
-    assert (boxes1[:, 2:] >= boxes1[:, :2]).all()
-    assert (boxes2[:, 2:] >= boxes2[:, :2]).all()
+    # Sanitize degenerate/NaN boxes defensively. This replaces the previous
+    # hard asserts so that transient numerical issues (e.g. fp16 overflow in
+    # predictions) don't crash training. Upstream code should still avoid
+    # producing NaN in the first place — this is a safety net, not a fix.
+    boxes1 = _sanitize_boxes_xyxy(boxes1)
```
simonschoelly: Not sure if this is an issue, but we only need to sanitize the ground trough boxes once.

src/lightly_train/_task_models/object_detection_components/box_ops.py
simonschoelly: I also wonder, if we should log a warning - although that needs an `if` statement. It might be possible to log this only once, so that we don't slow down training.

simonschoelly (commented):
* I wonder if we should still sanitize, as this might swallow inputs
* Is there an easy way to test this with a test that would fail before your changes? Probably not, as the CI/CD runs on CPU, right?

src/lightly_train/_task_models/object_detection_components/box_ops.py
liopeer: We can run GPU tests (they will run daily, not upon PR).

liopeer: Since this was a bug, I would also encourage to have a test. --> avoids that we re-introduce known bugs later on

simonschoelly:
Also from Claude - some of these things I should already have addressed in my comments:
                                                                                                                                                                                      
>  1. inter_ref_bbox dtype propagation. After the autocast block, inter_ref_bbox is now fp32 (since it was computed inside enabled=False). It feeds into ref_points_detach (line 371) and     
  ultimately into query_pos_head(ref_points_detach) on the next iteration. This means the query position MLP now receives fp32 input instead of fp16, which could subtly change behavior or
  reduce mixed-precision throughput. Is this intentional? You may want to cast inter_ref_bbox back to the original dtype after the block, or at least document that this is an accepted      
  tradeoff.                                                                  
>  2. score_head[i](output) still runs in fp16. The score head is left in autocast. If the overflow concern is about EMA weight drift in general, the score head could have similar issues. Is
   there evidence this is fine, or was it just not observed? Not necessarily a blocker, but worth a comment.                                                                                 
>  3. Eval path is not upcasted. The elif i == self.eval_idx branch (line 366-369) still uses inter_ref_bbox which is now fp32, so that's fine. But the bbox head computation that produced it
   is upcasted — this means inference is slightly slower. Acceptable, but worth noting since eval doesn't need the same numerical safety.

src/lightly_train/_task_models/object_detection_components/box_ops.py
yutong-xiang-97: I would move the sanitization to `HugarianMatcher.forward()` so that 1. we only sanitize predicted bboxes 2. use `posinf=1.0, neginf=0.0` because the boxes are normalized there.

src/lightly_train/_task_models/object_detection_components/box_ops.py
yutong-xiang-97: See above.

src/lightly_train/_task_models/object_detection_components/box_ops.py
yutong-xiang-97: Box degeneration happens inadvertently during training due to possible drifts, so it is not necessary a bug, just insufficiency in how it is handled. A unit test may validate the logic itself though, but since it is derived from RT-DETR, there was never a test for these code.

yutong-xiang-97: > I wonder if we should still sanitize, as this might swallow inputs

I would use it as a defensive mechanism for any possible overflow.

> I am just wondering if it is the right approach to sanitize these boxes. Because if there are still errors we might now silently swallow them.

What other errors could we have for bboxes after sanitization? Without sanitization it was simply an assertion error in the original code.

> I also wonder, if we should log a warning - although that needs an if statement. It might be possible to log this only once, so that we don't slow down training.

We can do logging.

liopeer: > A unit test may validate the logic itself though, but since it is copied from RT-DETR, there was never a test for these code.

It's fine that we never had one since we just copied the code. But I think now that we're fixing it we should also make sure that it never ever pops up again.

yutong-xiang-97:
> `inter_ref_bbox` dtype propagation. After the autocast block, inter_ref_bbox is now fp32 (since it was computed inside enabled=False). It feeds into ref_points_detach (line 371) and ultimately into query_pos_head(ref_points_detach) on the next iteration. This means the query position MLP now receives fp32 input instead of fp16, which could subtly change behavior or reduce mixed-precision throughput. Is this intentional? You may want to cast inter_ref_bbox back to the original dtype after the block, or at least document that this is an accepted tradeoff.

Fixed.

> `score_head[i](output)` still runs in fp16. The score head is left in autocast. If the overflow concern is about EMA weight drift in general, the score head could have similar issues. Is there evidence this is fine, or was it just not observed? Not necessarily a blocker, but worth a comment.

We only notice bbox issues atm. Would just leave it as-is unless it comes up.

> Eval path is not upcasted. The elif i == self.eval_idx branch (line 366-369) still uses inter_ref_bbox which is now fp32, so that's fine. But the bbox head computation that produced it is upcasted — this means inference is slightly slower. Acceptable, but worth noting since eval doesn't need the same numerical safety.

Fixed.

src/lightly_train/_task_models/object_detection_components/rtdetrv2_decoder.py
yutong-xiang-97: For `ref_points_detach` and evaluation I have to cast back. The rest I think is fine.

simonschoelly (approved):

PR #709, opened 2026-04-21
Object Detection Image Logging (Open for Discussion)

mrpositron:
## What has changed and why?

Added image logging to a local output directory during training and validation.

**Design choices:**
- **Training:** log images for the first few steps (up to 4 images).
- **Validation:** log images at every epoch for the first three steps.  
  - This required passing both `train_step` and `val_step` into the function `validation_step`.  
- **Visualization:**
  - Ground truth and predictions are plotted side by side.
  - Predicted bounding boxes are shown only if they exceed a threshold.
  - The number of predicted boxes per image is capped (configurable).
  - Class colors are fixed across training.
- **Distributed training:** logging is restricted to `rank=0` to prevent duplicate outputs and file write conflicts. Multi-GPU not yet validated.

**Modified files:**
- `src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py`: added logging integration.
- `src/lightly_train/_visualize/object_detection.py`: implemented side-by-side visualization.
- `src/lightly_train/_visualize/utils.py`: helper functions for plotting.

**TODO:**
- Add tests  
- Integrate image logging with MLflow, Weights & Biases, etc.  
- Validate on multi-GPU setup

train_step9_val_step2_img1

Example output: ground truth (left) vs. predictions (right), with consistent class coloring.

## How has it been tested?

I have implemented several tests. I will add them later.

- [ ] Yes

- [ ] Yes


src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
```diff
@@ -321,6 +321,8 @@ def validation_step(
         self,
         fabric: Fabric,
         batch: ObjectDetectionBatch,
+        val_step: int,
```
yutong-xiang-97: DINOv2 LT-DETR should also be supported.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -384,6 +415,22 @@ def validation_step(
         )
         self.val_metrics.update_with_predictions(results, targets)
 
+        # Track validation steps within epoch for visualization (first 3 steps only)
+        if val_step < 3:
```
yutong-xiang-97: Why `fabric.global_rank` is not used here?

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -481,6 +528,131 @@ def get_optimizer(
...
+            save_fname: Filename for saving the images. If None, images are not saved.
+
+        Returns:
+            List of PIL images with labels and predictions.
+        """
+        if fabric.global_rank != 0:
+            return []  # Only prepare visualization on rank 0 to avoid duplication
+
+        # Extract first 4 images for visualization
+        gt_images, gt_bboxes, gt_classes = (
```
yutong-xiang-97: Best move the data to CPU before any visualization.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -481,6 +528,131 @@ def get_optimizer(
...
+
+        if output_dir is None:
+            output_dir = Path.cwd()
+
+        # Save predictions
+        try:
+            predictions_dir = output_dir / f"visualizations_{split}"
+            predictions_dir.mkdir(parents=True, exist_ok=True)
+            for idx, pred_image in enumerate(debug_images):
+                pred_output_path = predictions_dir / f"{save_fname}_img{idx}.jpg"
```
yutong-xiang-97: `save_fname` already has `.jpg` in it.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -481,6 +528,131 @@ def get_optimizer(
...
+                pred_image.save(pred_output_path)
+        except Exception as e:
+            # Log the error but don't interrupt training.
+            print(f"Warning: Failed to save visualization images: {e}")
```
yutong-xiang-97: Use `logger` from `logger = logging.getLogger(__name__)` instead of print.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -384,6 +415,22 @@ def validation_step(
...
+                    fabric=fabric,
+                    batch=batch,
+                    results=results,
+                    orig_target_sizes=orig_target_sizes,
+                    split="val",
+                    save_fname=save_fname,
+                )
+            except Exception as e:
+                # Log the error but don't interrupt validation.
+                print(f"Warning: Failed to prepare visualization data: {e}")
```
yutong-xiang-97: Use `logger` from `logger = logging.getLogger(__name__)` instead of print.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -307,6 +313,29 @@ def training_step(
...
+                    fabric=fabric,
+                    batch=batch,
+                    results=results,
+                    orig_target_sizes=orig_target_sizes,
+                    split="train",
+                    save_fname=save_fname,
+                )
+            except Exception as e:
+                # Log the error but don't interrupt training.
+                print(f"Warning: Failed to prepare visualization data: {e}")
```
yutong-xiang-97: Use `logger` from `logger = logging.getLogger(__name__)` instead of print.

src/lightly_train/_visualize/object_detection.py
```diff
@@ -0,0 +1,162 @@
...
+        image_tensor = images[i].clone()
+        if mean is not None and std is not None:
+            image_tensor = denormalize_image(image_tensor, mean, std)
+
+        _, img_height, img_width = image_tensor.shape
+
+        # --- Ground truth side (left) ---
+        gt_image = torchvision_functional.to_pil_image(image_tensor)
+        gt_draw = ImageDraw.Draw(gt_image)
+        font = load_font(size=18)
```
yutong-xiang-97: This is loaded every iteration in the for loop. Consider moving it out.

src/lightly_train/_visualize/object_detection.py
```diff
@@ -0,0 +1,162 @@
...
+                        pred_draw, x1, y1, f"{class_name} {score:.2f}", color, font
+                    )
+
+        # Combine side by side
+        combined_width = gt_image.width + pred_image.width
+        combined = Image.new("RGB", (combined_width, gt_image.height))
+        combined.paste(gt_image, (0, 0))
+        combined.paste(pred_image, (gt_image.width, 0))
+
+        # Resize the height to 320 while maintaining aspect ratio (to save space)
```
yutong-xiang-97: Dead code

src/lightly_train/_visualize/__init__.py
```diff
@@ -0,0 +1,24 @@
+#
```
yutong-xiang-97, line 1: I would actually use `_plot` instead of `_visualize` to specify this is for plotting.

src/lightly_train/_visualize/object_detection.py
```diff
@@ -0,0 +1,162 @@
...
+
+from lightly_train._visualize.utils import (
+    denormalize_image,
+    draw_label,
+    get_class_color,
+    load_font,
+)
+
+
+def _cxcywh_to_xyxy(boxes: Tensor, w: int, h: int) -> Tensor:
```
yutong-xiang-97: Not to be addressed for now, but the same logic is also implemented in `object_detection_components` where we copied from RT-DETR repo. Maybe we need to extract useful helpers from there later on.

src/lightly_train/_visualize/utils.py
```diff
@@ -0,0 +1,225 @@
...
+    value = 0.95
+
+    # Convert HSV to RGB
+    r, g, b = colorsys.hsv_to_rgb(hue, saturation, value)
+
+    # Convert to 0-255 range
+    return (int(r * 255), int(g * 255), int(b * 255))
+
+
+def create_image_mosaic(images: list[PILImage], images_per_row: int = 2) -> PILImage:
```
yutong-xiang-97: The functions that are not used elsewhere should be kept private.

src/lightly_train/_task_models/train_model.py
```diff
@@ -74,7 +74,9 @@ def training_step(self, fabric: Fabric, batch, step: int) -> TaskStepResult:  #
         # Return dictionary with loss and metrics for logging.
         raise NotImplementedError()
 
-    def validation_step(self, fabric: Fabric, batch) -> TaskStepResult:  # type: ignore[no-untyped-def]
+    def validation_step(  # type: ignore[no-untyped-def]
+        self, fabric: Fabric, batch
```
liopeer: Make sure the `batch` is typed and remove the type ignore. You can e.g. create a new type in `types.py` to group all types of samples.

src/lightly_train/_visualize/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+from lightly_train._visualize.object_detection import (
+    plot_object_detection_comparison,
+)
+from lightly_train._visualize.utils import (
+    denormalize_image,
+    draw_label,
+    load_font,
+)
+
+__all__ = [
```
liopeer: I don't see the need for putting anything in this `__init__.py` file. We usually leave all init files empty (instead for the very top-level one `src/lightly_train/__init__.py`), since we prefer more explicit imports. And it might be that some of the 3rd party ones also have content (like one in the directory with the dinov2/dinov3 models).

src/lightly_train/_visualize/object_detection.py
liopeer: I would actually encourage to address this right now. This is the kind of helper that should absolutely be covered by a unit test (easy source of stupid bugs), and therefore should not be implemented twice.

liopeer: We can move such helpers to inside the `_data` subpackage. What I would not do right now though: replace all the other calls to this transform everywhere else in the codebase. But if we have an implementation (with a unit test) already somewhere else I would move that & use it here.

src/lightly_train/_visualize/object_detection.py
```diff
@@ -0,0 +1,162 @@
...
+    gt_classes: list[Tensor],
+    pred_bboxes: list[Tensor],
+    pred_classes: list[Tensor],
+    pred_scores: list[Tensor],
+    included_classes: dict[int, str],
+    score_threshold: float = 0.3,
+    max_pred_boxes: int = 32,
+    max_images: int | None = None,
+    mean: tuple[float, ...] | None = None,
+    std: tuple[float, ...] | None = None,
```
liopeer: FYI: We usually avoid setting defaults for internal functions (makes the code more readable). I would say arguments that can be `None` can often be considered an exception to that.

src/lightly_train/_visualize/object_detection.py
```diff
@@ -0,0 +1,162 @@
...
+) -> list[Image.Image]:
```
liopeer: We import classes directly, i.e. `from PIL.Image import Image` or (in this case probably better; you can see that elsewhere in the code): `from PIL.Image import Image as PILImage`.

src/lightly_train/_visualize/object_detection.py
```diff
@@ -0,0 +1,162 @@
...
+    for i in range(n):
+        image_tensor = images[i].clone()
+        if mean is not None and std is not None:
+            image_tensor = denormalize_image(image_tensor, mean, std)
+
+        _, img_height, img_width = image_tensor.shape
+
+        # --- Ground truth side (left) ---
+        gt_image = torchvision_functional.to_pil_image(image_tensor)
+        gt_draw = ImageDraw.Draw(gt_image)
```
liopeer: Also here probably: `PILDraw(gt_image)`.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -253,6 +262,10 @@ def set_train_mode(self) -> None:
         if self.model_args.backbone_freeze:
             self.model.freeze_backbone()
 
+    def on_validation_epoch_start(self) -> None:
```
liopeer: I don't see any issue with directly adding this to the `TrainModel` class. We'll definitely want this for the other train model instances as well.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
mrpositron: `save_fname` doesn't have .jpg in it. But I get that it is ambiguous. It is just a prefix that will be joined with the image `idx` to save it. I fixed it. It will be called `save_prefix_name` instead of `save_fname`

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
liopeer: Also better to keep Exception handling tight. We only want to handle exceptions that we actually expect, unexpected exceptions should not be handled and should actually make things fail.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
liopeer: Why are we only logging for the first 10 epochs?

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -481,6 +535,131 @@ def get_optimizer(
     def get_task_model(self) -> TaskModel:
         return self.model
 
+    def _prepare_visualization_data(
+        self,
+        fabric: Fabric,
+        batch: ObjectDetectionBatch,
+        results: Any,
```
liopeer: Is this untyped from because it's coming from RT-DETR? If yes, we should actually introduce a type for this.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -481,6 +535,131 @@ def get_optimizer(
...
+        orig_target_sizes: list[tuple[int, int]],
+        split: str = "val",
+        save_fname: str | None = None,
```
liopeer: Here, I definitely don't see a reason to use defaults.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -481,6 +535,131 @@ def get_optimizer(
...
+            included_classes=self.data_args.included_classes,
+            score_threshold=0.6,
+            max_pred_boxes=32,
+            max_images=max_images,
+            mean=mean,
+            std=std,
+        )
+
+        output_dir = None
+        for logger_instance in fabric.loggers:
```
liopeer: This is a hacky workaround to find the log dir. I would suggest that you do the saving of the actual predictions inside the training loop. There you have the output directory already available & this will make it usable for the other models as well.

src/lightly_train/_visualize/utils.py
```diff
@@ -0,0 +1,225 @@
...
+from pathlib import Path
+
+import torch
+from PIL import Image as PILImageClass
+from PIL import ImageDraw, ImageFont
+from PIL.Image import Image as PILImage
+from torch import Tensor
+
+
+def draw_label(
```
liopeer: Call it `draw_bbox_label` or similar. So that it's clear that it works only with bboxes.

src/lightly_train/_visualize/utils.py
```diff
@@ -0,0 +1,225 @@
...
+        # Convert to RGB if necessary
+        if image.mode != "RGB":
+            image = image.convert("RGB")
+
+        mosaic.paste(image, (x, y))
+
+    return mosaic
+
+
+def load_font(size: int = 14) -> ImageFont.FreeTypeFont | ImageFont.ImageFont:
```
liopeer: Again: Direct class imports.

liopeer (changes requested):

liopeer:
Would also encourage you to unit-test all the helper functions btw. But let's first make sure that the rest of the structure works!

src/lightly_train/_task_models/train_model.py
yutong-xiang-97: Yes also use `# type: ignore[xx]` sparingly. Everything should be typed unless it creates mess.

src/lightly_train/_visualize/object_detection.py
yutong-xiang-97: Yes, check our internal python dev guide for this.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
yutong-xiang-97: I would say we only log for first few steps to make things simple. Ultralytics does that too.

Also @mrpositron we don't use epoch in fine-tuning control flow. There is indeed a helper function to calculate epochs but that is for logging purposes.

src/lightly_train/_task_models/train_model.py
mrpositron: I will resolve this issue, although it existed prior to my changes. I wasn’t sure whether I should modify it.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
yutong-xiang-97: Yes, also use `Any` sparingly unless it is complex to type stuff.

src/lightly_train/_task_models/train_model.py
liopeer: Oh, indeed. It only changed because of formatting. But yes, while we're at it, it would be great to address it.

yutong-xiang-97: I think it is directly inherited from torch module, but yes enhance typing would be possible.

src/lightly_train/_visualize/object_detection.py
yutong-xiang-97: I would prefer it is in a follow-up issue rather than cluttering this PR. We can address them after everything finishes. Let @mrpositron focus on adding the functionality first.

mrpositron:
# Changes
* Refactored image saving logic into `_train_task_from_config()`.
* Introduced `VisualizationResult`, which is now passed through `TaskStepResult` and consumed within `_train_task_from_config()`.
* Made visualization-related functions and utilities private.
* Fixed typing inconsistencies.
* For validation only:
  * Adopted a design decision to overwrite existing images. -> Removed the need for a validation epoch.
* Since `val_step` is required, introduced a step parameter in `validation_step()` to avoid adding extra logic to count validation steps.

## Same to-do:
- Add tests
- Integrate image logging with MLflow, Weights & Biases, etc.
- Validate on multi-GPU setup

src/lightly_train/_commands/train_task.py
```diff
@@ -1690,7 +1698,7 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
                 # Reset GPU memory tracking before val phase.
                 timer.reset_gpu_max_memory("val")
 
-                train_model.on_validation_epoch_start()
+                # train_model.on_validation_epoch_start()
```
yutong-xiang-97: Dead code.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
yutong-xiang-97: This has not been addressed yet.

src/lightly_train/_task_models/train_model.py
```diff
@@ -69,14 +70,12 @@ class TrainModel(Module):
     # - predict_step
     # See: https://github.com/Lightning-AI/pytorch-lightning/blob/95f16c12fe23664ffa5198a43266f715717c6f45/src/lightning/fabric/wrappers.py#L47-L48
 
-    def training_step(self, fabric: Fabric, batch, step: int) -> TaskStepResult:  # type: ignore[no-untyped-def]
+    def training_step(self, fabric: Fabric, batch: Any, step: int) -> TaskStepResult:
```
yutong-xiang-97: Please do not use `Any` for typing. Either specify the BatchTypes, or you leave it as is. I think you can leave it here for now and let the `BatchType` go to a separate PR.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
```diff
@@ -321,6 +321,7 @@ def validation_step(
         self,
         fabric: Fabric,
         batch: ObjectDetectionBatch,
+        step: int,
```
yutong-xiang-97, line 353: Logic for DINOv2 LT-DETR should also be implemented.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
yutong-xiang-97: This has not been addressed yet.

src/lightly_train/_visualize/__init__.py
mrpositron: There is already python file called `src/lightly_train/_plot.py`. I did not want to touch it.

src/lightly_train/_task_models/train_model.py
yutong-xiang-97: Similar comment here: https://github.com/lightly-ai/lightly-train/pull/709#discussion_r3117984015 Btw, all types are under `src/types.py`.

src/lightly_train/_visualize/object_detection.py
yutong-xiang-97: This is not addressed

src/lightly_train/_task_models/train_model.py
mrpositron: Okay, I will revert these changes and keep it as an ignored type.

src/lightly_train/_visualize/__init__.py
yutong-xiang-97: `_plot.py` should rather be refactored to be included in `_plot/` as it is exactly the helper used for logging example images in pretraining (and it is exactly why we call it `_plot`. Please add a follow-up issue for this.

src/lightly_train/_task_models/train_model.py
```diff
@@ -118,6 +119,12 @@ def on_train_batch_end(self) -> None:
         pass
 
 
+@dataclass
+class VisualizationResult:
```
yutong-xiang-97: Why would you need a dataclass which only contains a list of images for visualization results? If you need to simply re-type it you could add it to `types.py`.

src/lightly_train/_commands/train_task.py
```diff
@@ -1618,6 +1618,14 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
             # Call the on_train_batch_end hook.
             train_model.on_train_batch_end()
 
+            # Save visualization images if produced by the training step.
+            if train_result.visualization is not None:
+                viz = train_result.visualization
+                viz_dir = out_dir / "visualizations_train"
```
yutong-xiang-97: I would prefer using nested directories like `image_examples/train`.

src/lightly_train/_commands/train_task.py
```diff
@@ -1708,8 +1717,17 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
...
                         val_result = train_model.validation_step(
-                            fabric=fabric, batch=val_batch
+                            fabric=fabric,
+                            batch=val_batch,
+                            step=val_step,
                         )
+                    # Save visualization images if produced by the validation step.
+                    if val_result.visualization is not None:
+                        viz = val_result.visualization
+                        viz_dir = out_dir / "visualizations_val"
```
yutong-xiang-97: Same here, I would prefer using nested directories like `image_examples/val`.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
yutong-xiang-97: OK let's push this to a next PR.

src/lightly_train/_commands/train_task.py
```diff
@@ -1618,6 +1618,14 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
             # Call the on_train_batch_end hook.
             train_model.on_train_batch_end()
 
+            # Save visualization images if produced by the training step.
```
yutong-xiang-97: From Codex:

`training_step()` runs once per microbatch inside the accumulation loop, but train_result is overwritten each time and only the last microbatch's visualization is saved after the loop. With `gradient_accumulation_steps > 1`, the first three logged training examples are not the first three dataloader batches requested in the spec; they are the last microbatches of optimizer steps 0, 1, and 2. Capture the first qualifying microbatch explicitly, or log from inside the accumulation loop using a real batch counter.

src/lightly_train/_visualize/object_detection.py
```diff
@@ -0,0 +1,157 @@
...
+
+    for i in range(n):
+        image_tensor = images[i].clone()
+        if mean is not None and std is not None:
+            image_tensor = _denormalize_image(image_tensor, mean, std)
+
+        _, img_height, img_width = image_tensor.shape
+
+        # --- Ground truth side (left) ---
+        gt_image = torchvision_functional.to_pil_image(image_tensor)
```
yutong-xiang-97: From Codex:

The spec requires three distinct artifact families: `train_labels_*`, `val_labels_*`, and `val_predictions_*`. The new visualization helper always returns one side-by-side GT/prediction image per sample, so training logs predictions when it should be labels-only, and validation never produces separate label and prediction artifacts. It also saves multiple per-sample files instead of one batch artifact per logged step.

Fix direction: render batch-level label grids and prediction grids separately, then have training return only the label grid while validation returns both grids.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -481,6 +520,104 @@ def get_optimizer(
     def get_task_model(self) -> TaskModel:
         return self.model
 
+    def _prepare_visualization_data(
```
yutong-xiang-97: From Claude Code:

Spec: "The visualization code must live outside the TrainModel classes. Instead, the visualization code must be able to create the label plots based on an ObjectDetectionBatch or any of the related batch types." The PR keeps ~100 lines of rendering prep inside `DINOv3LTDETRObjectDetectionTrain._prepare_visualization_data` (box scaling, mean/std lookup, rank-0 gating, result unpacking), and the step itself calls `_plot_object_detection_comparison` which returns PIL images embedded in `TaskStepResult.visualization`.

Fix direction: move all prep + plotting into `_visualize/object_detection.py` with a public entry point that accepts `batch: ObjectDetectionBatch`, raw `outputs/results`, `included_classes`, and the normalize spec. `TaskStepResult` should carry raw data (boxes/classes/scores + batch ref), not pre-rendered `PIL.Image` objects. Per spec this also enables sharing the module across tasks (TRN-1851/1852/1853/1854) and forwarding to Fabric loggers (TRN-1856).

src/lightly_train/_task_models/dinov2_eomt_panoptic_segmentation/train_model.py
```diff
@@ -355,7 +355,10 @@ def training_step(
         )
 
     def validation_step(
-        self, fabric: Fabric, batch: MaskPanopticSegmentationBatch
+        self,
+        fabric: Fabric,
+        batch: MaskPanopticSegmentationBatch,
+        step: int,
```
yutong-xiang-97, line 361: This is a breaking change and you should give a default for those without logging logic atm.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -307,20 +313,40 @@ def training_step(
...
             self.train_metrics.update_with_predictions(results, targets)
 
+        # Visualization of training labels (first 3 steps only)
+        visualization: VisualizationResult | None = None
+        if step < 3:
+            orig_target_sizes = batch["original_size"]
+            orig_target_sizes_tensor = torch.tensor(
+                orig_target_sizes, device=samples.device
+            )
+            results = self.model.postprocessor(
```
yutong-xiang-97: From Claude Code:

Training viz branch re-runs `self.model.postprocessor(...)` a second time even when `self.metric_args.train` is already true (lines 311-313 already compute results). Reuse results from the metrics branch when available to avoid a duplicate postprocessor call on every one of the first 3 training steps.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -481,6 +520,104 @@ def get_optimizer(
...
+                result = results[i]
+                if isinstance(result, dict):
+                    # Assuming result has 'boxes', 'labels', 'scores' keys.
+                    pred_bboxes.append(result.get("boxes", torch.empty(0, 4)))
+                    pred_classes.append(
+                        result.get("labels", torch.empty(0, dtype=torch.long))
+                    )
+                    pred_scores.append(result.get("scores", torch.empty(0)))
+                else:
+                    # Fallback: create empty tensors.
```
yutong-xiang-97: From Claude Code

`# Fallback: create empty tensors` branch guard against shapes the RT-DETR postprocessor never returns. Per repo conventions, drop the defensive branches — if the postprocessor contract changes, we want a loud failure, not empty viz. Same applies to the `if isinstance(r, dict) else r` in the CPU move at line 547-551.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -481,6 +520,104 @@ def get_optimizer(
...
+        debug_images = _plot_object_detection_comparison(
+            images=gt_images,
+            gt_bboxes=gt_bboxes,
+            gt_classes=gt_classes,
+            pred_bboxes=scaled_pred_bboxes,
+            pred_classes=pred_classes,
+            pred_scores=pred_scores,
+            included_classes=self.data_args.included_classes,
+            score_threshold=0.6,
+            max_pred_boxes=32,
```
yutong-xiang-97: From Claude Code

`score_threshold=0.6, max_pred_boxes=32, max_images=4` are hardcoded. The PR description says "number of predicted boxes per image is capped (configurable)" but it isn't. Surface them through `TrainModelArgs` (or a dedicated `VisualizationArgs`) so they round-trip through configs; otherwise tuning requires a code change.

src/lightly_train/_visualize/utils.py
```diff
@@ -0,0 +1,180 @@
...
+            ]
+        )
+
+    # Try to load the first available font
+    for font_path in font_paths:
+        if Path(font_path).exists():
+            try:
+                return ImageFont.truetype(font_path, size=size)
+            except Exception:
+                pass
```
yutong-xiang-97: From Claude Code:

`Exception: pass in _load_font` swallows everything. Narrow to `OSError/ImportError` or at minimum log once at `DEBUG` level so a broken system font is visible during debugging.

src/lightly_train/_visualize/object_detection.py
```diff
@@ -0,0 +1,157 @@
...
+        gt_draw = PILDraw(gt_image)
+
+        boxes = gt_bboxes[i]
+        class_ids = gt_classes[i]
+
+        if len(boxes) > 0:
+            boxes_xyxy = _cxcywh_to_xyxy(boxes, img_width, img_height)
+            for box, class_id in zip(boxes_xyxy, class_ids):
+                x1, y1, x2, y2 = box.tolist()
+                class_name = included_classes.get(int(class_id), f"Class {class_id}")
```
yutong-xiang-97: From Claude Code:

`f"Class {class_id}"` interpolates a 0-d tensor rather than an `int` (`class_id` hasn't been `int(...)`-cast on the fallback path). Use `int(class_id)` for a clean label.

mrpositron:
# Changes
- Extracted all visualization logic into a new `_visualize/` subpackage (`object_detection.py`, `utils.py`), decoupling it from `train_model.py`
- Replaced side-by-side layout with separate images per purpose:
  - `image_examples/train_labels_{step}.jpg` — ground-truth boxes for training steps 0–2
  - `image_examples/val_labels_{step}.jpg` — ground-truth boxes for the first 3 val batches per epoch
  - `image_examples/val_predictions_{step}.jpg` — plot predictions for the same val batches
- Image logging is restricted to `global_rank == 0` for distributed training compatibility
- Training captures images from the first microbatch only (correct behavior under gradient accumulation)

## Known limitations / follow-ups
- `viz_score_threshold`, `viz_max_pred_boxes`, `viz_max_images` are hardcoded in `__init__`; they should be exposed via the logger args / config system in a follow-up PR
- `batch` in `_task_models/train_model.py` `training_step`/`validation_step` is not typed with a `# type: ignore` workaround; proper typing requires a separate PR (?)
- No unit tests yet for the `_visualize/` helpers (`_cxcywh_to_xyxy`, `_render_grid`, `_denormalize_image`, etc.)
- MLflow / Weights & Biases integration not included
- Multi-GPU validation not yet verified

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -209,6 +217,10 @@ def __init__(
             loss_names=self.loss_names,
         )
 
+        self.viz_score_threshold = 0.1
```
yutong-xiang-97, line 224: Add a `# TODO (04/26, Nauryz): xxx` to explain your TODOs to improve this.

src/lightly_train/_visualize/object_detection.py
yutong-xiang-97: For now simply put it to `utils.py`

src/lightly_train/_visualize/object_detection.py
```diff
@@ -0,0 +1,211 @@
...
+                    draw.rectangle([x1, y1, x2, y2], outline=color, width=2)
+                    _draw_bbox_label(
+                        draw, x1, y1, f"{class_name} {score:.2f}", color, font
+                    )
+        pil_images.append(img)
+
+    return _render_grid(pil_images)
+
+
+def _render_grid(pil_images: list[PILImage]) -> PILImage:
```
yutong-xiang-97: You can put them in `utils.py` as you might want to reuse it for other tasks as well.

tests/_visualize/test_object_detection.py
```diff
@@ -0,0 +1,122 @@
...
+        result = object_detection._render_grid(images)
+
+        # Grid layout (2 cols): idx 0→(col=0,row=0), 1→(col=1,row=0),
+        #                        2→(col=0,row=1), 3→(col=1,row=1)
+        assert result.getpixel((5, 5)) == colors[0]  # top-left
+        assert result.getpixel((15, 5)) == colors[1]  # top-right
+        assert result.getpixel((5, 15)) == colors[2]  # bottom-left
+        assert result.getpixel((15, 15)) == colors[3]  # bottom-right
+
+    def test__render_grid__returns_rgb_image(self) -> None:
```
yutong-xiang-97: Did you check that if we allow non-RGB input for object detection?

tests/_visualize/test_utils.py
```diff
@@ -0,0 +1,171 @@
...
+        assert len(set(colors)) == len(colors)
+
+    def test__get_class_color_large_class_id(self) -> None:
+        r, g, b = utils._get_class_color(10_000)
+        assert 0 <= r <= 255
+        assert 0 <= g <= 255
+        assert 0 <= b <= 255
+
+
+class Test_load_font:
```
yutong-xiang-97: From Claude Code:

`test__load_font_returns_pil_font` and `test__load_font_different_sizes` only assert `font is not None`, but `_load_font` cannot return `None` — the final fallback is `ImageFont.load_default()`. The tests are trivially green.

Fix direction: assert that the returned object responds to getbbox / has the right type (`isinstance(font, (PILImageFont, PILFreeTypeFont))`), or remove these tests.

src/lightly_train/_task_models/dinov2_eomt_instance_segmentation/train_model.py
```diff
@@ -316,7 +316,10 @@ def training_step(
         )
 
     def validation_step(
-        self, fabric: Fabric, batch: InstanceSegmentationBatch
+        self,
+        fabric: Fabric,
+        batch: InstanceSegmentationBatch,
+        step: int = 0,
```
yutong-xiang-97: Just realized that since you've passed `step=val_step` you don't have to keep a default here. Same for the other `train_model.py`.

src/lightly_train/_commands/train_task.py
```diff
@@ -1708,8 +1718,20 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
...
+                            batch=val_batch,
+                            step=val_step,
+                        )
+                    if val_result.prediction_image is not None:
+                        viz_dir = out_dir / "image_examples"
+                        viz_dir.mkdir(parents=True, exist_ok=True)
+                        val_result.prediction_image.save(
+                            viz_dir / f"val_predictions_{val_step}.jpg"
                         )
+                        if val_result.label_image is not None:
```
yutong-xiang-97: From Claude Code

The label save is nested under `if val_result.prediction_image is not None`. Today both fields are set together in dinov3_ltdetr, but TaskStepResult declares them as independent Optional fields, so any future model that returns labels without predictions will silently never persist them.

Fix direction: hoist `viz_dir = ...; viz_dir.mkdir(...)` and the label/prediction blocks to siblings, each guarded by their own is not None check.

src/lightly_train/_commands/train_task.py
```diff
@@ -1708,8 +1718,20 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
...
-                            fabric=fabric, batch=val_batch
+                            fabric=fabric,
+                            batch=val_batch,
+                            step=val_step,
+                        )
+                    if val_result.prediction_image is not None:
+                        viz_dir = out_dir / "image_examples"
+                        viz_dir.mkdir(parents=True, exist_ok=True)
+                        val_result.prediction_image.save(
+                            viz_dir / f"val_predictions_{val_step}.jpg"
```
yutong-xiang-97, line 1730: I remember that you mentioned having fixed the issue, but Claude Code still complains. Maybe I missed something on my end. What was the decision how we keep these images?

`val_predictions_{val_step}.jpg` is overwritten on every validation cycle, so only the final cycle's predictions remain on disk. Combined with the "save labels once" optimization, you keep the GT but lose the training‑progression view.
Fix direction: include the global step in the prediction filename (e.g., `val_predictions_step{step:06d}_{val_step}.jpg`) so successive cycles are preserved. Alternatively, document the overwrite‑by‑design behavior with a one‑line comment so it isn't read as a bug. (No code change required if the overwrite is intended.)

src/lightly_train/_visualize/utils.py
```diff
@@ -0,0 +1,182 @@
...
+    value = 0.95
+
+    # Convert HSV to RGB
+    r, g, b = colorsys.hsv_to_rgb(hue, saturation, value)
+
+    # Convert to 0-255 range
+    return (int(r * 255), int(g * 255), int(b * 255))
+
+
+def _load_font(size: int) -> PILImageFont | PILFreeTypeFont:
```
yutong-xiang-97: From Claude Code


`_load_font` walks the platform path list and calls `ImageFont.truetype` on every visualization call. The result is invariant for a given size — caching it would noticeably reduce per‑step overhead in the OD viz path.

Fix direction: wrap with `functools.lru_cache(maxsize=None)` (PIL fonts are not size‑mutable, so cache by size)

src/lightly_train/_visualize/utils.py
```diff
@@ -0,0 +1,182 @@
...
+            ]
+        )
+
+    # Try to load the first available font
+    for font_path in font_paths:
+        if Path(font_path).exists():
+            try:
+                return ImageFont.truetype(font_path, size=size)
+            except OSError as e:
+                logging.debug("Failed to load font %s: %s", font_path, e)
```
yutong-xiang-97: Use `logger = logging.getLogger(__name__)` and `logger.debug` instead of `logging.debug` directly.

src/lightly_train/_visualize/object_detection.py
```diff
@@ -0,0 +1,211 @@
...
+    boxes_xyxy = boxes.clone()
+    cx, cy, bw, bh = boxes[:, 0], boxes[:, 1], boxes[:, 2], boxes[:, 3]
+    boxes_xyxy[:, 0] = (cx - bw / 2) * w
+    boxes_xyxy[:, 1] = (cy - bh / 2) * h
+    boxes_xyxy[:, 2] = (cx + bw / 2) * w
+    boxes_xyxy[:, 3] = (cy + bh / 2) * h
+    return boxes_xyxy
+
+
+def plot_object_detection_labels(
```
yutong-xiang-97, line 26: This is a public interface and you should really test it. You can manually create batches like shown in `tests/_transforms/test_copyblend.py`

src/lightly_train/_visualize/object_detection.py
```diff
@@ -0,0 +1,211 @@
...
+                )
+                color = _get_class_color(int(class_id))
+                draw.rectangle([x1, y1, x2, y2], outline=color, width=2)
+                _draw_bbox_label(draw, x1, y1, class_name, color, font)
+        pil_images.append(img)
+
+    return _render_grid(pil_images)
+
+
+def plot_object_detection_predictions(
```
yutong-xiang-97, line 85: Same here. This needs tests.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -307,20 +319,34 @@ def training_step(
...
 
+        label_image: PILImage | None = None
+        if step < 3 and fabric.global_rank == 0:
+            normalize_mean = (
+                tuple(self._normalize.mean) if self._normalize is not None else None
+            )
+            normalize_std = (
+                tuple(self._normalize.std) if self._normalize is not None else None
+            )
+            label_image = plot_object_detection_labels(
```
yutong-xiang-97, line 334: Now I see why we would need to put the plotting logic outside `TrainModel` with this comment below.

In training_step, `plot_object_detection_labels` is invoked on every gradient‑accumulation microbatch when `step < 3` and `rank == 0`, but the caller in `train_task.py:1613` only keeps `acc_step == 0`. With `gradient_accumulation_steps=8`, you do 8× the (CPU‑side) PIL rendering and font loading per outer step.

And I also figured out a possible way of doing this. See DM.

yutong-xiang-97 (commented):
Added more reviews. Now I see more clearly why visualization code must live outside the `TrainModel` classes. For the tests, I would also love to see tests for the two public plotting functions.

tests/_visualize/test_object_detection.py
mrpositron: Yeah, it should work with non-RGB images too. I will fix it

src/lightly_train/_visualize/utils.py
```diff
@@ -37,6 +33,7 @@ def _draw_bbox_label(
     padding = 4
 
     # Measure text
+    font = ImageFont.load_default(size=20)
```
yutong-xiang-97: You can instead make this a constant so that it doesn't have to be loaded every single time.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
```diff
@@ -302,15 +315,31 @@ def training_step(
             orig_target_sizes_tensor = torch.tensor(
                 orig_target_sizes, device=samples.device
             )
-            results = self.model.postprocessor(
+            results: list[dict[str, Tensor]] = self.model.postprocessor(
```
yutong-xiang-97, line 318: Is this needed?

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
yutong-xiang-97, line 321: If the typing is needed this should raise a mypy error.

src/lightly_train/_visualize/object_detection.py
```diff
@@ -0,0 +1,162 @@
...
+        draw = PILDraw(img)
+
+        result = results[i]
+        boxes = result["boxes"]
+        class_ids = result["labels"]
+        scores = result["scores"]
+
+        if len(boxes) > 0:
+            # Scale predicted boxes from original image coordinates to tensor dimensions.
+            orig_width, orig_height = orig_target_sizes[i]
```
yutong-xiang-97, line 144: The box scaling logic should be tested here.

tests/_visualize/test_object_detection.py
```diff
@@ -0,0 +1,245 @@
...
+            max_images=1,
+            score_threshold=0.5,
+            max_pred_boxes=10,
+        )
+        assert result.size == (32, 32)
+
+    def test_plot_object_detection_predictions__score_threshold_filters_low_scores(
+        self,
+    ) -> None:
+        # All scores (0.3) below threshold (0.5) → no boxes drawn.
```
yutong-xiang-97: For this you need to test the other case where the boxes are actually drawn. For now you only test the output sizes but not boxes.

tests/_visualize/test_object_detection.py
```diff
@@ -0,0 +1,245 @@
...
+            max_images=1,
+            score_threshold=0.5,
+            max_pred_boxes=10,
+        )
+        assert result.size == (32, 32)
+
+    def test_plot_object_detection_predictions__max_pred_boxes_limits_drawn_boxes(
+        self,
+    ) -> None:
+        # 10 boxes with max_pred_boxes=3; only top 3 by score should be drawn.
```
yutong-xiang-97: Same here. How do you verify there are three boxes indeed?

tests/_visualize/test_object_detection.py
```diff
@@ -0,0 +1,245 @@
...
+            batch=batch,
+            results=results,
+            included_classes={0: "cat"},
+            max_images=1,
+            score_threshold=0.5,
+            max_pred_boxes=3,
+        )
+        assert result.size == (32, 32)
+
+    def test_plot_object_detection_predictions__unknown_class_uses_fallback(
```
yutong-xiang-97: Same here. How do you verify the fallback is actually used?

tests/_visualize/test_object_detection.py
```diff
@@ -0,0 +1,245 @@
...
+        batch = _make_batch()
+        results = _make_results()
+        result = object_detection.plot_object_detection_predictions(
+            batch=batch,
+            results=results,
+            included_classes={},
+            max_images=1,
+            score_threshold=0.5,
+            max_pred_boxes=10,
+            mean=(0.485, 0.456, 0.406),
```
yutong-xiang-97: How would mean and std affect your result then?

tests/_visualize/test_object_detection.py
```diff
@@ -0,0 +1,245 @@
...
+        results = _make_results(boxes_per_image=0, img_h=32, img_w=32)
+        result = object_detection.plot_object_detection_predictions(
+            batch=batch,
+            results=results,
+            included_classes={},
+            max_images=1,
+            score_threshold=0.5,
+            max_pred_boxes=10,
+        )
+        assert result.size == (32, 32)
```
yutong-xiang-97: What does "empty boxes" mean and how do you verify it?

tests/_visualize/test_object_detection.py
```diff
@@ -0,0 +1,245 @@
...
+    def test_plot_object_detection_labels__with_bboxes(self) -> None:
+        bboxes = [torch.tensor([[0.5, 0.5, 0.4, 0.4]])]
+        classes = [torch.tensor([1], dtype=torch.long)]
+        batch = _make_batch(bboxes=bboxes, classes=classes)
+        result = object_detection.plot_object_detection_labels(
+            batch=batch,
+            included_classes={1: "dog"},
+            max_images=1,
+        )
+        assert result.size == (32, 32)
```
yutong-xiang-97: The test does not check bboxes at all.

tests/_visualize/test_object_detection.py
```diff
@@ -0,0 +1,245 @@
...
+    def test_plot_object_detection_labels__unknown_class_uses_fallback(self) -> None:
+        bboxes = [torch.tensor([[0.5, 0.5, 0.4, 0.4]])]
+        classes = [torch.tensor([99], dtype=torch.long)]
+        batch = _make_batch(bboxes=bboxes, classes=classes)
+        result = object_detection.plot_object_detection_labels(
+            batch=batch,
+            included_classes={},
+            max_images=1,
+        )
+        assert result.size == (32, 32)
```
yutong-xiang-97: The test does not check if the fallback is used.

tests/_visualize/test_object_detection.py
```diff
@@ -0,0 +1,245 @@
...
+    def test_plot_object_detection_labels__with_mean_std(self) -> None:
+        batch = _make_batch()
+        result = object_detection.plot_object_detection_labels(
+            batch=batch,
+            included_classes={},
+            max_images=1,
+            mean=(0.485, 0.456, 0.406),
+            std=(0.229, 0.224, 0.225),
+        )
+        assert result.size == (32, 32)
```
yutong-xiang-97: The test does not check how mean and std would affect the result.

tests/_visualize/test_utils.py
```diff
@@ -0,0 +1,255 @@
...
+        assert result.getpixel((5, 15)) == colors[2]  # bottom-left
+        assert result.getpixel((15, 15)) == colors[3]  # bottom-right
+
+    def test__render_grid__returns_rgb_image(self) -> None:
+        images = [Image.new("RGB", (4, 4)) for _ in range(2)]
+        result = utils._render_grid(images)
+        assert result.mode == "RGB"
+
+
+class Test_draw_bbox_label:
```
yutong-xiang-97: The presence of the text pixels are not tested here.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -209,6 +217,12 @@ def __init__(
             loss_names=self.loss_names,
         )
 
+        # TODO (Nauryz, 04/26): These visualization thresholds are currently hardcoded, but we may want to make them configurable in the future (with logger_args).
+
+        self.viz_score_threshold = 0.1
+        self.viz_max_pred_boxes = 32
+        self.viz_max_images = 4
```
yutong-xiang-97, line 226: In the future where you create an `xxxArgs` class for these params with Pydantic, make sure it is a positive integer. Otherwise your current `_render_grid()` does not properly handle the case when num of images is zero.

yutong-xiang-97 (commented):
Make some comments mainly on the tests. 

Also be sure that you update the integration tests for object detection with the checks for logging (I believe there is something like `test_train_object_detection_xxx` in `test_train_task.py` or anywhere `.train_object_detection()` is used).

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
mrpositron: The RT-DETR postprocessor returns `list[dict[str, Tensor]] | tuple[Tensor, Tensor, Tensor]`

Here, I annotate it as `list[dict[str, Tensor]]` because later I assume that specific type (in `plot_object_detection_predictions`)

tests/_visualize/test_utils.py
mrpositron: The current tests cover the placement logic, whether the label goes above or below the bounding box. Testing that PIL actually renders the text pixels would be testing PIL itself, not our code.

tests/_visualize/test_object_detection.py
```diff
@@ -70,176 +91,272 @@ def _make_results(
...
+        batch = _make_batch_from_image(image=torch.zeros(1, 3, 32, 32))
         result = object_detection.plot_object_detection_labels(
             batch=batch,
             included_classes={},
             max_images=1,
+            mean=(0.5, 0.5, 0.5),
+            std=(0.5, 0.5, 0.5),
         )
-        assert result.size == (32, 32)
+        assert result.getextrema() != ((0, 0), (0, 0), (0, 0))
```
yutong-xiang-97: This assertion is too loose. It only proves some pixel is nonzero but nothing checks the denormalization math. Consider using a deterministic pixel value (e.g. 0) and verify the value matches the math (e.g with mean=std=0.5 for RGB you should get 128.

tests/_visualize/test_object_detection.py
```diff
@@ -70,176 +91,272 @@ def _make_results(
...
+            results=_make_results(batch_size=1, boxes_per_image=0, img_h=32, img_w=32),
             included_classes={},
             max_images=1,
             score_threshold=0.5,
             max_pred_boxes=10,
+            mean=(0.5, 0.5, 0.5),
+            std=(0.5, 0.5, 0.5),
         )
-        assert result.size == (32, 32)
+        assert result.getextrema() != ((0, 0), (0, 0), (0, 0))
```
yutong-xiang-97: Same here. Nothing validates the denormalize logic.

tests/_visualize/test_object_detection.py
```diff
@@ -70,176 +91,272 @@ def _make_results(
...
+        )
         result = object_detection.plot_object_detection_labels(
-            batch=batch,
-            included_classes={},
-            max_images=4,
+            batch=batch, included_classes={1: "dog"}, max_images=1
         )
-        # ceil(sqrt(4))=2 cols, ceil(4/2)=2 rows → (16, 16)
-        assert result.size == (16, 16)
+        assert result.getpixel((0, 0)) != (0, 0, 0)
```
yutong-xiang-97: This comparison is really weak. This only validates if something is drawn, but you still won't catch if the color is as expected.

I understand that using `utils.get_class_color` would result in a chicken-egg loop so I would suggest the following:

Instead of using (0,0,0) you can use a non-trivial value for the background in `_make_batch`, something like `image = torch.full((batch_size, 3, height, width), fill)`. Then you have a deterministic value to expect if nothing draws. Here you `assert result.getpixel((0, 0)) != expected_value`.

This is still not perfect but cheap enough to buy a lot of extra coverages, e.g. channel-order or mode bug in the images, or drawing accidentally fills the whole canvas instead of just the box.

tests/_visualize/test_object_detection.py
```diff
@@ -70,176 +91,272 @@ def _make_results(
...
+            image=torch.zeros(1, 3, 32, 32), bboxes=bboxes, classes=classes
+        )
         result = object_detection.plot_object_detection_labels(
-            batch=batch,
-            included_classes={1: "dog"},
-            max_images=1,
+            batch=batch, included_classes={}, max_images=1
         )
-        assert result.size == (32, 32)
+        assert result.getpixel((0, 0)) != (0, 0, 0)
```
yutong-xiang-97: Same here.

tests/_visualize/test_object_detection.py
```diff
@@ -70,176 +91,272 @@ def _make_results(
...
-            batch=batch,
-            included_classes={},
-            max_images=1,
-            mean=(0.485, 0.456, 0.406),
-            std=(0.229, 0.224, 0.225),
+            batch=batch, included_classes={0: "cat"}, max_images=2
         )
-        assert result.size == (32, 32)
+        # Grid is 2×1 (64 wide, 32 tall): image 0 at x=0..31, image 1 at x=32..63.
+        assert result.getpixel((0, 0)) != (0, 0, 0)  # box edge drawn on image 0
```
yutong-xiang-97: Same here.

tests/_visualize/test_object_detection.py
```diff
@@ -70,176 +91,272 @@ def _make_results(
...
-            included_classes={},
-            max_images=1,
+            included_classes={0: "cat"},
+            max_images=2,
             score_threshold=0.5,
             max_pred_boxes=10,
         )
-        assert result.size == (32, 32)
+        # Grid is 2×1 (64 wide, 32 tall): image 0 at x=0..31, image 1 at x=32..63.
+        assert result.getpixel((0, 0)) != (0, 0, 0)  # box edge drawn on image 0
```
yutong-xiang-97: Same here

tests/_visualize/test_object_detection.py
```diff
@@ -70,176 +91,272 @@ def _make_results(
...
+                    "scores": torch.tensor([score]),
+                }
+            ],
             included_classes={0: "cat"},
             max_images=1,
             score_threshold=0.5,
             max_pred_boxes=10,
         )
-        assert result.size == (32, 32)
+        assert (result.getpixel((0, 0)) != (0, 0, 0)) == drawn
```
yutong-xiang-97: Same here

tests/_visualize/test_object_detection.py
```diff
@@ -70,176 +91,272 @@ def _make_results(
...
+                    "labels": torch.tensor([42], dtype=torch.long),
+                    "scores": torch.tensor([0.9]),
+                }
+            ],
+            included_classes={},
+            max_images=1,
+            score_threshold=0.5,
+            max_pred_boxes=10,
+        )
+        assert result.getpixel((0, 0)) != (0, 0, 0)
```
yutong-xiang-97: Same here

tests/_visualize/test_object_detection.py
```diff
@@ -70,176 +91,272 @@ def _make_results(
...
+                    "labels": torch.zeros(1, dtype=torch.long),
+                    "scores": torch.tensor([0.9]),
+                }
+            ],
+            included_classes={0: "obj"},
+            max_images=1,
+            score_threshold=0.5,
+            max_pred_boxes=10,
+        )
+        assert result.getpixel((32, 32)) != (0, 0, 0)  # scaled box top-left corner
```
yutong-xiang-97: Same here

tests/_visualize/test_object_detection.py
```diff
@@ -70,176 +91,272 @@ def _make_results(
...
+            ],
+            included_classes={0: "obj"},
             max_images=1,
             score_threshold=0.5,
             max_pred_boxes=10,
-            mean=(0.485, 0.456, 0.406),
-            std=(0.229, 0.224, 0.225),
         )
-        assert result.size == (32, 32)
+        assert result.getpixel((48, 40)) != (0, 0, 0)  # left edge of scaled box
```
yutong-xiang-97: Same here.

tests/_visualize/test_object_detection.py
yutong-xiang-97: From Claude Code

The `boxes_per_image > 0` branch of `_make_results` is never exercised — every call passes `boxes_per_image=0` (default) or builds the result dict inline. Either drop the unused branch or write a test that uses it. As-is, it adds cruft and lies about what the helper is verified to do.

tests/_visualize/test_utils.py
```diff
@@ -11,138 +11,86 @@
...
-        x1, y1 = 10, 100  # plenty of space above
-
-        utils._draw_bbox_label(draw=draw, x1=x1, y1=y1, text="dog", color=color)
-
-        # Label is drawn above y1: pixel just inside the top-left of the label rect is colored.
-        px = image.getpixel((x1 + 1, y1 - 1))
+        utils._draw_bbox_label(draw=draw, x1=10, y1=100, text="dog", color=color)
+        px = image.getpixel((11, 99))
         assert isinstance(px, tuple)
         assert px[:3] == color
```
yutong-xiang-97: You could further enhance it by also asserting the opposite side is still background white.

tests/_visualize/test_utils.py
```diff
@@ -11,138 +11,86 @@
...
-        x1, y1 = 10, 2  # y1 too small to draw above
-
-        utils._draw_bbox_label(draw=draw, x1=x1, y1=y1, text="cat", color=color)
-
-        # Label is drawn below y1: pixel just inside the top-left of the label rect is colored.
-        px = image.getpixel((x1 + 1, y1 + 1))
+        utils._draw_bbox_label(draw=draw, x1=10, y1=2, text="cat", color=color)
+        px = image.getpixel((11, 3))
         assert isinstance(px, tuple)
         assert px[:3] == color
```
yutong-xiang-97: Same here. You could further enhance it by also asserting the opposite side is still background white.

yutong-xiang-97 (commented):
Some comments to improve the test strictness.

yutong-xiang-97 (approved):

liopeer (approved):

PR #710, opened 2026-04-22
Decode RLE encoded COCO segmetation datasets

simonschoelly:
## What has changed and why?

This PR adds support for RLE encoded masks for COCO instance segmentation datasets.

As we use pycocotools to decode these masks and pycocotools only works on Python >= 3.9, we throw an exception if such a mask is encountered on Python 3.8.

## How has it been tested?

Added a test that ensures that COCO datasets with compresses and uncompressed RLE masks work.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_data/instance_segmentation_dataset.py
+++ b/src/lightly_train/_data/instance_segmentation_dataset.py
@@ -123,16 +123,17 @@ class InstanceSegmentationDataset(TaskDataset):
                 mask_list.append(mask)
             elif isinstance(segment, dict):
                 # RLE format. pycocotools is only available for Python >= 3.9.
-                if sys.version_info < (3, 9):
+                if sys.version_info >= (3, 9):
+                    from pycocotools import mask as coco_mask
+
+                    mask_list.append(
+                        coco_mask.decode(segment).astype(np.bool_)  # type: ignore[arg-type]
+                    )
+                else:
                     raise RuntimeError(
                         "RLE encoded segmentation requires Python >= 3.9 "
                         "for pycocotools support."
                     )
-                from pycocotools import mask as coco_mask
-
-                mask_list.append(
-                    coco_mask.decode(segment).astype(np.bool_)  # type: ignore[arg-type]
-                )
         binary_masks_np = (
             np.stack(mask_list) if mask_list else np.zeros((0, h, w), dtype=np.bool_)
         )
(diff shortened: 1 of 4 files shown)


src/lightly_train/_data/instance_segmentation_dataset.py
```diff
@@ -410,58 +432,100 @@ def list_image_info(self) -> Iterable[dict[str, str]]:
...
+
+                            # Ensure RLE is in compressed format.
+                            if isinstance(segmentation.get("counts"), list):
+                                rle = coco_mask.frPyObjects(
+                                    cast("_EncodedRLE", segmentation),
+                                    image_height_pixel,
+                                    image_width_pixel,
+                                )
+                            else:
+                                rle = cast("_EncodedRLE", segmentation)
```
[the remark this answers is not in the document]
simonschoelly: I think that is ok for now, we don;t seem to use strict MyPy

src/lightly_train/_data/instance_segmentation_dataset.py
```diff
@@ -106,15 +112,30 @@ def __getitem__(self, index: int) -> InstanceSegmentationDatasetItem:
...
-            for polygon_group in polygons
-        ]
+
         bboxes_np = np.array(bboxes, dtype=np.float64).reshape(len(bboxes), 4)
         class_labels_np = np.array(class_labels, dtype=np.int64)
 
-        binary_masks_np = yolo_helpers.binary_masks_from_polygons(
-            polygons=polygons_np, height=image_np.shape[0], width=image_np.shape[1]
+        h, w = image_np.shape[0], image_np.shape[1]
+        mask_list: list[np.ndarray[Any, Any]] = []
```
liopeer: Prefer using `numpy.typing.NDarray`, which also allows for the additional dtype annotation.

src/lightly_train/_data/instance_segmentation_dataset.py
```diff
@@ -444,58 +465,98 @@ def list_image_info(self) -> Iterable[dict[str, str]]:
             image_id = image["id"]
             image_filepath = image_dir / image["file_name"]
 
-            polygons = []
+            segments: list[Any] = []
```
liopeer: Nit: Narrower typing would be appreciated, but not super important.

src/lightly_train/_data/instance_segmentation_dataset.py
```diff
@@ -444,58 +465,98 @@ def list_image_info(self) -> Iterable[dict[str, str]]:
...
             bboxes = []
             class_labels = []
             if image_id in annotations_by_image_id:
```
liopeer, line 564: There is now a lot of logic in this loop. I think we could potentially unravel this a bit and move certain functionalities to dedicated module-level functions. Would make testing/reviewing a bit easier.

liopeer: E.g. the normalization, checks on whether it is a valid polygon, generation of bbox if not existing, ...

tests/_data/test_instance_segmentation_dataset.py
```diff
@@ -358,3 +359,142 @@ def test_mmap_hash_changes_when_annotations_modified(self, tmp_path: Path) -> No
...
+                    count += 1
+                else:
+                    counts.append(count)
+                    current = int(val)
+                    count = 1
+            counts.append(count)
+            uncompressed_rle_annotation = {
+                "counts": counts,
+                "size": [height, width],
+            }
```
liopeer: Would move to a helper/fixture to make the test more concise/better readable.

liopeer (commented):
Would recommend to do a couple of small improvements, but it looks correct!

src/lightly_train/_data/instance_segmentation_dataset.py
simonschoelly: Done, I've also added some tests.

tests/_data/test_instance_segmentation_dataset.py
simonschoelly: Done

src/lightly_train/_data/instance_segmentation_dataset.py
```diff
@@ -106,15 +204,30 @@ def __getitem__(self, index: int) -> InstanceSegmentationDatasetItem:
...
+                    polygons_np, height=h, width=w
+                )
+            elif isinstance(segment, dict):
+                # Compressed RLE format.
+                if coco_mask is None:
+                    raise RuntimeError(
+                        "RLE encoded segmentation requires Python >= 3.9 "
+                        "for pycocotools support."
+                    )
+                mask = coco_mask.decode(segment).astype(np.bool_)  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
simonschoelly: I think that is ok for now

liopeer (approved):
LGTM!

PR #711, opened 2026-04-24
Add D-FINE Decoder

yutong-xiang-97:
## What has changed and why?

Add D-FINE decoder based on the [original implementation](https://github.com/Peterande/D-FINE/blob/master/src/zoo/dfine/dfine_decoder.py). Meanwhile, we retain backward compatibility with using the old RTDETRv2 decoder with a configurable `model_args` entry.

Major changes include (based on D-FINE paper):
- replacing original decoder projection layer with a Target Gating Layer
- applying Fine-grained Distribution Refinement (FDR)

The `dfine_decoder.py` and `dfine_utils.py` are mostly carbon copies of the original except for:
- adding ``score_head_reuse_or_reinit_hook`` and (when denoising is enabled) ``denoising_class_embed_reuse_or_reinit_hook`` as `load_state_dict` pre-hooks, so pretrained D-FINE checkpoints can be adapted to a new ``num_classes`` at load time. The same hooks are refactored from the `rtdetrv2_decoder.py` to a separate `hooks.py`.
- registering ``anchors`` / ``valid_mask`` for evaluation as non-persistent buffers instead so they are not written into checkpoints.
- making ``value_op`` return the flat ``[bs, L, n_head, c]`` tensor that helper expects rather than a pre-split per-level tuple to reuse existing `deformable_attention_core_func_v2` in `utils.py` without modification.
- making `bbox_head` and `pre_bbox_head` FP32. See 
  - https://github.com/Peterande/D-FINE/issues/3, 
  - https://github.com/Peterande/D-FINE/issues/43, and 
  - https://github.com/Peterande/D-FINE/issues/199 in D-FINE repo and 
  - #707 in our repo. 
  I also created an issue for potential ONNX export failure.

Note that currently:
- this does not include the necessary changes for the D-FINE losses. Will add in a follow-up PR.
- the default decoder is the original RTDETRv2 decoder. Will switch once we conclude the loss implementation and get proper results from benchmarks.

## How has it been tested?

- N/A (not in any testable state)

- [ ] Yes
- [x] Not needed (to be added after finishing D-FINE loss)

- [ ] Yes

### DIFF
--- a/src/lightly_train/_task_models/dinov2_ltdetr_object_detection/task_model.py
+++ b/src/lightly_train/_task_models/dinov2_ltdetr_object_detection/task_model.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 
 import logging
 from pathlib import Path
-from typing import Any
+from typing import Any, Literal
 
 import torch
 from PIL.Image import Image as PILImage
@@ -27,6 +27,9 @@ from lightly_train._task_models.dinov2_ltdetr_object_detection.dinov2_vit_wrappe
     DINOv2STAs,
 )
 from lightly_train._task_models.object_detection_components import tiling_utils
+from lightly_train._task_models.object_detection_components.dfine_decoder import (
+    DFINETransformer,
+)
 from lightly_train._task_models.object_detection_components.hybrid_encoder import (
     HybridEncoder,
 )
@@ -40,6 +43,38 @@ from lightly_train._task_models.task_model import TaskModel
 from lightly_train.types import PathLike
 
 logger = logging.getLogger(__name__)
+
+_LTDETRDecoder = Literal["rtdetrv2", "dfine"]
+
+
+def _build_decoder(
+    *,
+    config: _DINOv2LTDETRObjectDetectionConfig,
+    decoder: _LTDETRDecoder,
+    num_classes: int,
+    image_size: tuple[int, int],
+    cross_attn_method: str | None = None,
+) -> RTDETRTransformerv2 | DFINETransformer:
+    if decoder == "rtdetrv2":
+        decoder_config = config.rtdetr_transformer.model_dump()
+        if cross_attn_method is not None:
+            decoder_config["cross_attn_method"] = cross_attn_method
+        decoder_config.update({"num_classes": num_classes})
+        return RTDETRTransformerv2(  # type: ignore[no-untyped-call]
+            **decoder_config,
+            eval_spatial_size=image_size,
+        )
+    elif decoder == "dfine":
+        decoder_config = config.dfine_transformer.model_dump()
+        if cross_attn_method is not None:
+            decoder_config["cross_attn_method"] = cross_attn_method
+        decoder_config.update({"num_classes": num_classes})
+        return DFINETransformer(  # type: ignore[no-untyped-call]
+            **decoder_config,
+            eval_spatial_size=image_size,
+        )
+    else:
+        raise ValueError(f"Unsupported LTDETR decoder: {decoder}")
 
 
 class _HybridEncoderConfig(PydanticConfig):
@@ -212,6 +247,50 @@ class _RTDETRTransformerv2ViTGConfig(_RTDETRTransformerv2Config):
     dim_feedforward: int = 12288
 
 
+class _DFINETransformerConfig(PydanticConfig):
+    feat_channels: list[int] = [256, 256, 256]
+    feat_strides: list[int] = [8, 16, 32]
+    hidden_dim: int = 256
+    num_levels: int = 3
+    num_layers: int = 6
+    num_queries: int = 300
+    num_denoising: int = 100
+    label_noise_ratio: float = 0.5
+    box_noise_scale: float = 1.0
+    eval_idx: int = -1
+    num_points: list[int] = [3, 6, 3]
+    query_select_method: str = "default"
+    cross_attn_method: str = "default"
+    dim_feedforward: int = 1024
+    reg_max: int = 32
+    reg_scale: float = 4.0
+    layer_scale: float = 1.0
+
+
+class _DFINETransformerViTSConfig(_DFINETransformerConfig):
+    feat_channels: list[int] = [224, 224, 224]
+    feat_strides: list[int] = [7, 14, 28]
+    hidden_dim: int = 224
+
+
+class _DFINETransformerViTBConfig(_DFINETransformerConfig):
+    feat_channels: list[int] = [768, 768, 768]
+    feat_strides: list[int] = [7, 14, 28]
+    hidden_dim: int = 768
+
+
+class _DFINETransformerViTLConfig(_DFINETransformerConfig):
+    feat_channels: list[int] = [1024, 1024, 1024]
+    feat_strides: list[int] = [7, 14, 28]
+    hidden_dim: int = 1024
+
+
+class _DFINETransformerViTGConfig(_DFINETransformerConfig):
+    feat_channels: list[int] = [1536, 1536, 1536]
+    feat_strides: list[int] = [7, 14, 28]
+    hidden_dim: int = 1536
+
+
 class _BackboneWrapperViTSConfig(PydanticConfig):
     interaction_indexes: list[int] = [5, 8, 11]
     finetune: bool = True
@@ -245,8 +324,10 @@ class _RTDETRPostProcessorConfig(PydanticConfig):
 
 
 class _DINOv2LTDETRObjectDetectionConfig(PydanticConfig):
+    decoder: _LTDETRDecoder = "rtdetrv2"
     hybrid_encoder: _HybridEncoderConfig
     rtdetr_transformer: _RTDETRTransformerv2Config
+    dfine_transformer: _DFINETransformerConfig
     rtdetr_postprocessor: _RTDETRPostProcessorConfig
 
 
@@ -256,6 +337,9 @@ class _DINOv2LTDETRObjectDetectionViTSConfig(_DINOv2LTDETRObjectDetectionConfig)
     )
     rtdetr_transformer: _RTDETRTransformerv2ViTSConfig = Field(
         default_factory=_RTDETRTransformerv2ViTSConfig
+    )
+    dfine_transformer: _DFINETransformerViTSConfig = Field(
+        default_factory=_DFINETransformerViTSConfig
     )
     rtdetr_postprocessor: _RTDETRPostProcessorConfig = Field(
         default_factory=_RTDETRPostProcessorConfig
@@ -272,6 +356,9 @@ class _DINOv2LTDETRObjectDetectionViTBConfig(_DINOv2LTDETRObjectDetectionConfig)
     rtdetr_transformer: _RTDETRTransformerv2ViTBConfig = Field(
         default_factory=_RTDETRTransformerv2ViTBConfig
     )
+    dfine_transformer: _DFINETransformerViTBConfig = Field(
+        default_factory=_DFINETransformerViTBConfig
+    )
     rtdetr_postprocessor: _RTDETRPostProcessorConfig = Field(
         default_factory=_RTDETRPostProcessorConfig
     )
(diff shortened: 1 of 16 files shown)


src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
```diff
@@ -75,6 +75,7 @@ class DINOv2LTDETRObjectDetectionTrainArgs(TrainModelArgs):
     backbone_url: str = ""
     backbone_args: dict[str, Any] = {}
     backbone_freeze: bool = False
+    decoder: Literal["rtdetrv2", "dfine"] = "rtdetrv2"
```
liopeer: Suggestion: Name it `decoder_name` (in accordance with `model_name`).

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/task_model.py
```diff
@@ -41,6 +44,38 @@
 
 logger = logging.getLogger(__name__)
 
+_LTDETRDecoder = Literal["rtdetrv2", "dfine"]
+
+
+def _build_decoder(
```
liopeer: Nit: I prefer to have private functions at the bottom of a module.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/task_model.py
```diff
@@ -212,6 +247,50 @@ class _RTDETRTransformerv2ViTGConfig(_RTDETRTransformerv2Config):
     dim_feedforward: int = 12288
 
 
+class _DFINETransformerConfig(PydanticConfig):
```
liopeer, line 220: Did not check all of these values. Assuming you copied them from the configs of the RTDETR decoder that are just above.

liopeer: Actually, is it deliberate that we have `num_layers=6`?

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/task_model.py
```diff
@@ -41,6 +44,38 @@
...
+    *,
+    config: _DINOv2LTDETRObjectDetectionConfig,
+    decoder: _LTDETRDecoder,
+    num_classes: int,
+    image_size: tuple[int, int],
+    cross_attn_method: str | None = None,
+) -> RTDETRTransformerv2 | DFINETransformer:
+    if decoder == "rtdetrv2":
+        decoder_config = config.rtdetr_transformer.model_dump()
+        if cross_attn_method is not None:
```
liopeer: Is this actually even possible? Should it not throw an error?

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/task_model.py
```diff
@@ -212,6 +247,50 @@ class _RTDETRTransformerv2ViTGConfig(_RTDETRTransformerv2Config):
...
+    num_layers: int = 6
+    num_queries: int = 300
+    num_denoising: int = 100
+    label_noise_ratio: float = 0.5
+    box_noise_scale: float = 1.0
+    eval_idx: int = -1
+    num_points: list[int] = [3, 6, 3]
+    query_select_method: str = "default"
+    cross_attn_method: str = "default"
+    dim_feedforward: int = 1024
```
liopeer: This is very small for the big models, no?

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/task_model.py
```diff
@@ -373,6 +467,7 @@ def __init__(
         config_name = config_name.replace("-noreg", "")
         config_cls = config_mapping[config_name]
         config = config_cls()
+        config.decoder = decoder
```
liopeer: Another reason to use `decoder_name` and not `decoder`. I am surprised that mypy does not afterwards complain, because here it's typed as `config.decoder: str` and in line 486 it will be typed as `config.decoder: RTDETRTransformerv2 | DFINETransformer`.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/task_model.py
```diff
@@ -782,6 +878,7 @@ def __init__(
         config_name = parsed_name["backbone_name"]
         config_cls = config_mapping[config_name]
         config = config_cls()
+        config.decoder = decoder
```
liopeer: Same here ofc.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
```diff
@@ -36,6 +36,9 @@
     DINOv3STAs,
 )
 from lightly_train._task_models.object_detection_components import tiling_utils
+from lightly_train._task_models.object_detection_components.dfine_decoder import (
```
liopeer, line 39: Same points as in the DINOv2 variant.

src/lightly_train/_task_models/object_detection_components/hooks.py
```diff
@@ -0,0 +1,228 @@
+#
```
liopeer, line 1: AFAIK this module is now entirely from us, right? If that's correct we should not have it under Apache 2.0, i.e. we should adjust the the Makefile. I would actually also include another license file for D-FINE (and another header), even though RT-DETRv2 and D-FINE are both Apache 2.0. It just then makes it clear from the Makefile which module was vendored from which other project.

src/lightly_train/_task_models/object_detection_components/hooks.py
```diff
@@ -0,0 +1,228 @@
...
+import logging
+from typing import Any, Literal
+
+import torch
+from torch.nn import Module, ModuleList
+
+logger = logging.getLogger(__name__)
+
+
+def denoising_class_embed_reuse_or_reinit_hook(
```
liopeer, line 31: Assuming all these functions are directly copied.

src/lightly_train/_task_models/object_detection_components/rtdetrv2_decoder.py
```diff
@@ -12,39 +12,40 @@
...
 
-# Modifications Copyright 2025 Lightly AG:
-# - added load state dict pre hooks to reinitialize the classification score heads
-#   and denoising class embedding if the number of classes has changed
-# - implemented `score_head_reuse_or_reinit_hook`, `_score_head_reuse_or_reinit_hook`,
-#   `_reuse_or_reinit`, and `denoising_class_embed_reuse_or_reinit_hook` functions
+# Modifications Copyright 2026 Lightly AG:
+#  - Registered ``score_head_reuse_or_reinit_hook`` and (when denoising is
+#     enabled) ``denoising_class_embed_reuse_or_reinit_hook`` as
+#     load-state-dict pre-hooks, so pretrained D-FINE checkpoints can be
```
liopeer:
```suggestion
#     load-state-dict pre-hooks, so pretrained RT-DETR checkpoints can be
```

src/lightly_train/_task_models/object_detection_components/hooks.py
liopeer: We also have a separate license file for e.g. TIMM, which is also Apache 2.0

src/lightly_train/_task_models/object_detection_components/dfine_utils.py
```diff
@@ -0,0 +1,79 @@
...
+
+import torch
+from torch import Tensor
+
+from lightly_train._task_models.object_detection_components.box_ops import (
+    box_xyxy_to_cxcywh,
+)
+
+
+def weighting_function(
```
liopeer, line 28: Not checking, since I assume it's directly from D-FINE.

liopeer (commented):
Overall looking solid. Left a few comments for smaller improvements!

src/lightly_train/_task_models/object_detection_components/hooks.py
yutong-xiang-97: These hooks are by us entirely, but yes we need to include D-FINE in NOTICE and Makefile.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/task_model.py
yutong-xiang-97: For decoder's `dim_feedforward` D-FINE uses 1024 across size S to X. For hybrid encoder, it does use 2048 for X.

That being said I will use DEIMv2 configs anyway and that DEIMv2 decoder's `dim_feedforward` does scale with the model size.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/task_model.py
yutong-xiang-97: Ah good point! This is different across the model size. With small models this is 3. Will fix.

yutong-xiang-97: Actually we may need to use DEIMv2's default here since D-FINE only uses HGNetv2 as backbone. DEIMv2's decoder also uses D-FINE's decoder and losses.

liopeer: Now we have `num_layers=4` everywhere except for the super class. I am a bit confused.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/task_model.py
```diff
@@ -212,6 +217,58 @@ class _RTDETRTransformerv2ViTGConfig(_RTDETRTransformerv2Config):
...
+    num_layers: int = 6
+    num_queries: int = 300
+    num_denoising: int = 100
+    label_noise_ratio: float = 0.5
+    box_noise_scale: float = 1.0
+    eval_idx: int = -1
+    num_points: list[int] = [3, 6, 3]
+    query_select_method: str = "default"
+    cross_attn_method: str = "default"
+    dim_feedforward: int = 2048
```
liopeer: Would remove the default if it's something else in all the models. Easy source of errors.

liopeer: Same applies to `num_layers` (if different across model sizes).

src/lightly_train/_task_models/object_detection_components/hooks.py
liopeer: Seems to still be Apache 2.0 licensed.

liopeer (commented):
Overall LGTM, but I am not sure if every comment was addressed that I previously raised. Would be great to quickly get answers to this before full approval.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/task_model.py
yutong-xiang-97: Could you point me to where the error could occur? The training runs without errors and `_RTDETRTransformerv2Config` is fully serializable.

src/lightly_train/_task_models/object_detection_components/dfine_decoder.py
```diff
@@ -0,0 +1,1063 @@
...
+    ):
+        """Multi-Scale Deformable Attention"""
+        super().__init__()
+        self.embed_dim = embed_dim
+        self.num_heads = num_heads
+        self.num_levels = num_levels
+        self.offset_scale = offset_scale
+
+        if isinstance(num_points, list):
+            assert len(num_points) == num_levels, ""
```
simonschoelly, line 101: What is the purpose of emitting "" here if the assert fails?

yutong-xiang-97: This is from the original d-fine code, I would just leave it as-is.

liopeer (approved):
LGTM! But leave @simonschoelly to resolve his comment on his side.

PR #712, opened 2026-04-28
Add d-fine losses

yutong-xiang-97:
## What has changed and why?

Add D-FINE losses based on the [original implementation](https://github.com/Peterande/D-FINE/blob/master/src/zoo/dfine/dfine_criterion.py). Meanwhile, we retain backward compatibility with using the old RTDETRv2 losses if the decoder is configured to "rtdetrv2".

Major changes include (based on D-FINE paper):
- Applying Fine-grained Localization (FGL) Loss
- Applying Decoupled Distillation Focal (DDF) Loss

The `dfine_criterion.py`, `dfine_utils.py`, and `dist_utils.py` are mostly carbon copies of the original except for:
 - Removing upstream registry hooks and adapted imports for LightlyTrain.
 - Adding explicit ``world_size`` forwarding for LightlyTrain's Fabric training loop.
 - Supporting validation-time loss computation when auxiliary outputs are absent.
 - Adding FP32 guards around bbox refinement for stability in mixed precision.

Note that currently the default decoder is the original RTDETRv2 decoder. Will switch once we conclude the loss implementation and get proper results from benchmarks

## How has it been tested?

- Manually

- [x] Yes

- [ ] Yes
- [x] Not needed (in a separate issue)

### DIFF
--- a/src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
+++ b/src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
@@ -88,31 +88,6 @@ _DFINE_EXTRA_LOSSES: list[str] = ["local"]
 _DFINE_LOSS_NAMES: list[str] = [*_RTDETRV2_LOSS_NAMES, *_DFINE_EXTRA_LOSS_WEIGHT_DICT]
 
 
-def _get_loss_log_dict(
-    *,
-    total_loss: Tensor,
-    loss_dict: dict[str, Tensor],
-    loss_names: list[str],
-) -> dict[str, Tensor]:
-    zero = total_loss.detach() * 0
-    log_dict = {"loss": total_loss.detach()}
-    for loss_name in loss_names:
-        if loss_name == "loss":
-            continue
-        if loss_name == "loss_ddf":
-            log_dict[loss_name] = sum(
-                (
-                    v.detach()
-                    for k, v in loss_dict.items()
-                    if k == "loss_ddf" or k.startswith("loss_ddf_")
-                ),
-                start=zero,
-            )
-        else:
-            log_dict[loss_name] = loss_dict.get(loss_name, zero).detach()
-    return log_dict
-
-
 class DINOv2LTDETRObjectDetectionTrainArgs(TrainModelArgs):
     default_batch_size: ClassVar[int] = 16
     default_steps: ClassVar[int] = (
@@ -238,13 +213,14 @@ class DINOv2LTDETRObjectDetectionTrain(TrainModel):
         loss_weight_dict = model_args.loss_weight_dict
         losses = model_args.losses
         criterion: DFINECriterion | RTDETRCriterionv2
-        if model_args.decoder == "dfine":
+        if model_args.decoder_name == "dfine":
             loss_weight_dict = {**_DFINE_EXTRA_LOSS_WEIGHT_DICT, **loss_weight_dict}
             losses = [
                 *losses,
                 *(name for name in _DFINE_EXTRA_LOSSES if name not in losses),
             ]
-            self.loss_names = _DFINE_LOSS_NAMES
+            self.train_loss_names = _DFINE_LOSS_NAMES
+            self.val_loss_names = _RTDETRV2_LOSS_NAMES
             if not isinstance(self.model.decoder, DFINETransformer):
                 raise TypeError("decoder='dfine' requires a DFINETransformer decoder.")
             criterion = DFINECriterion(  # type: ignore[no-untyped-call]
@@ -257,7 +233,8 @@ class DINOv2LTDETRObjectDetectionTrain(TrainModel):
                 reg_max=self.model.decoder.reg_max,
             )
         else:
-            self.loss_names = _RTDETRV2_LOSS_NAMES
+            self.train_loss_names = _RTDETRV2_LOSS_NAMES
+            self.val_loss_names = _RTDETRV2_LOSS_NAMES
             criterion = RTDETRCriterionv2(  # type: ignore[no-untyped-call]
                 matcher=matcher,
                 weight_dict=loss_weight_dict,
@@ -275,7 +252,7 @@ class DINOv2LTDETRObjectDetectionTrain(TrainModel):
             split="train",
             class_names=class_names,
             box_format="xyxy",
-            loss_names=self.loss_names,
+            loss_names=self.train_loss_names,
             train_loss_running_mean_window=gradient_accumulation_steps,
         )
         self.val_metrics = ObjectDetectionTaskMetric(
@@ -283,7 +260,7 @@ class DINOv2LTDETRObjectDetectionTrain(TrainModel):
             split="val",
             class_names=class_names,
             box_format="xyxy",
-            loss_names=self.loss_names,
+            loss_names=self.val_loss_names,
         )
 
         # TODO(Nauryz, 04/2026): These visualization thresholds are currently hardcoded, but we may want to make them configurable in the future (with logger_args).
@@ -367,7 +344,7 @@ class DINOv2LTDETRObjectDetectionTrain(TrainModel):
             loss_dict=_get_loss_log_dict(
                 total_loss=total_loss,
                 loss_dict=loss_dict,
-                loss_names=self.loss_names,
+                loss_names=self.train_loss_names,
             ),
             weight=samples.shape[0],
         )
@@ -474,7 +451,7 @@ class DINOv2LTDETRObjectDetectionTrain(TrainModel):
             loss_dict=_get_loss_log_dict(
                 total_loss=total_loss,
                 loss_dict=loss_dict,
-                loss_names=self.loss_names,
+                loss_names=self.val_loss_names,
             ),
             weight=samples.shape[0],
         )
@@ -610,3 +587,31 @@ class DINOv2LTDETRObjectDetectionTrain(TrainModel):
                 max_norm=self.model_args.gradient_clip_val,
                 error_if_nonfinite=False,
             )
+
+
+def _get_loss_log_dict(
+    *,
+    total_loss: Tensor,
+    loss_dict: dict[str, Tensor],
+    loss_names: list[str],
+) -> dict[str, Tensor]:
+    zero = total_loss.new_zeros(())
+    log_dict = {"loss": total_loss.detach()}
+    for loss_name in loss_names:
+        if loss_name == "loss":
+            continue
+        if loss_name == "loss_ddf":
+            loss_values = [
+                v.detach()
+                for k, v in loss_dict.items()
+                if k == "loss_ddf" or k.startswith("loss_ddf_")
+            ]
+            if not loss_values:
+                raise KeyError(
+                    "No loss entries found for 'loss_ddf'. Available losses: "
+                    f"{sorted(loss_dict.keys())}"
+                )
+            log_dict[loss_name] = sum(loss_values, start=zero)
+        else:
+            log_dict[loss_name] = loss_dict[loss_name].detach()
+    return log_dict
--- a/src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
+++ b/src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
@@ -45,6 +45,12 @@ from lightly_train._task_models.dinov2_ltdetr_object_detection.transforms import
     DINOv2LTDETRObjectDetectionValTransform,
     DINOv2LTDETRObjectDetectionValTransformArgs,
 )
+from lightly_train._task_models.object_detection_components.dfine_criterion import (
+    DFINECriterion,
+)
+from lightly_train._task_models.object_detection_components.dfine_decoder import (
+    DFINETransformer,
+)
 from lightly_train._task_models.object_detection_components.ema import ModelEMA
 from lightly_train._task_models.object_detection_components.matcher import (
     HungarianMatcher,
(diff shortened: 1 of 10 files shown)


src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
```diff
@@ -64,6 +70,43 @@
...
+}
+_RTDETRV2_LOSSES: list[str] = ["vfl", "boxes"]
+_RTDETRV2_LOSS_NAMES: list[str] = ["loss", *_RTDETRV2_LOSS_WEIGHT_DICT]
+
+_DFINE_EXTRA_LOSS_WEIGHT_DICT: dict[str, float] = {"loss_fgl": 0.15, "loss_ddf": 1.5}
+_DFINE_EXTRA_LOSSES: list[str] = ["local"]
+_DFINE_LOSS_NAMES: list[str] = [*_RTDETRV2_LOSS_NAMES, *_DFINE_EXTRA_LOSS_WEIGHT_DICT]
+
+
+def _get_loss_log_dict(
```
liopeer: Would move to bottom of module.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
```diff
@@ -64,6 +70,43 @@
...
+            log_dict[loss_name] = sum(
+                (
+                    v.detach()
+                    for k, v in loss_dict.items()
+                    if k == "loss_ddf" or k.startswith("loss_ddf_")
+                ),
+                start=zero,
+            )
+        else:
+            log_dict[loss_name] = loss_dict.get(loss_name, zero).detach()
```
liopeer: Is the `zero` fallback desired? Or should we at least raise a warning? Also, why are we generating `zero` from the detached loss? Since it's detached we can just use a zero scalar tensor, no?

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
```diff
@@ -184,17 +227,40 @@ def __init__(
...
+        loss_weight_dict = model_args.loss_weight_dict
+        losses = model_args.losses
+        criterion: DFINECriterion | RTDETRCriterionv2
+        if model_args.decoder == "dfine":
+            loss_weight_dict = {**_DFINE_EXTRA_LOSS_WEIGHT_DICT, **loss_weight_dict}
+            losses = [
+                *losses,
+                *(name for name in _DFINE_EXTRA_LOSSES if name not in losses),
+            ]
+            self.loss_names = _DFINE_LOSS_NAMES
```
liopeer: Can't we do that resolution already in the pydantic class? All those 3  things: `loss_weight_dict`, `losses` and `model_args.decoder` are part of the config.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/train_model.py
```diff
@@ -183,17 +226,40 @@ def __init__(
...
-            weight_dict=model_args.loss_weight_dict,
-            losses=model_args.losses,
-            alpha=model_args.loss_alpha,
-            gamma=model_args.loss_gamma,
-            num_classes=len(data_args.included_classes),
-        )
+        loss_weight_dict = model_args.loss_weight_dict
+        losses = model_args.losses
+        criterion: DFINECriterion | RTDETRCriterionv2
+        if model_args.decoder == "dfine":
```
liopeer: Same comment.

src/lightly_train/_task_models/object_detection_components/dfine_criterion.py
```diff
@@ -0,0 +1,647 @@
...
+# limitations under the License.#
+"""
+D-FINE: Redefine Regression Task of DETRs as Fine-grained Distribution Refinement
+Copyright (c) 2024 The D-FINE Authors. All Rights Reserved.
+---------------------------------------------------------------------------------
+Modified from RT-DETR (https://github.com/lyuwenyu/RT-DETR)
+Copyright (c) 2023 lyuwenyu. All Rights Reserved.
+"""
+
+# Modifications Copyright 2026 Lightly AG:
```
liopeer, line 21: Did not look at this.

src/lightly_train/_task_models/object_detection_components/dfine_utils.py
```diff
@@ -12,6 +12,9 @@
 # limitations under the License.#
 """Copyright (c) 2024 The D-FINE Authors. All Rights Reserved."""
 
+# Modifications Copyright 2026 Lightly AG:
+#  - Added type annotations.
```
liopeer, line 16: Seems we added more (?)

src/lightly_train/_task_models/object_detection_components/dist_utils.py
```diff
@@ -0,0 +1,29 @@
...
+- https://github.com/pytorch/vision/blob/main/references/detection/utils.py
+- https://github.com/facebookresearch/detr/blob/master/util/misc.py#L406
+
+Copyright(c) 2023 lyuwenyu. All Rights Reserved.
+"""
+
+import torch
+
+
+def is_dist_available_and_initialized():
```
liopeer, line 24: If at all possible I would somehow try to keep RT-DETR vendored code and D-FINE vendored code separate. Now it seems we're starting to mix stuff by turning the RT-DETR code slowly into D-FINE code. I could imagine a structure that looks like this (but might be overkill):
```shell
object_detection_components/
├── rtdetrv2_components/          # RT-DETRv2 Apache 2.0 license
├── dfine_components/             # D-FINE Apache 2.0 license
└── helpers/                      # owned by Lightly; or just a single helpers.py instead
```
However, I would generally try to change as little as possible to vendored code.

src/lightly_train/_task_models/object_detection_components/rtdetrv2_criterion.py
```diff
@@ -12,6 +12,9 @@
 # limitations under the License.#
 """Copyright(c) 2023 lyuwenyu. All Rights Reserved."""
 
+# Modifications Copyright 2026 Lightly AG:
+# - Introduce sanitize_boxes_cxcywh_normalized function to ensure boxes are in the expected format
```
liopeer, line 16: Is this a remainder from an old PR?

src/lightly_train/_task_models/object_detection_components/dfine_decoder.py
```diff
@@ -986,7 +986,7 @@ def forward(self, feats, targets=None):
         )
 
         # decoder
-        out_bboxes, out_logits, _out_corners, _out_refs, _pre_bboxes, _pre_logits = (
+        out_bboxes, out_logits, out_corners, out_refs, pre_bboxes, pre_logits = (
```
liopeer, line 991: These changes should probably also be documented in the header, no?

src/lightly_train/_task_models/object_detection_components/dfine_decoder.py
```diff
@@ -1053,9 +1082,29 @@ def _set_aux_loss(self, outputs_class, outputs_coord):
...
         ]
 
-
-# TODO (Yutong, 04/26):
-# Replace the ``_set_aux_loss`` call for ``aux_outputs`` / ``dn_outputs``
-#     with the ``_set_aux_loss2`` variant that forwards ``out_corners``,
-#     ``out_refs``, and the last-layer ``teacher_corners``/``teacher_logits``
-#     used for self-distillation (DDF).
+    @torch.jit.unused
+    def _set_aux_loss2(
```
liopeer, line 1088: I need to look at this again tomorrow.

liopeer (commented):
Looking mostly good. Will give it another (more detailed) round of review, but wanted to submit this today.

src/lightly_train/_task_models/object_detection_components/dfine_utils.py
yutong-xiang-97: If you mean functions no. The others are simply in `utils.py` See https://github.com/Peterande/D-FINE/blob/master/src/zoo/dfine/dfine_utils.py

src/lightly_train/_task_models/object_detection_components/dist_utils.py
yutong-xiang-97: RT-DETR and D-FINE indeed share many components (e.g. matcher). In these cases I use RT-DETR with minimum changes. I don't want to have the shared part duplicated in two subdirs. However, I could imagine putting the shared part under `object_detection_components` directly, but this should in any case go to a standalone refactoring PR. I will keep a note for this.

src/lightly_train/_task_models/object_detection_components/rtdetrv2_criterion.py
yutong-xiang-97: Yes #707.

src/lightly_train/_task_models/object_detection_components/dfine_decoder.py
yutong-xiang-97: No this was only made private in the previous PR because of the absence of the losses. See https://github.com/Peterande/D-FINE/blob/d6694750683b0c7e9f523ba6953d16f112a376ae/src/zoo/dfine/dfine_decoder.py#L865

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
yutong-xiang-97: D-FINE does not log validation losses for the two new losses. But after taking a closer look, I think it is better that we use

```
            self.train_loss_names = _DFINE_LOSS_NAMES
            self.val_loss_names = _RTDETRV2_LOSS_NAMES
```

instead of a single `self.loss_names`.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
```diff
@@ -69,6 +75,18 @@
...
 
+_RTDETRV2_LOSS_WEIGHT_DICT: dict[str, float] = {
+    "loss_vfl": 1.0,
+    "loss_bbox": 5.0,
+    "loss_giou": 2.0,
+}
+_RTDETRV2_LOSSES: list[str] = ["vfl", "boxes"]
+_RTDETRV2_LOSS_NAMES: list[str] = ["loss", *_RTDETRV2_LOSS_WEIGHT_DICT]
+
+_DFINE_EXTRA_LOSS_WEIGHT_DICT: dict[str, float] = {"loss_fgl": 0.15, "loss_ddf": 1.5}
```
liopeer, line 87: Using the logic with "extras" that you have to merge seems a bit overcomplicated compared to just listing all loss names/weights twice for the different models, but it's ofc okay this way.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
```diff
@@ -192,32 +210,57 @@ def __init__(
...
+        losses = model_args.losses
+        criterion: DFINECriterion | RTDETRCriterionv2
+        if model_args.decoder_name == "dfine":
+            loss_weight_dict = {**_DFINE_EXTRA_LOSS_WEIGHT_DICT, **loss_weight_dict}
+            losses = [
+                *losses,
+                *(name for name in _DFINE_EXTRA_LOSSES if name not in losses),
+            ]
+            self.train_loss_names = _DFINE_LOSS_NAMES
+            self.val_loss_names = _RTDETRV2_LOSS_NAMES
```
liopeer, line 236: Is this an error or are we really only evaluation on the RT-DETR losses?

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
```diff
@@ -546,3 +587,31 @@ def clip_gradients(self, fabric: Fabric, optimizer: Optimizer) -> None:
                 max_norm=self.model_args.gradient_clip_val,
                 error_if_nonfinite=False,
             )
+
+
+def _get_loss_log_dict(
```
liopeer, line 624: Cover with test.

src/lightly_train/_task_models/object_detection_components/dist_utils.py
liopeer: Yes, please create an issue for this.

src/lightly_train/_task_models/object_detection_components/dfine_utils.py
liopeer: Ok. Did not realize that initially we did not copy the whole module. Again: For vendored code I would try to alter it as little as possible, it's otherwise hard to keep track of the changes, i.e. we should imo always copy the entire module, even if we don't use parts of it.

src/lightly_train/_task_models/object_detection_components/dfine_decoder.py
liopeer: Ok. Looks good.

liopeer (commented):
Only thing to have answered is the comment about `self.val_loss_names = _RTDETRV2_LOSS_NAMES`, the rest does not necessarily need to be addressed.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
yutong-xiang-97: Yes the new losses in the original d-fine paper are train-only. These losses are probabilistic (KL-div) so do not have make much sense in val.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
liopeer: Potentially.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
yutong-xiang-97: I would just keep it as is.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
yutong-xiang-97: I would only add tests after refactoring. Currently the function is replicated for DINOv2 and v3.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
yutong-xiang-97: Found a nice way to do it. Will post a fix.

src/lightly_train/_task_models/dinov2_ltdetr_object_detection/train_model.py
```diff
@@ -127,6 +145,23 @@ class DINOv2LTDETRObjectDetectionTrainArgs(TrainModelArgs):
         validation_alias=AliasChoices("lr_warmup_steps", "scheduler_warmup_steps"),
     )
 
+    @computed_field  # type: ignore[prop-decorator]
```
liopeer, line 151: Nice, did not know this!

liopeer (approved):
LGTM!

PR #713, opened 2026-04-29
Fix default font loading to handle TypeError gracefully

mrpositron:
## What has changed and why?

- `ImageFont.load_default(size=20)` raises `TypeError` on Pillow < 10.1.0 because
  the `size` keyword argument was not available in older versions
- Wrapped the call in a `try/except TypeError` so older Pillow installations fall
  back to `ImageFont.load_default()` (the built-in bitmap font at default size)

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_visualize/utils.py
+++ b/src/lightly_train/_visualize/utils.py
@@ -16,7 +16,10 @@ from PIL.Image import Image as PILImage
 from PIL.ImageDraw import ImageDraw as PILDraw
 from torch import Tensor
 
-_DEFAULT_FONT = ImageFont.load_default(size=20)
+try:
+    _DEFAULT_FONT = ImageFont.load_default(size=20)
+except TypeError:
+    _DEFAULT_FONT = ImageFont.load_default()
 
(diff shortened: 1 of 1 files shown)


src/lightly_train/_visualize/utils.py
```diff
@@ -16,7 +16,10 @@
 from PIL.ImageDraw import ImageDraw as PILDraw
 from torch import Tensor
 
-_DEFAULT_FONT = ImageFont.load_default(size=20)
+try:
+    _DEFAULT_FONT = ImageFont.load_default(size=20)
+except TypeError:
```
liopeer, line 21: If we know the exact version that we need to check/target we usually use `RequirementCache` in a conditional statement, see e.g. here:
https://github.com/lightly-ai/lightly-train/blob/232ab89412c07928f23b277fc35a2a2ff365e9de/src/lightly_train/_transforms/normalize.py#L19

But it's also fine like this here.

liopeer (approved):

PR #714, opened 2026-04-30
Add image logging for classification training

mrpositron:
## What has changed and why?

Adds image logging for the image classification task, mirroring the existing object detection image logging.
`training_step` now logs a label grid for the first 3 steps (rank 0 only), showing each image annotated with its ground-truth class label(s).

`validation_step` logs two grids for the first 3 steps: a ground-truth label grid and a top-k prediction grid (top 3 by default), with predicted class names and confidence scores.

Supporting infrastructure added:
- `_visualize/image_classification.py` - two new public functions: plot_image_classification_labels and plot_image_classification_predictions. Both denormalize images when mean/std are provided, annotate with colored corner labels, and return a PIL image grid.
- `_visualize/utils.py` - new _draw_corner_label helper that draws a semi-transparent black background with colored text in the top-left corner of an image, stackable vertically for multi-label cases.

Example of a groud truth label grid:

image

## How has it been tested?

23 new unit tests in `tests/_visualize/test_image_classification.py` covering:

- Grid is capped at max_images
- Corner label overlay is drawn (pixel darkened at top-left)
- Unknown class IDs fall back to "Class {id}" label
- Empty class list produces no overlay
- Denormalization with parametrized mean/std values (zero image, non-zero, per-channel, clamping at 0 and 1)
- No denormalization when mean/std are omitted
- Multiple labels stack vertically
- Mixed annotated/unannotated images in the same batch
- `effective_k` expands to match the number of ground-truth labels in multi-label classification

- [ ] Yes

- [ ] Yes


mrpositron:
@codex review

mrpositron:
## Changes
- Use `matplotlib` legends to display top-k predictions for improved visualization
- Add `matplotlib` to `pyproject.toml`

Image logging now outputs classification results by printing the top-k predictions:

image

src/lightly_train/_visualize/object_detection.py
```diff
@@ -51,7 +51,7 @@ def plot_object_detection_labels(
 
     pil_images: list[PILImage] = []
     for i in range(n):
-        image_tensor = gt_images[i].clone()
+        image_tensor = gt_images[i].clone().to(dtype=torch.float32)
```
liopeer, line 54: Does Fabric already turn them in to fp16/bf16 if that strategy is set? Or why is this needed?

src/lightly_train/_task_models/image_classification/train_model.py
```diff
@@ -120,17 +125,27 @@ def __init__(
...
+
         image_size = no_auto(val_transform_args.image_size)
         normalize = no_auto(val_transform_args.normalize)
+        normalize_dict: dict[str, Any] | None
+        self._normalize = normalize
+
+        if normalize is None:
+            normalize_dict = None
+        else:
+            normalize_dict = normalize.model_dump()
```
liopeer: This should not actually be necessary. We already have the normalization values in the `TaskModel` (I don't know if this is guaranteed by an interface, but we always need it for the `predict` method). I would do the following:
1. Ensure that `self.image_normalize` is available in the TaskModel. You can do this by e.g. creating an abstract property if it does not exist yet.
2. Access it in the `training_step`/`validation_step` through `self.model.image_normalize`.

Also, I don't think we allow/should allow the normalization to be `None`. I see that this is the case for PicoDet, but that was not deliberately introduced afaik (all the others don't allow it). This saves a bit of the logic that you're creating here.

src/lightly_train/_task_models/image_classification/train_model.py
```diff
@@ -163,6 +178,15 @@ def __init__(
             train_loss_running_mean_window=gradient_accumulation_steps,
         )
 
+        self._internal_class_names: dict[int, str] = {
```
liopeer: In the task model we already have the `self.classes` attribute (I think we have it in every TaskModel subclass). Could we move that to an abstract property of the `TaskModel` and then re-use it from there?

tests/_visualize/test_image_classification.py
```diff
@@ -0,0 +1,415 @@
...
+    if classes is None:
+        classes = [torch.zeros(0, dtype=torch.long) for _ in range(batch_size)]
+    return ImageClassificationBatch(
+        image_path=[f"img_{i}.jpg" for i in range(batch_size)],
+        image=image,
+        classes=classes,
+    )
+
+
+class TestPlotImageClassificationLabels:
```
liopeer: We only group tests in classes if we test methods of said class: https://www.notion.so/getlightly/Python-testing-78e2dc49baf340178fa44ce20031c52e?source=copy_link

tests/_visualize/test_image_classification.py
```diff
@@ -0,0 +1,415 @@
...
+    def test_plot_image_classification_labels_grid_caps_at_max_images(self) -> None:
```
liopeer: We usually use the naming convention like this: `test_<name-of-function-including-potential-underscores>__the_test_case`, i.e. with a double underscore before the test description. You can find a nice example here: https://github.com/lightly-ai/lightly-train/blob/main/tests/_loggers/test_logger_helpers.py
In internal docs it's currently different, but that's the convention we've been using at LightlyTrain so far.

liopeer: We usually use the naming convention like this: `test_<name-of-function-including-potential-underscores>__the_test_case`, i.e. with a double underscore before the test description. You can find a nice example here: https://github.com/lightly-ai/lightly-train/blob/main/tests/_loggers/test_logger_helpers.py
In internal docs it's currently different, but that's the convention we've been using at LightlyTrain so far.

tests/_visualize/test_image_classification.py
```diff
@@ -0,0 +1,415 @@
...
+            # Per-channel std with zero mean: 0.5 * (0.2, 0.4, 0.6) = (0.1, 0.2, 0.3)
+            # -> (25, 51, 76).
+            (0.5, (0.0, 0.0, 0.0), (0.2, 0.4, 0.6), (25, 51, 76)),
+            # Values > 1 are clamped to 1 -> 255.
+            (5.0, (0.0, 0.0, 0.0), (1.0, 1.0, 1.0), (255, 255, 255)),
+            # Values < 0 are clamped to 0.
+            (-5.0, (0.0, 0.0, 0.0), (1.0, 1.0, 1.0), (0, 0, 0)),
+        ],
+    )
+    def test_plot_image_classification_labels_mean_std_denormalizes_image(
```
liopeer: This should imo not be tested like this. These kinds of broad parametrized unit tests should appear at the lowest level of functions, i.e. you can use this to test `_denormalize_image` image function. But for the `plot_classification_labels` function you should only test: 
1. Does it normalize when you specify mean/std?
2. Does it *not* normalize when you pass `None`?

liopeer (changes requested):
Overall looking good! However, I would highly recommend using *fewer* tests, but more targeted. 

The lowest levels of functions you can test with all possible combinations of inputs and varying values etc. But as soon as functions are composed the testing should be targeted and only cover the things that are not yet covered on the lower level.

Also, tests are an important unit of documentation (i.e. they document what a function/method is supposed to do; in addition to potential docstrings). So we should try to keep them relevant, so that people actually look at them.

src/lightly_train/_visualize/image_classification.py
```diff
@@ -0,0 +1,141 @@
...
+    _render_grid,
+)
+from lightly_train.types import ImageClassificationBatch
+
+
+def plot_image_classification_labels(
+    batch: ImageClassificationBatch,
+    included_classes: dict[int, str],
+    max_images: int,
+    image_normalize: dict[str, tuple[float, ...]] | None = None,
```
liopeer: Nit: Avoid defaults in functions that are meant for internal use.

liopeer (approved):
✨ CLEAN ✨ 

One small comment (optional to address it), otherwise it's now very nice!

PR #717, opened 2026-05-05
Fix DINOv3 patch size precedence

gabrielfruet:
## What has changed and why?

 `DINOv3Package.get_model()` now respects an explicit `model_args["patch_size"]` over the patch size encoded in the model name.

 This fixes the case where passing `patch_size` to DINOv3 models was ineffective, which impacted consumers like DINOv3 EoMT semantic segmentation. When an
override happens, we now log a warning so the behavior is explicit.

 ## How has it been tested?

 - `pytest tests/_models/dinov3/test_dinov3_package.py -k patch_size -q`

 This covers:
 - explicit `patch_size` taking precedence
 - warning logging when the model-name patch size is overwritten

 - [ ] Yes

 - [x] Yes
 - [] Not needed (internal change without effects for user)

### DIFF
--- a/docs/source/settings/train_settings.md
+++ b/docs/source/settings/train_settings.md
@@ -102,12 +102,11 @@ restored correctly.
 Dictionary with model-specific training parameters. The available keys vary by
 architecture. The table lists the most commonly tuned options:
 
-| Key                                             | Type                      | Description                                                                   |
-| ----------------------------------------------- | ------------------------- | ----------------------------------------------------------------------------- |
-| [`lr`](#lr)                                     | `float`                   | Base learning rate.                                                           |
-| [`backbone_weights`](#backbone_weights)         | `Path`<br>`str`<br>`None` | Path to backbone weights to load.                                             |
-| [`metric_log_classwise`](#metric_log_classwise) | `bool`                    | Whether to log class-wise metrics.                                            |
-| [`patch_size`](#patch_size)                     | `int`                     | DINOv3 ViT default from model name; overridden by `model_args["patch_size"]`. |
+| Key                                             | Type                      | Description                        |
+| ----------------------------------------------- | ------------------------- | ---------------------------------- |
+| [`lr`](#lr)                                     | `float`                   | Base learning rate.                |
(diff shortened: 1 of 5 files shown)


docs/source/settings/train_settings.md
```diff
@@ -102,11 +102,12 @@ restored correctly.
...
-| [`metric_log_classwise`](#metric_log_classwise) | `bool`                    | Whether to log class-wise metrics. |
+| Key                                             | Type                      | Description                                                                   |
+| ----------------------------------------------- | ------------------------- | ----------------------------------------------------------------------------- |
+| [`lr`](#lr)                                     | `float`                   | Base learning rate.                                                           |
+| [`backbone_weights`](#backbone_weights)         | `Path`<br>`str`<br>`None` | Path to backbone weights to load.                                             |
+| [`metric_log_classwise`](#metric_log_classwise) | `bool`                    | Whether to log class-wise metrics.                                            |
+| [`patch_size`](#patch_size)                     | `int`                     | DINOv3 ViT default from model name; overridden by `model_args["patch_size"]`. |
```
liopeer: This only really applies to ViTs, so I would not list it here. I created an issue to properly document the `model_args` for all fine-tuning tasks (with an automatic dump of the defaults, similar to how we do it for the `method_args` in `docs/prebuild.py`.

docs/source/settings/train_settings.md
```diff
@@ -195,6 +196,12 @@ lightly_train.train_object_detection(
 )
 ```
 
+#### `patch_size`
```
liopeer: Means that I would also not list it here.

src/lightly_train/_models/dinov3/dinov3_package.py
liopeer: While we're at it, can we change the docstring here: Can also return ConvNeXt models.

liopeer (changes requested):
Overall very nice work! A couple of small things that I would address.

liopeer (commented):
LGTM!

liopeer (approved):

PR #720, opened 2026-05-07
Feat support variable patch size for DINOv3 LT-DETR

gabrielfruet:
## What has changed and why?

This PR adds configurable `patch_size` support for DINOv3 LT-DETR.

- `patch_size` now resolves from:
  1. an explicit `model_args.patch_size`
  2. `model_init_args["patch_size"]`
  3. the DINOv3 model name when `patch_size="auto"`
- The resolved value is forwarded to the backbone, including the ConvNeXt wrapper, so LT-DETR can work with variable patch sizes consistently.

## How has it been tested?

- Added/updated tests in `tests/_task_models/dinov3_ltdetr_object_detection/test_task_model.py` covering:
  - auto resolution from model name
  - resolution from `model_init_args`
  - explicit `patch_size`
  - ConvNeXt `auto` passthrough
  - explicit ConvNeXt override

- [x] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
+++ b/src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
@@ -220,7 +220,8 @@ class _RTDETRTransformerv2Config(PydanticConfig):
     eval_idx: int = -1
     num_points: list[int] = [4, 4, 4]
 
-    def resolve_auto(self, patch_size: int) -> None:
+    def resolve_auto(self, patch_size: int | None) -> None:
+        patch_size = patch_size or 16
         if self.feat_strides == "auto":
             self.feat_strides = [
                 int(patch_size * (2 ** (i - 1))) for i in range(self.num_levels)
@@ -404,7 +405,7 @@ class _DINOv3LTDETRObjectDetectionConfig(PydanticConfig):
     dfine_transformer: _DFINETransformerConfig
     rtdetr_postprocessor: _RTDETRPostProcessorConfig
 
-    def resolve_auto(self, patch_size: int) -> None:
+    def resolve_auto(self, patch_size: int | None) -> None:
         self.rtdetr_transformer.resolve_auto(patch_size=patch_size)
 
 
--- a/src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
+++ b/src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
@@ -449,12 +449,38 @@ class DINOv3LTDETRObjectDetection(TaskModel):
         model_name: str,
         classes: dict[int, str],
         image_size: tuple[int, int],
+        patch_size: int,
         image_normalize: dict[str, Any] | None = None,
         backbone_freeze: bool = False,
         backbone_weights: PathLike | None = None,
         backbone_args: dict[str, Any] | None = None,
         load_weights: bool = True,
     ) -> None:
+        """Create a DINOv3 LTDETR object detection model.
+
+        Args:
+            model_name:
+                The model name. For example ``"vitt16-ltdetr"``.
+            classes:
+                A dict mapping class IDs to class names.
+            image_size:
+                The input image size.
+            patch_size:
+                Patch size used to initialize the DINOv3 backbone. This is stored in
+                ``init_args`` so exported checkpoints can be reconstructed with the
+                same backbone patch size.
+            image_normalize:
+                A dict containing normalization statistics with the keys ``"mean"``
+                and ``"std"``.
+            backbone_freeze:
+                Whether to freeze the backbone during training.
+            backbone_weights:
+                Path to the DINOv3 backbone weights.
+            backbone_args:
+                Additional arguments to pass to the DINOv3 backbone.
+            load_weights:
+                If False, then no pretrained weights are loaded.
+        """
         super().__init__(init_args=locals(), ignore_args={"load_weights"})
         parsed_name = self.parse_model_name(model_name=model_name)
 
@@ -480,7 +506,9 @@ class DINOv3LTDETRObjectDetection(TaskModel):
 
         # NOTE(Guarin, 08/25): We don't set drop_path_rate=0 here because it is already
         # set by DINOv3.
-        backbone_model_args: dict[str, Any] = {}
+        backbone_model_args: dict[str, Any] = {
+            "patch_size": patch_size,
+        }
         if backbone_args is not None:
             backbone_model_args.update(backbone_args)
         if backbone_weights is not None:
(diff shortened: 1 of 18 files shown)


tests/_task_models/dinov3_ltdetr_object_detection/test_task_model.py
```diff
@@ -112,6 +192,65 @@ def _create_train_model(
...
+        patch_size=patch_size,
+        image_normalize=None,
+        backbone_freeze=False,
+        backbone_weights=None,
+        backbone_args=None,
+        load_weights=False,
+    )
+
+
+def _fake_get_model(*args: Any, **kwargs: Any) -> Any:
```
liopeer: You will need to refactor this whenever we change the actual `get_model` method.

tests/_task_models/dinov3_ltdetr_object_detection/test_task_model.py
```diff
@@ -77,10 +82,85 @@ def test_freeze_backbone_on_set_train_mode(should_freeze: bool) -> None:
...
+        model_init_args={"patch_size": 14},
+        monkeypatch=monkeypatch,
+    )
+
+    assert model_args.patch_size == 14
+    assert calls[0]["model_args"]["patch_size"] == 14
+    assert train_model.model.backbone.patch_size == 14
+
+
+def test_resolve_auto__uses_explicit_patch_size(
```
liopeer: From what I am understanding, you're not actually testing the "real" `get_model` method, but only a fake variant of it that lives here in this test module. So when we change the real implementation, we might not see an error here, which is exactly what tests are supposed to surface.

liopeer: Note: I would avoid mocking whenever possible. Mocking is really nice in integration tests, where certain setups/calls are very expensive, but in these types of tests I would try to always go with the real deal.

liopeer: If you struggle to properly test the functionality of your implementation, you might have to change the implementation a little bit, by e.g. creating helper functions that you can test separately etc.

tests/_task_models/dinov3_ltdetr_object_detection/test_task_model.py
```diff
@@ -77,10 +82,85 @@ def test_freeze_backbone_on_set_train_mode(should_freeze: bool) -> None:
...
+    assert roundtrip_model.init_args["patch_size"] == 14
+
+
+def test_resolve_auto__uses_explicit_patch_size_for_convnext(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    model_args = DINOv3LTDETRObjectDetectionTrainArgs(patch_size=14)
+    train_model, calls = _create_train_model_with_capture(
+        model_args,
+        model_name="dinov3/convnext-small-ltdetr",
```
liopeer: If you're worried about model size we either have the `_test` models (if not available for convnext, you can add it).

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
```diff
@@ -449,12 +449,38 @@ def __init__(
...
         classes: dict[int, str],
         image_size: tuple[int, int],
+        patch_size: int,
         image_normalize: dict[str, Any] | None = None,
         backbone_freeze: bool = False,
         backbone_weights: PathLike | None = None,
         backbone_args: dict[str, Any] | None = None,
         load_weights: bool = True,
     ) -> None:
+        """Create a DINOv3 LTDETR object detection model.
```
liopeer, line 579: Thx for the docstring! We never want users to directly interact with the constructor (in the [docs](https://docs.lightly.ai/train/stable/python_api/lightly_train.html#models), we therefore don't have it documented), but it's still nice to have it!

liopeer (commented):
Implementation looks good overall (pending the already discussed things). However, I would avoid mocking in the tests.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
gabrielfruet: Yep, although users may not interact with that is good for us. Sometimes is hard to infer somethings only from the code.

liopeer: Absolutely agree. Documentation is never wrong. Just wanted to make this clear that it's not a public interface.

tests/_task_models/dinov3_ltdetr_object_detection/test_task_model.py
gabrielfruet: Addressed, no more mocking.

tests/_task_models/dinov3_ltdetr_object_detection/test_task_model.py
gabrielfruet: Not using mocking anymore

tests/_commands/test_train_task_helpers.py
```diff
@@ -7,14 +7,74 @@
...
+    None
+):
+    data_args = YOLOObjectDetectionDataArgs(
+        path=Path("/tmp/data"),
+        train=Path("train") / "images",
+        val=Path("val") / "images",
+        names={0: "class_0", 1: "class_1"},
+    )
+
+    train_model_args = cast(
```
liopeer, line 45: Can't you use a normal type hint here?

tests/_task_models/dinov3_ltdetr_object_detection/test_task_model.py
```diff
@@ -57,6 +61,16 @@ def _is_module_frozen(m: nn.Module) -> bool:
     return all(not param.requires_grad for param in m.parameters())
 
 
+@pytest.fixture()
+def dummy_yolo_detection_data_args() -> YOLOObjectDetectionDataArgs:
```
liopeer: The location of this fixture seems a bit arbitrary...I would probably move it to the top.

liopeer (approved):
Nice work. Left 2 comments, but they are suggestions and don't necessarily have to be addressed!

tests/_commands/test_train_task_helpers.py
gabrielfruet: Here we have to cast, because the `get_train_model_args` returns the super type, and we cannot assign a parent class to a child class, it'd be a violation of the Liskov substitution principle.

If we don't use cast but normal type hints instead: 

```
    train_model_args: DINOv3LTDETRObjectDetectionTrainArgs = get_train_model_args(
        model_args={"patch_size": 14},
        model_args_cls=DINOv3LTDETRObjectDetectionTrainArgs,
        total_steps=1000,
        model_name="dinov3/vitt16-notpretrained-ltdetr",
        model_init_args={},
        data_args=data_args,
    )

=========== mypy ===========

tests/_commands/test_train_task_helpers.py:45: error: Incompatible types in assignment (expression has type "TrainModelArgs", variable has type "DINOv3LTDETRObjectDetectionTrainArgs")  [assignment]
```

tests/_task_models/dinov3_ltdetr_object_detection/test_task_model.py
gabrielfruet: Rearranged stuff in this file.

1. Helper functions
2. Fixtures
3. Tests

PR #721, opened 2026-05-07
Fix PicoDet finetune with mismatched num_classes

mrpositron:
## What has changed and why?

Fixes `RuntimeError: size mismatch for internal_class_to_class` when finetuning PicoDet from a checkpoint with a different `num_classes` (e.g. COCO 80 → custom 70).

`internal_class_to_class` was a persistent buffer, so PyTorch's shape check fired before `strict` gating and raised even under `strict=False`.

Mirrors the `DINOv3LTDETRObjectDetection` pattern:
- Register the buffer with `persistent=False` (it is fully derived from `classes`, which `init_args` already persists).
- Pop the legacy key in `load_train_state_dict` so older checkpoints still load under `strict=True`.

`gfl_cls` reinit is already handled by the existing pre-hook in `pico_head.py`. 

## How has it been tested?

- Re-ran the reported COCO→70-class finetune script; loads cleanly.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_task_models/picodet_object_detection/task_model.py
+++ b/src/lightly_train/_task_models/picodet_object_detection/task_model.py
@@ -108,6 +108,7 @@ class PicoDetObjectDetection(TaskModel):
         self.register_buffer(
             "internal_class_to_class",
             torch.tensor(internal_class_to_class, dtype=torch.long),
+            persistent=False,
         )
 
         config = _MODEL_CONFIGS.get(model_name)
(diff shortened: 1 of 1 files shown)


src/lightly_train/_task_models/picodet_object_detection/task_model.py
```diff
@@ -336,12 +337,13 @@ def load_train_state_dict(
...
 
-        if "internal_class_to_class" not in new_state_dict:
-            new_state_dict["internal_class_to_class"] = (
-                self.internal_class_to_class.detach().clone()
-            )
+        # TODO(Nauryzbay, 05/2026): Remove this pop once no PicoDet checkpoints
+        # with the legacy persistent internal_class_to_class buffer exist in
+        # the wild. It is only kept for backwards compatibility with checkpoints
+        # saved before the buffer became non-persistent.
+        new_state_dict.pop("internal_class_to_class", None)
```
yutong-xiang-97: Let's remove the pop for now. It is a good way to see if people actually use PicoDet. Also add a follow-up to update the picodet checkpoints.

yutong-xiang-97 (approved):

PR #723, opened 2026-05-08
Fix RF-DETR hosted model names for latest upstream release

gabrielfruet:
## What has changed and why?

Update RF-DETR model discovery to use the latest upstream `ModelWeights` list, add support for the new model names, and keep `load_weights=False` working.

## How has it been tested?

Updated unit coverage in `tests/_models/rfdetr/test_rfdetr_package.py`.

- [x] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_models/rfdetr/rfdetr_package.py
+++ b/src/lightly_train/_models/rfdetr/rfdetr_package.py
@@ -35,13 +35,12 @@ class RFDETRPackage(Package):
     @classmethod
     def list_model_names(cls) -> list[str]:
         try:
-            from rfdetr.main import HOSTED_MODELS
+            from rfdetr.assets.model_weights import ModelWeights
         except ImportError:
             return []
-        # We use the model names from the checkpoint .pth filenames Roboflow provided
         return [
-            f"{cls.name}/{model_name.split('.')[0]}"
-            for model_name in HOSTED_MODELS.keys()
+            f"{cls.name}/{Path(model_name).stem}"
+            for model_name in ModelWeights.list_models()
         ]
 
     @classmethod
@@ -59,15 +58,21 @@ class RFDETRPackage(Package):
         load_weights: bool = True,
     ) -> RFDETR:
         try:
-            from rfdetr import (
+            from rfdetr.detr import (
                 RFDETRBase,
                 RFDETRLarge,
                 RFDETRMedium,
                 RFDETRNano,
+                RFDETRSeg2XLarge,
+                RFDETRSegLarge,
+                RFDETRSegMedium,
+                RFDETRSegNano,
                 RFDETRSegPreview,
+                RFDETRSegSmall,
+                RFDETRSegXLarge,
                 RFDETRSmall,
             )
-            from rfdetr.main import HOSTED_MODELS
+            from rfdetr.assets.model_weights import ModelWeights
         except ImportError:
             raise ValueError(
                 f"Cannot create model '{model_name}' because rfdetr is not installed."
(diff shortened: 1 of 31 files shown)


src/lightly_train/_models/rfdetr/rfdetr_package.py
```diff
@@ -59,15 +58,21 @@ def get_model(
...
     ) -> RFDETR:
         try:
-            from rfdetr import (
+            from rfdetr.assets.model_weights import ModelWeights
+            from rfdetr.detr import (
                 RFDETRBase,
                 RFDETRLarge,
                 RFDETRMedium,
                 RFDETRNano,
+                RFDETRSeg2XLarge,
```
yutong-xiang-97: Have you tested whether these new models can be trained properly? I don't think the unit tests are enough to verify this.

yutong-xiang-97: If it does not work we can simply exclude them. No need to spend too much time on this.

yutong-xiang-97: Then you would also need to update the docs: https://docs.lightly.ai/train/stable/pretrain_distill/models/rfdetr.html

yutong-xiang-97: Also have a look at the RF-DETR versions for the support for these new models. RF-DETR makes a lot of breaking changes from 1.1.0-1.6.0 and we may not be able to accommodate for everything.

src/lightly_train/_models/rfdetr/rfdetr_package.py
```diff
@@ -82,40 +87,44 @@ def get_model(
...
+            "rf-detr-nano": RFDETRNano,
+            "rf-detr-small": RFDETRSmall,
+            "rf-detr-medium": RFDETRMedium,
+            "rf-detr-seg-preview": RFDETRSegPreview,
+            "rf-detr-seg-nano": RFDETRSegNano,
+            "rf-detr-seg-small": RFDETRSegSmall,
+            "rf-detr-seg-medium": RFDETRSegMedium,
+            "rf-detr-seg-large": RFDETRSegLarge,
+            "rf-detr-seg-xlarge": RFDETRSegXLarge,
+            "rf-detr-seg-xxlarge": RFDETRSeg2XLarge,
```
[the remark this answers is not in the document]
yutong-xiang-97: Same as I mentioned. Beware of the versions.

src/lightly_train/_models/rfdetr/rfdetr_package.py
gabrielfruet: I think for now would be best to just not add them, we can verify it later. WDYT? 

Maybe we can create an issue for that.

yutong-xiang-97: Yes, for now just let mypy pass. You can even add `type:ignore`s for that. We may need to revisit the whole RF-DETR support thing later.

gabrielfruet: I modified to not use the new models, only the old ones.

Did you meant that I should let the code be the same as previously and just add the type ignore?

yutong-xiang-97: I would propose use `OPEN_SOURCE_MODELS` like you did in your previous commit in another PR. Now it works for all the RF-DETR versions we support
- https://github.com/lightly-ai/lightly-train/blob/09b2ab980729345cbdcd18d47ef34a2a7c994c3b/pyproject.toml#L92
- https://github.com/roboflow/rf-detr/blob/1.5.2/src/rfdetr/main.py

and add type:ignore if necessary. WDYT?

And also create an issue for this to look into our RF-DETR support.

gabrielfruet: I think it's a good idea. We can take a closer look at the RF-DETR when we have more time. I'll address this on Monday.

gabrielfruet: I'm thinking about closing this issue, within the versions we support, I have no problems with mypy type checking. 

Or shall I just change for the `OPEN_SOURCE_MODELS` ? That anyway is already deprecated in newer versions. WDYT?

yutong-xiang-97: Yes we can close this PR for now. I've already created another issue to look into RF-DETR support as a whole.

PR #726, opened 2026-05-11
Refactor training visualizations into task-specific handlers

mrpositron:
## What has changed and why?

This PR refactors training-step visualization so plotting is handled outside the individual train models and generated lazily through task-specific visualization helpers.

- Added a new `train_task_visualization` helper to centralize saving train/validation example images.
- Replaced eager `label_image` / `prediction_image` creation in `TaskStepResult` with a lazy `visualization` payload that can create images on demand.
- Updated train models across image classification, object detection, and segmentation tasks to return task-specific visualization objects instead of pre-rendered PIL images.
- Standardized visualization inputs to use `image_normalize` and `class_names`.
- Removed the object detection visualization cap on predicted boxes.

This keeps train models focused on training logic, removes duplicated plotting/saving code, and avoids carrying fully rendered images through the training loop when they may never be written. It also makes visualization behavior more consistent across tasks and easier to extend.

## How has it been tested?

- Updated visualization unit tests for image classification and object detection to match the refactor and new plotting signatures.

- [ ] Yes

- [ ] Yes


liopeer (approved):
LGTM!

src/lightly_train/_visualize/object_detection.py
```diff
@@ -16,21 +18,51 @@
 from lightly_train.types import ObjectDetectionBatch
 
 
+@dataclass
+class ObjectDetectionTaskStepVisualization:
+    batch: ObjectDetectionBatch
+    class_names: dict[int, str]
+    image_normalize: dict[str, tuple[float, ...]] | None
+    max_images: int
+    score_threshold: float = 0.0
```
liopeer: This is a super low default, should this not be 0.5-0.7?

mrpositron: `score_threshold` is only used for prediction visualizations; label visualizations ignore it since ground-truth boxes have no scores.

In this training flow, predictions pass `self.viz_score_threshold = 0.1` explicitly, so `0.0` is only the dataclass fallback. That said, I agree a safer default would be better.

PR #730, opened 2026-05-13
Transition to `uv sync`

liopeer:
## What has changed and why?
- this change is supposed to do away with the requirement for manual venv handling
- as soon as this is approved, I will do 2 things:
  - make a PR to the local CI workflows
  - as soon as the local CI workflows PR is approved I will here change the checks to be fulfileld for merging and then merge both PRs

## How has it been tested?
- tried all different install commands locally

- [ ] Yes

- [ ] Yes


.pre-commit-config.yaml
```diff
@@ -10,6 +10,7 @@ repos:
     - id: check-added-large-files
       description: "Avoid committing large files."
       args: ["--maxkb=2000", "--enforce-all"]
+      exclude: ^uv\.lock$
```
liopeer, line 13: `uv.lock` is 24MB large

Makefile
```diff
@@ -244,40 +244,33 @@ endif
 # 
 # This problem is fixed in pywinpty>2.0.14 but these versions are not compatible with
 # Python 3.8.
-#
+
+comma := ,
+to_uv_extras = --extra $(subst $(comma), --extra ,$(subst ],,$(subst [,,$(1))))
```
liopeer: A formula to translate the pip dependencies in the format `[dev,mflow,...]` to `uv sync --extra dev --extra mlflow --extra ...`.

Makefile
```diff
@@ -244,40 +244,33 @@ endif
...
+
 # SuperGradients is excluded as it is outdated and causes issues in CI.
-EXTRAS_PY38 := [dev,dicom,mlflow,onnx,tensorboard,timm,ultralytics,wandb]
+EXTRAS_PY38 := [dicom,mlflow,onnx,tensorboard,timm,ultralytics,wandb]
```
liopeer, line 257: `dev` is removed, because it's not an extra anymore, instead it's now a group. This means that we don't distribute the `dev` extra via PyPI anymore, which imo is desirable, no one wants to be able to install `pip install lightly-train[dev]`.

Makefile
```diff
@@ -244,40 +244,33 @@ endif
...
 # default development dependencies.
 
 # SuperGradients is excluded as it is not compatible with Python>=3.10.
-EXTRAS_DEV := [dev,dicom,mlflow,notebook,onnx,onnxruntime,onnxslim,rfdetr,tensorboard,timm,ultralytics,wandb]
+EXTRAS_DEV := [dicom,mlflow,notebook,onnx,onnxruntime,onnxslim,rfdetr,tensorboard,timm,ultralytics,wandb]
 
 # Exclude ultralytics from docker extras as it has an AGPL license and we should not
 # distribute it with the docker image.
-DOCKER_EXTRAS := --extra mlflow --extra tensorboard --extra timm --extra wandb --extra rfdetr
+DOCKER_EXTRAS := [mlflow,tensorboard,timm,wandb,rfdetr]
```
liopeer, line 267: Idk why this was already like this.

pyproject.toml
```diff
@@ -48,13 +48,14 @@ dependencies = [
...
+    "numpy>1.23.0; python_version>='3.12'",  # numpy<=1.23 fails to build on Python>3.12 (no distutils).
 ]
 
-[project.optional-dependencies]
+[dependency-groups]
 dev = [
     "licenseheaders>=0.8.8",
     "mdformat>=0.7.21; python_version>='3.9'",  # Older versions change formatting behavior
-    "mdformat-myst>=0.3.0; python_version>='3.9'", # MyST support for mdformat
+    "mdformat-myst>=0.3.0; python_version>='3.10'", # MyST support for mdformat
```
liopeer, line 58: Was apparently wrong, and only supported for Python>=3.10.

pyproject.toml
```diff
@@ -77,24 +78,46 @@ dev = [
     "sphinxcontrib-googleanalytics>=0.4",
     "sphinx>=7.1",              # Allow old sphinx version for Python 3.8 compatibility
 ]
+
+# Pinned versions for Torch and TorchVision to avoid issues with the CUDA/driver version
```
liopeer, line 87: These are really just for development and have 0 influence on how the package can be installed from PyPI.

pyproject.toml
```diff
@@ -111,9 +134,11 @@ xformers = [
 ]
 onnx = [
     "onnx>=1.15.0",
+    "onnx>=1.18.0; python_version>='3.9'",  # onnx 1.18+ requires Python 3.9+
```
simonschoelly, line 142: I wonder if we should not just group `onnx`, `onnxslim` and `onnxruntime` together - or at least `onnx` and `onnxruntime`.

One could just use `onnx` to export - but as soon as the slim option is used, `onnxslim` comes into play and as soon as verification is used, `onnxruntime` is used.

liopeer: Fair point. I think we could indeed do that. I don't really see a reason not to group them tbh.

pyproject.toml
```diff
@@ -197,8 +222,40 @@ convention = "google"
 [tool.uv]
 override-dependencies = [
   "sphinx>=7.1",              # Override dependency from SuperGradients on sphinx<=4.1
+  "onnx>=1.15.0",             # Override dependency from SuperGradients on onnx==1.15.0
```
liopeer: This seems outdated.

liopeer: Done

pyproject.toml
liopeer: I left it for now. The diff was already big enough.

src/lightly_train/_models/rfdetr/rfdetr_package.py
```diff
@@ -29,19 +29,40 @@ class RFDETR:  # type: ignore[no-redef]
...
+    Raises ImportError if rfdetr is not installed.
+    Raises RuntimeError if rfdetr is installed but the registry constant is missing
+    (indicates an incompatible rfdetr version).
+    """
+    import rfdetr.main as rfdetr_main
+
+    # rfdetr renamed HOSTED_MODELS to OPEN_SOURCE_MODELS in v1.x
+    registry: dict[str, str] | None = getattr(
+        rfdetr_main, "OPEN_SOURCE_MODELS", None
+    ) or getattr(rfdetr_main, "HOSTED_MODELS", None)
```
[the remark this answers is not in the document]
liopeer: Yeah, no. We can leave that.

Makefile
```diff
@@ -246,78 +251,44 @@ endif
...
+EXTRAS_PY313 := [dicom,mlflow,notebook,onnx,onnxruntime,onnxslim,rfdetr,tensorboard,timm,ultralytics,wandb]
 
 # SuperGradients is excluded as it is not compatible with Python>=3.10.
-EXTRAS_DEV := [dev,dicom,mlflow,notebook,onnx,onnxruntime,onnxslim,rfdetr,tensorboard,timm,ultralytics,wandb]
+EXTRAS_DEV := [dicom,mlflow,notebook,onnx,onnxruntime,onnxslim,rfdetr,tensorboard,timm,ultralytics,wandb]
 
 # Exclude ultralytics from docker extras as it has an AGPL license and we should not
 # distribute it with the docker image.
-DOCKER_EXTRAS := --extra mlflow --extra tensorboard --extra timm --extra wandb --extra rfdetr
+DOCKER_EXTRAS := [mlflow,tensorboard,timm,wandb,rfdetr]
```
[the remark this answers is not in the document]
liopeer: Valid point. Will address.

Makefile
```diff
@@ -327,61 +298,65 @@ install-dev:
...
 # version available at EXCLUDE_NEWER_DATE.
 #
 # See install-pinned-3.8 for more information.
 .PHONY: install-pinned-3.13
 install-pinned-3.13:
-	uv pip install --exclude-newer ${EXCLUDE_NEWER_DATE} --reinstall ${EDITABLE} ".${EXTRAS_PY313}" --requirement pyproject.toml \
-		${PINNED_TORCH_VERSION_PY313} ${PINNED_TORCHVISION_VERSION_PY313}
+	uv sync --frozen --python=${MAXIMAL_PYTHON_VERSION} \
+		--exclude-newer ${EXCLUDE_NEWER_DATE} ${NO_EDITABLE} --group dev \
+		--group pinned-torch-maximal $(call to_uv_extras,$(EXTRAS_PY313))
```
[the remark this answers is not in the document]
liopeer: `--upgrade-group` sounds amazing, will check if that's a thing.

Makefile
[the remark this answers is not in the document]
liopeer: Done.

Makefile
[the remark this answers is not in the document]
liopeer: Done.

simonschoelly:
Probably not a big concern, but I tried a completely new installation, and by default it does not pick up the lockfile
```bash
 simon@simontower  ~/lighlty/lightly-train   lionel-dependency-locking  uv sync
Using CPython 3.13.1
Creating virtual environment at: .venv
Resolving despite existing lockfile due to removal of global exclude newer
```

src/lightly_train/_models/rfdetr/rfdetr_package.py
```diff
@@ -29,19 +29,40 @@ class RFDETR:  # type: ignore[no-redef]
...
+    if registry is None:
+        raise RuntimeError(
+            "rfdetr is installed but neither 'OPEN_SOURCE_MODELS' nor 'HOSTED_MODELS' "
+            "was found in rfdetr.main. The installed rfdetr version may be incompatible."
```
simonschoelly, line 48: Could we tell the user what rfdetr version they need to install? And could we not enforce the rfdetr version in pyproject.toml?

simonschoelly:
I noticed that we also need to use `--frozen` just when using `uv run python` otherwise the lock file gets recreated. Perhaps there is nothing we can do here, but it is a bit unfortunate.

simonschoelly:
Should we update documentation and README.md? Or should we keep `pip install` there?

src/lightly_train/_models/rfdetr/rfdetr_package.py
liopeer: I think we generally want to be able to support as many versions as possible.

liopeer:
> Probably not a big concern, but I tried a completely new installation, and by default it does not pick up the lockfile

I deliberately use `--exclude-newer` with the lock file so that it does not constantly change (see `make lock` command). Equally, you actually need to use `uv sync --frozen`, otherwise it checks again if there should be an update to the lock. The `make install-dev` should avoid that.

> Should we update documentation and README.md? Or should we keep pip install there?

`uv sync` is only for development. It has no influence on the distribution via PyPI, so the README/docs should not be updated (only the `CONTRIBUTING.md`).

src/lightly_train/_models/rfdetr/rfdetr_package.py
simonschoelly: I mean just a meaningful error message that tells the user what bounds to use - otherwise they just see a message that they have the wrong version but don't know which one to install.

simonschoelly (approved):

PR #736, opened 2026-05-18
Adds image logging to ML tracking platforms

mrpositron:
## What has changed and why?

Training/validation example images (labels and predictions) were previously only saved to disk under `image_examples/`. The Fabric loggers (TensorBoard, MLflow, Wandb) had no visibility into them, so users running with those loggers couldn't see the visualizations in their dashboards.

This PR routes those visualizations through the active Fabric loggers in addition to saving them to disk:

- Added `logger_helpers.log_image_to_loggers` that dispatches a PIL image to each logger using its native image API: add_image for `TensorBoardLogger`, log_image for `WandbLogger` / `MLFlowLogger`. Loggers without image supportare silently skipped.
- `save_train_step_visualizations` and `save_val_step_visualizations` now accept loggers and forward each saved image. Validation also receives the `global_step` so logged points align with the global training step on the dashboard.
- `_train_task_from_config` passes fabric.loggers (and the global step on val) into the visualization helpers.
Logging keys use the `train_images/labels_{step}` and `val_images/{predictions,labels}_{val_step}` format so each step lands as a separate entry in the loggers' image galleries.

### MLFLow Logging

Note: older MLflow server versions don't group logged images into nested sections. When logging an image with key `val_images/image_0`, instead of creating a val_images section, MLflow renders it as `val_images~image_0` under the Model metrics tab. This is a UI limitation, not a bug, and is fixed in `MLflow >3.0` - we recommend installing version 3.11.1.

image

### TensorBoard Logging

image

### WandB Logging

WandB logging also works properly.

## How has it been tested?

- Added unit tests in `tests/_loggers/test_logger_helpers.py`:
  - `test_log_image_to_loggers__tensorboard` - writes an image via the helper and reads it back with EventAccumulator to assert the tag and step are correct.
  - `test_log_image_to_loggers__jsonl_skipped` - confirms a logger without image support is silently skipped (no exception).
  - `test_log_image_to_loggers__no_loggers` - confirms the helper is a no-op with an empty logger list.

- Verified end-to-end by running a training job locally with MLflow and TensorBoard loggers enabled and confirming the train/val images appear in each dashboard at the expected steps (see screenshots above).

- [ ] Yes

- [ ] Yes


src/lightly_train/_commands/train_task_visualization.py
```diff
@@ -7,21 +7,32 @@
...
+from lightning_fabric.loggers.logger import Logger as FabricLogger
 from PIL.Image import Image as PILImage
 
+from lightly_train._loggers import logger_helpers
 from lightly_train._task_models.train_model import TaskStepResult
 
 _IMAGE_EXAMPLES_DIR = "image_examples"
 
 
 def save_train_step_visualizations(
```
yutong-xiang-97: Maybe I wasn't involved in the full loop, but why a separate module of `train_task_visualization.py` instead of adding them to `train_task_helpers.py` if you don't expose them as public APIs?

src/lightly_train/_commands/train_task_visualization.py
```diff
@@ -30,27 +41,44 @@ def save_val_step_visualizations(
...
+            step=global_step,
         )
 
 
 def _image_examples_dir(out_dir: Path) -> Path:
     return out_dir / _IMAGE_EXAMPLES_DIR
 
 
-def _save_image(*, image: PILImage | None, path: Path) -> None:
+def _save_and_log(
```
yutong-xiang-97: `_save_and_log` sounds a bit obscure to me. Maybe use a more concrete name like `_save_images_to_dir_and_loggers`

src/lightly_train/_loggers/logger_helpers.py
```diff
@@ -63,3 +67,26 @@ def get_loggers(logger_args: LoggerArgs, out: Path) -> list[Logger]:
...
+    ``MLFlowLogger``. Other loggers, such as ``JSONLLogger``, are silently
+    skipped.
+    """
+    image_tensor = None
+    for log in loggers:
+        if isinstance(log, TensorBoardLogger):
+            if image_tensor is None:
+                image_tensor = torchvision_functional.pil_to_tensor(image)
+            log.experiment.add_image(tag=key, img_tensor=image_tensor, global_step=step)
+        elif isinstance(log, (WandbLogger, MLFlowLogger)):
```
yutong-xiang-97, line 91: Raise an error if the loggers are not recognized as a safeguard.

tests/_loggers/test_logger_helpers.py
```diff
@@ -161,3 +163,37 @@ def test_get_callbacks__mlflow_user_config(tmp_path: Path) -> None:
...
+
+def test_log_image_to_loggers__jsonl_skipped(tmp_path: Path) -> None:
+    # JSONLLogger has no image support; the helper must silently skip it.
+    jsonl_logger = JSONLLogger(save_dir=tmp_path)
+    image = PILImageModule.new("RGB", (4, 4))
+
+    logger_helpers.log_image_to_loggers(
+        loggers=[jsonl_logger], key="train/labels", image=image, step=0
+    )
+
```
yutong-xiang-97, line 200: Here it looks confusing since nothing is asserted. Instead we can use `mocker.spy`

```
log_metrics_spy = mocker.spy(jsonl_logger, "log_metrics")
...

log_metrics_spy.assert_not_called()
```

tests/_loggers/test_logger_helpers.py
```diff
@@ -161,3 +163,37 @@ def test_get_callbacks__mlflow_user_config(tmp_path: Path) -> None:
...
+
+def test_log_image_to_loggers__no_loggers() -> None:
```
yutong-xiang-97: What is this test supposed to do? You are essentially testing the built-in empty list behavior of Python at the moment.

yutong-xiang-97 (commented):
Reviewed. 

Re. tests for MLflow and WandB I would say it is fine that you use an actual logger instance (isntead of mocking). 
- for MLflow: you can use a local tracking URI (starting with `file://`
- for WandB: not so familiar but it is said that it has an offline mode. However, if it turns out to be a hassle, we can simply skip it.

src/lightly_train/_commands/train_task_visualization.py
mrpositron: Mostly for readability. `train_task_helpers.py` is about 1.5k lines of code. But I guess that is not really important. I will move the code from `train_task_visualization.py` there.

yutong-xiang-97: We might need to refactor `train_task` along with helpers but since it is the main loop I would be really careful.

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -1494,3 +1497,71 @@ def get_torch_compile_args(
         train_model_cls.torch_compile_args_cls, torch_compile_args
     )
     return args
+
+
+_IMAGE_EXAMPLES_DIR = "image_examples"
```
yutong-xiang-97: Put all constants on top under all imports.

src/lightly_train/_loggers/logger_helpers.py
yutong-xiang-97: Now I think it would make more sense if you raise an warning and just skip this logger.

yutong-xiang-97 (approved):
Approved with two small comment.

PR #738, opened 2026-05-20
Add Predict Batch Method to Object Detection

mrpositron:
## What has changed and why?

Added a predict_batch method to `DINOv3LTDETRObjectDetection` for running inference on multiple images in a single forward pass.

Changes:

- `TaskModel` base class - added an abstract inference pipeline split into **four overridable stages**: `preprocess_image` (per-image preprocessing + metadata extraction), `preprocess_batch` (batch-level normalization on stacked tensor), `forward_backend` (model forward pass), and `postprocess` (raw output → per-image result dicts). The `predict_batch` method in the base class composes these stages, so subclasses only need to implement the stages rather than the full batching loop.

- `DINOv3LTDETRObjectDetection` - implemented all four stages by extracting logic from the existing `predict` method and refactored the expected-channel calculation into `preprocess_image`. `predict_batch` delegates to the base class after ensuring deploy mode is active.

The existing `predict` and `forward` methods were not touched to preserve backward compatibility and keep existing tests passing.

## How has it been tested?

```python
import lightly_train
import torch
import os

if __name__ == "__main__":
    images = [
        "/coco128_yolo/images/val2017/000000012120.jpg",
        "/coco128_yolo/images/val2017/000000011760.jpg",
        "/coco128_yolo/images/val2017/000000007888.jpg",
        "/coco128_yolo/images/val2017/000000003661.jpg",
    ]
    model = lightly_train.load_model(
        model="dinov3/vitt16-ltdetr-coco",
        device="cpu",
    )

    results_single = [model.predict(image=img) for img in images]
    results_batch = model.predict_batch(images=images)

    for i, (single, batch) in enumerate(zip(results_single, results_batch)):
        name = os.path.basename(images[i])
        n_single = single["bboxes"].shape[0]
        n_batch = batch["bboxes"].shape[0]
        if n_single != n_batch:
            print(f"Image {i} ({name}): detection count mismatch — predict={n_single}, predict_batch={n_batch}")
            continue
        max_diff = (single["bboxes"] - batch["bboxes"]).abs().max().item()
        bboxes_close = torch.allclose(single["bboxes"], batch["bboxes"], atol=1e-4)
        labels_match = torch.equal(single["labels"], batch["labels"])
        print(f"Image {i} ({name}): n={n_single}, bboxes_close={bboxes_close}, max_diff={max_diff:.2e}, labels={labels_match}")

```
Output:
```bash
Image 0 (000000012120.jpg): n=10, bboxes_close=True, max_diff=6.10e-05, labels=True
Image 1 (000000011760.jpg): n=3, bboxes_close=True, max_diff=7.63e-05, labels=True
Image 2 (000000007888.jpg): n=2, bboxes_close=True, max_diff=6.10e-05, labels=True
Image 3 (000000003661.jpg): n=5, bboxes_close=True, max_diff=1.83e-04, labels=True
```

- [ ] Yes

- [ ] Yes


mrpositron:
@copilot+gpt-5.4

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
```diff
@@ -799,6 +800,101 @@ def deploy(self) -> Self:
...
+                  (x_min, y_min, x_max, y_max) in absolute pixel coordinates of the
+                  original image.
+                - "scores": Tensor of shape (N,) with confidence scores.
+        """
+        if self.training or not self.postprocessor.deploy_mode:
+            self.deploy()
+        return super().predict_batch(images, threshold=threshold)
+
     @torch.no_grad()
     def predict(
```
yutong-xiang-97, line 925: You can already refactor `predict()` with your new preprocessing and postprocessing methods!

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
```diff
@@ -799,6 +810,103 @@ def deploy(self) -> Self:
...
+        out: list[dict[str, Tensor]] = []
+        labels_batch, boxes_batch, scores_batch = postprocessor_out
+
+        labels_batch = self.internal_class_to_class[labels_batch]
+        for i in range(len(metadata)):
+            keep = scores_batch[i] > threshold
+            out.append(
+                {
+                    "labels": labels_batch[i][keep],
+                    "bboxes": boxes_batch[i][keep],
```
yutong-xiang-97, line 880: If you output original bbox coordinates, then it must export `orig_target_size` as a second input. Alternatively you can keep exported boxes are in resized model-input coordinates, but then users must rescale boxes outside ONNX.

cc @simonschoelly for visibility

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
yutong-xiang-97: Now that we will have modules to deal with backend-related logic, we should actually think about what `TaskModel.forward()` should actually do. For now it seems that it is for ONNX export and also contains `forward_backend` and post processing. Maybe it can be as simple as `forward_backend` and the rest can be refactored out to pre- and post-processors or somewhere in the backend handling logic. I have a rough draft but seems like it will affect current `export_onnx`.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
```diff
@@ -799,6 +810,103 @@ def deploy(self) -> Self:
...
+        # 1-channel expansion is a convenience inherited from RGB-only models.
+        expected_c = self._expected_input_channels
+        if x.shape[-3] == 1 and expected_c > 1:
+            x = x.expand(expected_c, -1, -1)
+        elif x.shape[-3] != expected_c:
+            raise ValueError(
+                f"Image has {x.shape[-3]} channels but model expects {expected_c}."
+            )
+
+        x = transforms_functional.to_dtype(x, dtype=dtype, scale=True)
```
liopeer, line 833: One thing that just came to my mind. `preprocess_image` and `preprocess_batch` will also need to support label propagation. This is not your task to implement right now, but it means we need a nice abstraction here. And a decision if we want to see it as a chance to move to torchvision for the transforms.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
```diff
@@ -799,6 +810,103 @@ def deploy(self) -> Self:
...
+        return batch
+
+    def forward_backend(self, x: Tensor) -> Any:
+        x = self.backbone(x)
+        x = self.encoder(x)
+        return self.decoder(x)
+
+    def postprocess(  # type: ignore[override]
+        self,
+        raw_outputs: Any,
```
liopeer: You can also type as `raw_outputs: Any | MoreNarrowType` to make it more legible, and then add a ValueError if it's not the right thing.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
```diff
@@ -799,6 +810,103 @@ def deploy(self) -> Self:
...
+        metadata: Sequence[dict[str, Any]],
+        threshold: float = 0.6,
```
liopeer: I expect this to be an internal method, so I would not use any defaults.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
liopeer: I think eventually it should become what `forward_backend` is now. But I would do it step-wise and keep using `forward_backend` for now and leave `forward` just be anything (whatever this model requires.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
```diff
@@ -799,6 +810,103 @@ def deploy(self) -> Self:
...
+        Returns:
+            A list with one dict per input image. Each dict contains:
+                - "labels": Tensor of shape (N,) with predicted class indices.
+                - "bboxes": Tensor of shape (N, 4) with bounding boxes in
+                  (x_min, y_min, x_max, y_max) in absolute pixel coordinates of the
+                  original image.
+                - "scores": Tensor of shape (N,) with confidence scores.
+        """
+        if self.training or not self.postprocessor.deploy_mode:
+            self.deploy()
```
liopeer, line 912: We should really try to get rid of this. The behaviour of a torch model should not be different, based on that.

src/lightly_train/_task_models/task_model.py
```diff
@@ -90,6 +92,74 @@ def predict(self, image: PathLike | PILImage | Tensor) -> Any:
         """
         raise NotImplementedError()
 
+    def preprocess_image(
```
liopeer, line 94: We can also turn `TaskModel` into an `ABC` --> `TaskModel(ABC, Module)`, then you can configure them as 
```
@abstractmethod
preprocess_image(): ...
```
and it will already fail at instantiation.

src/lightly_train/_task_models/task_model.py
```diff
@@ -90,6 +92,74 @@ def predict(self, image: PathLike | PILImage | Tensor) -> Any:
...
+        self,
+        raw_outputs: Any,
+        metadata: Sequence[dict[str, Any]],
+        **kwargs: Any,
+    ) -> list[Any]:
+        """Map raw outputs and per-image metadata into one result dict per image."""
+        raise NotImplementedError()
+
+    @torch.no_grad()
+    def predict_batch(
```
liopeer, line 127: This should imo also only be abstract, not implemented here. Inheritance makes sense in some cases (especially if you can inherit the whole method), but already in the the current case, we actually have to override the method again and resort to a `super().predict_batch` call. If it's anyway not fully inheritable, I would prefer putting that logic into a helper function (for now that can be in the module of the OD model, if it proves transferrable between models, we can move it to some helper module).

liopeer (changes requested):
Looking correct, all comments mainly a question of taste.

src/lightly_train/_task_models/task_model.py
liopeer: This is purely optional!

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
mrpositron: Do you want this issue to be addressed in a current PR? Or should we address in a separate refactoring PR?

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
```diff
@@ -799,6 +810,117 @@ def deploy(self) -> Self:
...
+        x = file_helpers.as_image_tensor(image).to(device)
+        image_h, image_w = x.shape[-2:]
+
+        # Expand grayscale to the expected channel count so images can be stacked.
+        # TODO(Nauryzbay, 05/26): Revisit grayscale handling — the implicit
+        # 1-channel expansion is a convenience inherited from RGB-only models.
+        expected_c = self._expected_input_channels
+        if x.shape[-3] == 1 and expected_c > 1:
+            x = x.expand(expected_c, -1, -1)
+        elif x.shape[-3] != expected_c:
```
simonschoelly, line 828: I think I also saw some 4-channel code (RGBA?) do we also need logic for that?

src/lightly_train/_task_models/task_model.py
```diff
@@ -90,6 +91,52 @@ def predict(self, image: PathLike | PILImage | Tensor) -> Any:
...
+        """Batch-level preprocessing on a stacked (B, C, H, W) tensor.
+
+        Kept separate from `preprocess_image` so this stage could be baked into an
+        exported graph (ONNX/TRT) — torchvision.transforms.v2 ops are batch-
+        friendly and run on GPU.
+        """
+        raise NotImplementedError()
+
+    def forward_backend(self, x: Tensor) -> Any:
+        """Run the model on a batched input and return raw outputs (pre-postprocess)."""
```
simonschoelly, line 115: Is `forward_batch` a better name? And if I understand correctly this does not run pre- and postprocessing? The docstring is a bit unclear. Or why do we not just use the `forward` method here?

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
simonschoelly: I think most important is, that we correctly document this for ONNX export.

src/lightly_train/_task_models/task_model.py
mrpositron: The idea of the function `forward_backend` is to run a forward pass through the model. Existing `forward` function includes processing steps.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
liopeer: Separate, but would be great if you could write the issue for it.

src/lightly_train/_task_models/task_model.py
simonschoelly: I see - will the processing steps work with ONNX export? Otherwise we will need to create a wrapper model for the ONNX export that calls forward_batch - not a very complicated thing though.

simonschoelly:
One argument for not doing the pre/postprocessing step in forward, but having all models use `forward` by doing `model.(...)` is that one can register hooks for the `forward` method.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
mrpositron: As far as I know, the model should not support 4-channel images. The normalization parameters, `mean` and `std`, are defined for 3 channels.

src/lightly_train/_task_models/task_model.py
mrpositron: At this point I have not looked at ONNX export.

src/lightly_train/_task_models/task_model.py
mrpositron: Agree, but I think it's worth its own PR, it touches every TaskModel subclass. Opened TRN-2060 to follow up.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
mrpositron: Opened TRN-2059 to follow up.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
gabrielfruet: @mrpositron why the `forward` does not use `forward_backend` in its underlyings?

mrpositron: The PR focuses on adding batch prediction. I’d prefer not to change `forward` here to preserve backward compatibility.

gabrielfruet: IMHO I don't like the idea of having duplicated `forward_backend` logic in the model, it's easy for us to forget about (since it's a refactor and not high priority) and leave it that way. Someone may modify the `forward` and not the `forward_backend` and then we would have a bug.

mrpositron: It is not exactly duplicated logic: the current `forward` behaves differently from `forward_backend`, since the processing step is included in `forward`. This should be definetely refactored, but I would prefer to keep this PR focused and address that separately.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
liopeer: We do actually support multi-channel: https://docs.lightly.ai/train/stable/data/multi_channel.html

liopeer: But then those parameters that you are querying for should indeed be there.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
gabrielfruet: The 3 lines from `forward_backend` is duplicated in `forward` and they have the exact same purpose: Running the underlying NN. If you'd prefer to not address on this PR, I'd ask you to create an issue for that and we do it in the next sprint or in a follow up PR. This can be very easily source of bugs in the future.

mrpositron: Yes, those lines are duplicated, but `forward` currently has different semantics: it returns the processed output, which is used by `predict_sahi` and `predict`. My intention was to avoid changing that behavior in this PR and keep the scope focused on batch prediction. I agree this should be refactored, so I will create a follow-up issue/PR.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
mrpositron: This PR doesn't touch ONNX-related functions (`forward` / `export_onnx`). I will add a TODO in `export_onnx` to expose `orig_target_size` as a second ONNX input when `forward` and `predict` functions are refactored.

src/lightly_train/_task_models/dinov3_ltdetr_object_detection/task_model.py
liopeer: Fully agree with @mrpositron 's approach here. There will anyway be significant changes still to these interfaces in the next few sprints, so let's make sure we merge.

src/lightly_train/_task_models/task_model.py
liopeer: Yup, it's not actually possible right now without proper implementations of the other subclasses.

tests/_task_models/dinov3_ltdetr_object_detection/test_task_model.py
```diff
@@ -389,6 +391,45 @@ def test_rtdetr_transformer_v2_config__resolve_auto__patch_size(
     config.resolve_auto(patch_size=patch_size)
 
 
+def test_predict_batch__composes_stages_in_order(mocker: MockerFixture) -> None:
+    model = DINOv3LTDETRObjectDetection(
```
liopeer, line 395: Optional: Use smaller image size and even smaller `_vittest16` here for faster tests.

liopeer (approved):
LGTM!

PR #719, opened 2026-05-06
Add Flat Cosine Scheduler for LT-DETR

gabrielfruet:
- **feat: add lt-detr scheduler selector**
- **test: cover lt-detr scheduler branching**
- **test: remove redundant helper coverage**
- **fix: raise on unknown scheduler mode**

## What has changed and why?

- New `scheduler` TrainArg for choosing between `linear` and `flat-cosine`
- `get_optimizer` now can either return a LinearLR or `WarmupCosineScheduler`

## How has it been tested?

- Testing if the code path properly selects the correct scheduler

## Did you update [CHANGELOG.md](../CHANGELOG.md)?

- [x] Yes
- [ ] Not needed (internal change)

## Did you update the documentation?

- [x] Yes
- [ ] Not needed (internal change without effects for user)

PR #740, opened 2026-05-21
add resolve auto for scheduler params

yutong-xiang-97:
## What has changed and why?

Automatically resolve LT-DETR scheduler defaults from the model setup instead of hardcoding them, following suggestions in https://github.com/lightly-ai/lightly-train/pull/719#discussion_r3208438688. This way users get correctly scaled LT-DETR defaults the same way as augmentations by default.

Notably, this PR involves
- refactoring `resolve_ltdetr_step_schedule` so that we have a separate module `ltdetr_schedule` to serve both scheduler and augmentation (I agree it is not optimal. We would really need to refactor the whole `object_detection_components/` later on.) 
- interface changes in `ModelArgs.resolve_auto()` so that we could pass the params from training loop
- relevant unit tests
- relevant docs

## How has it been tested?

- Unit tests

- [ ] Yes

- [x] Yes


src/lightly_train/_task_models/image_classification/train_model.py
```diff
@@ -72,6 +72,8 @@ class ImageClassificationTrainArgs(TrainModelArgs):
     def resolve_auto(
         self,
         total_steps: int,
+        gradient_accumulation_steps: int,
```
gabrielfruet, line 75: Not a thing you should address. Just wanted to understand what is the relation of gradient accum steps and train num batches. What exactly they are in our pipeline?

yutong-xiang-97: `train_num_batches` is the number of dataloader batches per epoch after the configured global/per-device batch size, distributed setup, and `drop_last` behavior are applied. 

`gradient_accumulation_steps` is how many of those dataloader batches we process before one optimizer step. See [here](https://intuitiveshorts.substack.com/p/short-7-what-is-gradient-accumulation) for more details on how it works.

So for step-based schedules, the effective optimizer steps per epoch are `train_num_batches` / `gradient_accumulation_steps`.

tests/_transforms/test_object_detection_transform.py
```diff
@@ -223,55 +223,68 @@ def _get_ltdetr_train_transform_args() -> DINOv3LTDETRObjectDetectionTrainTransf
...
-        (7200, 400, 4000, 6000),
-    ],
-)
-def test_resolve_ltdetr_step_schedule__resolved_windows(
-    total_steps: int,
-    expected_step_start: int,
-    expected_step_flat: int,
-    expected_step_stop: int,
-) -> None:
+def test_resolve_ltdetr_step_schedule_for_augmentation__applies_windows() -> None:
```
gabrielfruet: Why not parametrize here?

src/lightly_train/_commands/train_task.py
```diff
@@ -1332,14 +1332,44 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
             config.batch_size * config.gradient_accumulation_steps
         )
 
+        # TODO(Guarin, 07/25): Handle auto batch_size/num_workers.
+        # Build dataloaders without a collate function; we install the
+        # fully-resolved collate on the dataloader further below.
+        train_dataloader = helpers.get_train_dataloader(
```
gabrielfruet, line 1338: I'm not 100% sure but this may have facilitated this

https://linear.app/lightly/issue/TRN-2021/figure-out-a-first-class-way-to-pass-resolved-model-args-to-transforms

gabrielfruet: In the sense of being easier to resolve this problem that I had.

src/lightly_train/_task_models/object_detection_components/ltdetr_schedule.py
```diff
@@ -0,0 +1,111 @@
...
+    - ``step_start`` begins the scheduled LTDETR transition window.
+    - ``step_flat`` marks the boundary between the flat phase and the
+      no-augmentation tail.
+    - ``step_stop`` begins the no-augmentation tail.
+    """
+
+    steps_per_epoch = train_num_batches / gradient_accumulation_steps
+    total_epochs = total_steps / steps_per_epoch
+
+    no_aug_epochs_resolved = min(
```
gabrielfruet, line 82: I assume this is just to match the original implementation. Why we don't let the model have no_aug_epochs greater than `LTDETR_REFERENCE_NO_AUG_EPOCHS`?

gabrielfruet (approved):
Left some comments, but in general, everything looks good!

src/lightly_train/_task_models/object_detection_components/ltdetr_schedule.py
yutong-xiang-97: In the original paper, the logic of `no_aug_epochs` is tied to the model size. Anyway let's focus less on the parameter tuning.

PR #739, opened 2026-05-20
Fix failing ONNX exports

simonschoelly:
## What has changed and why?

This PR fixes the failing ONNX exports for the models
- `dinov3_eomt_instance_segmentation`
- `dinov3_eomt_panoptic_segmentation`
- `dinov3_eomt_semantic_segmentation`

There was an issue with shape inference, probably related to this issue https://github.com/onnx/onnx/issues/6664

Another option would to try the export with `dynamo=True`.

## How has it been tested?

Tests previously marked as `xfail` pass now.

## Did you update [CHANGELOG.md](../CHANGELOG.md)?

- [x] Yes
- [ ] Not needed (internal change)

## Did you update the documentation?

- [ ] Yes
- [x] Not needed (internal change without effects for user)

PR #752, opened 2026-05-27
Fix failing onnx export tests by switching to torch dynamo

simonschoelly:
## What has changed and why?

This PR fixes the failing ONNX exports for the models

* `dinov3_eomt_instance_segmentation`
* `dinov3_eomt_panoptic_segmentation`
* `dinov3_eomt_semantic_segmentation`

It does so, by using Dynamo instead of TorchScript for the export.

One downside is, that exporting `dinov3_eomt_semantic_segmentation` now only works for square images. The reason is, that Dynamo struggles to infer the correct shapes within the tiling algorithm. In a further issue we can see how we can rewrite the tiling mechanism so that it would work correctly with Dynamo.

This is an alternative approach to #739. 

## How has it been tested?

Tests marked previously as xfail pass now.

- [ ] Yes

- [ ] Yes


src/lightly_train/_task_models/dinov3_eomt_instance_segmentation/task_model.py
```diff
@@ -761,11 +761,12 @@ def export_onnx(
         width = self.image_size[1] if width is None else width
         num_channels = len(self.image_normalize["mean"])
 
-        # if the batch size is dynamic the dummy input must have multiple batches
-        # as torch might still think the batch size is one
         if dynamic_batch_size:
             batch_size = 2
-        dynamic_axes = {"images": {0: "N"}} if dynamic_batch_size else None
+            batch_dim = torch.export.Dim("batch_size", min=1, max=2**31 - 1)
```
liopeer, line 769: Nice, was not aware of the existence of this one!

tests/_task_models/dinov3_eomt_panoptic_segmentation/test_task_model.py
```diff
@@ -77,7 +76,70 @@ def test_export_onnx(model: DINOv3EoMTPanopticSegmentation, tmp_path: Path) -> N
     model.export_onnx(out=out, simplify=False, verify=True)
 
 
-@pytest.mark.xfail(strict=True, reason="dinov3 ONNX export shape mismatch")
+@pytest.mark.skipif(
+    not RequirementCache("torch>=2.8.0"), reason="requires PyTorch >= 2.8.0"
```
liopeer: Ah, so any torch dynamo export would ofc. require torch 2.8+. I think that's totally fine, since you can easily train in some version and upgrade for the export if needed (in case you have an old driver where newer torch versions are not supported anymore). We just need to make sure to properly document it in the docstring of the export function (also that only square works).

simonschoelly: No this is a leftover and should have been deleted, as I figured out that the issuse was not the torch version.

simonschoelly: But I think this test should only run on our internal machine, as it needs to download a model.

simonschoelly:
@liopeer I added a `@pytest.mark.long_running_test` marker to one of my tests. I am skipping  now tests if the have this variable set and we are running on the GITHUB CI - I feel this is the better way than only including this tests when an env variable is set as otherwise I user might locally forget to set such variable. But I don't have a too strong opinion if you want to do it another way.

src/lightly_train/_task_models/dinov3_eomt_instance_segmentation/task_model.py
simonschoelly: I am not sure if this is the best way though, I think there would also have been a way to use `dynamic_shapes` without specifying a min/max. But if you specify it this way, it is apparently very important to also specify max, as otherwise Dynamo was strugeling.

src/lightly_train/_task_models/dinov3_eomt_instance_segmentation/task_model.py
```diff
@@ -761,11 +762,19 @@ def export_onnx(
         width = self.image_size[1] if width is None else width
         num_channels = len(self.image_normalize["mean"])
 
-        # if the batch size is dynamic the dummy input must have multiple batches
-        # as torch might still think the batch size is one
+        torch_version = version.parse(torch.__version__.split("+", 1)[0])
```
liopeer: Would use `RequirementCache` on module level. If the message appears multiple times, then you can also have the message on module level.

tests/conftest.py
```diff
@@ -48,6 +50,23 @@ def lightly_train_cache_dir(
...
 
 
+def pytest_configure(config: pytest.Config) -> None:
+    config.addinivalue_line(
+        "markers",
+        "long_running_test: marks tests as long running (skipped on GitHub CI)",
+    )
+
+
+def pytest_collection_modifyitems(
```
gabrielfruet, line 60: I thought we already had something like this, but apparently not.

tests/_task_models/dinov3_eomt_semantic_segmentation/test_task_model.py
```diff
@@ -123,13 +120,18 @@ def test_export_onnx__static_batch_size(
     )
 
 
-@pytest.mark.xfail(strict=True, reason="dinov3 ONNX export shape mismatch")
+@pytest.mark.xfail(
+    reason="ONNX export does not support non-square images yet (requires tiling support).",
```
gabrielfruet, line 134: Nicer message

gabrielfruet: I mean, the message now is nicer :)

tests/_task_models/dinov3_eomt_instance_segmentation/test_task_model.py
```diff
@@ -82,30 +82,26 @@ def test_export_onnx__dynamic_batch_size(
...
-
-    inputs = np.random.randn(3, 3, 16, 16).astype(np.float32)
+    assert input_batch_dim.dim_param == "batch_size"
 
     session = ort.InferenceSession(str(out), providers=["CPUExecutionProvider"])
-    onnx_outputs = session.run(None, {"images": inputs})
 
-    with torch.no_grad():
-        torch_outputs = model(torch.from_numpy(inputs))
+    for batch_size in [1, 3]:
```
gabrielfruet: I think this should be a pytest.mark.parametrize instead, it makes the code easier to read

src/lightly_train/_task_models/dinov3_eomt_instance_segmentation/task_model.py
```diff
@@ -761,11 +762,19 @@ def export_onnx(
...
+        if torch_version < version.parse("2.5.0"):
```
gabrielfruet: We should use this standard for pinning versions 

https://github.com/lightly-ai/lightly-train/blob/9471af773027e6201e7a708619bee6172832c2c6/src/lightly_train/_task_models/picodet_object_detection/transforms.py#L32-L33

Make a module level constant like TORCH_GEQ_2_5 or TORCH_GREATHER_OR_EQUAL_2_5 and use RequirementCache. Then you can use this variable as your conditional logic to raise errors and etc. Not necessarily you need to extract this into a function as I used as example earlier in the Slack channel.

gabrielfruet (changes requested):

gabrielfruet (changes requested):
Requested some changes related to RequirementsCache and also testing. Everything else looks good.

tests/conftest.py
simonschoelly: No so far we only skipped some test on Github by checking if cuda was available.

src/lightly_train/_task_models/dinov3_eomt_instance_segmentation/task_model.py
liopeer: Thanks. I looked at the docs and the preferred way would probably indeed be to not specify `min` and `max`. But since it covers any int32 I am fine with leaving as is.

src/lightly_train/_task_models/dinov3_eomt_panoptic_segmentation/task_model.py
```diff
@@ -1084,12 +1085,12 @@ def export_onnx(
         input_names = ["images"]
         output_names = ["masks", "segment_ids", "scores"]
 
-        # Define dynamic axes.
-        dynamic_axes = {
-            "masks": {1: "num_segments"},
-            "segment_ids": {1: "num_segments"},
-            "scores": {1: "num_segments"},
-        }
+        torch_version = version.parse(torch.__version__.split("+", 1)[0])
```
liopeer: Since it appears in multiple modules, we can also create a new `src/lightly_train/_dependency.py` module and move it there.

src/lightly_train/_task_models/dinov3_eomt_semantic_segmentation/task_model.py
```diff
@@ -850,11 +860,27 @@ def export_onnx(
...
         width = self.image_size[1] if width is None else width
+        if height != width:
+            raise ValueError(
+                f"ONNX export requires square images (height == width), got "
+                f"height={height}, width={width}. Tiling for non-square images "
+                f"is not supported during ONNX export."
+            )
         num_channels = len(self.image_normalize["mean"])
 
+        torch_version = version.parse(torch.__version__.split("+", 1)[0])
```
liopeer: replace then here as well

liopeer (approved):

tests/_task_models/dinov3_eomt_instance_segmentation/test_task_model.py
simonschoelly: Done

src/lightly_train/_task_models/dinov3_eomt_instance_segmentation/task_model.py
simonschoelly: Done

gabrielfruet (approved):
LGTM

PR #762, opened 2026-06-02
Add Depth Estimation (DAv3) to LightlyTrain

mrpositron:
[updated 05.06.26]

## What has changed and why?

This is the first step toward depth estimation in LightlyTrain: inference for Depth Anything v3 (DAv3) added to `_task_models`.

I created the folder `dinov2_dav3_relative_depth_estimation` to match the existing naming convention (both DAv2 and DAv3 build on a DINOv2 backbone).

## Design Choices
- Reuses the existing DINOv2 backbone. Rather than reimplementing the ViT, the backbone is constructed through `DINOV2_VIT_PACKAGE.get_model(...)` - the same path every other task model (object detection, segmentation) uses. The backbone is configured to faithfully reproduce the plain ViT-L that DAv3 is built on, so its state-dict keys match the official checkpoint.

- Loads from a converted LightlyTrain checkpoint, not the original HF weights. Following the existing convention, the model is served from a single pre-converted `.pt` via `lightly_train.load_model(...)`:
```python
model = lightly_train.load_model(".../ckpt/dav3_relative_large.pt")
prediction = model.predict(".../example.jpg")
```
- `convert_checkpoint.py` is a **temporary file**. It downloads the official DAv3 HF weights, remaps their keys onto the LightlyTrain layout, and re-exports them as a .pt. It is the single place that understands the official DAv3 weight format. **It will be removed in the final PR**, once we host the converted checkpoints ourselves.

## How has it been tested?

Run sample image with [official implementation](https://github.com/bytedance-seed/depth-anything-3) and and current implementation, and compared them side-by-side. Output is close to each other.

Some stats for this image:
```python
diff = np.abs(np_gt - np_pred)

print(f"Stats for GT: min={np_gt.min():.4f}, max={np_gt.max():.4f}, mean={np_gt.mean():.4f}, std={np_gt.std():.4f}")
print(f"Stats for Pred: min={np_pred.min():.4f}, max={np_pred.max():.4f}, mean={np_pred.mean():.4f}, std={np_pred.std():.4f}")
print(f"Difference stats: min={diff.min():.4f}, max={diff.max():.4f}, mean={diff.mean():.8f}, std={diff.std():.8f}")
```

```bash
Stats for GT: min=0.1932, max=2.7362, mean=1.1163, std=0.8576
Stats for Pred: min=0.1932, max=2.7363, mean=1.1163, std=0.8576
Difference stats: min=0.0000, max=0.0197, mean=0.00002760, std=0.00014143
```

image

- [ ] Yes

- [ ] Yes


src/lightly_train/_task_models/dinov2_dav3_depth_estimation/__init__.py
```diff
@@ -0,0 +1,12 @@
+#
+# Copyright (c) Lightly AG and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+#
+from lightly_train._task_models.dinov2_dav3_depth_estimation.task_model import (
```
liopeer: We usually leave `__init__` files empty, unless we want to add them to the public API. No user should ever directly use the `DepthAnythingV3MonocularDepthEstimation` class, but instead instantiate the model through `lightly_train.load_model`.

src/lightly_train/_task_models/dinov2_dav3_depth_estimation/task_model.py
```diff
@@ -0,0 +1,303 @@
...
+    }
+}
+
+_MODEL_ALIASES: dict[str, str] = {
+    "da3mono-large": "da3mono-large",
+    "depth-anything-v3/da3mono-large": "da3mono-large",
+    "depth-anything/da3mono-large": "da3mono-large",
+}
+
+_DEFAULT_IMAGE_NORMALIZE = {
```
liopeer: Make sure to then re-use this in the training transform, i.e. add `TODO`.

src/lightly_train/_task_models/dinov2_dav3_depth_estimation/task_model.py
```diff
@@ -0,0 +1,303 @@
...
+    "mean": (0.485, 0.456, 0.406),
+    "std": (0.229, 0.224, 0.225),
+}
+
+
+class DepthAnythingV3MonocularDepthEstimation(TaskModel):
+    """Depth Anything V3 monocular relative-depth inference model."""
+
+    model_suffix = "depth-estimation"
```
liopeer: `"depth-estimation"` is the task, but should not be the model name. The model name should probably be `"depthanythingv3"` or `"dav3"`. Best to also avoid `"-"` in this model name. For OD we have `"ltdetr"` and for others we simply have `"eomt"`.

src/lightly_train/_task_models/dinov2_dav3_depth_estimation/task_model.py
```diff
@@ -0,0 +1,303 @@
...
+
+    def __init__(
+        self,
+        *,
+        model_name: str = "depth-anything-v3/da3mono-large",
+        image_size: int = 504,
+        image_normalize: dict[str, tuple[float, ...]] | None = None,
```
liopeer: Would not use defaults for these 3, since they are always readily available anyway.

src/lightly_train/_task_models/dinov2_dav3_relative_depth_estimation/task_model.py
```diff
@@ -0,0 +1,303 @@
...
+                ``drop_path_rate``. These override the Depth Anything V3 defaults.
+            load_weights:
+                If True, the backbone is initialized from the pretrained DINOv2
+                ViT-L/14 weights that Depth Anything V3 was fine-tuned from. The DPT
+                head is always randomly initialized. To obtain the fully pretrained
+                depth model, use ``lightly_train.load_model(<exported .pt>)``; the
+                ``convert_checkpoint`` script produces that file from the official
+                Depth Anything V3 weights. Set to False to skip the backbone download,
+                e.g. when the weights are loaded from an exported checkpoint via
+                ``load_train_state_dict``.
```
liopeer: Will this behavior change after we discard the conversion script? Currently, this seems a bit counterintuitive. I would assume: `load_weights=True` loads DAv3 weights, and `load_weights=False` uses randomly initialized head and (probably) the pretrained DINOv2 checkpoint from Meta (since this is the default from the DINOv2_PACKAGE).

src/lightly_train/_task_models/dinov2_dav3_depth_estimation/task_model.py
```diff
@@ -0,0 +1,303 @@
...
+        model_args: dict[str, Any] | None = None,
+        backbone_args: dict[str, Any] | None = None,
+        load_weights: bool = True,
+    ) -> None:
+        """
+        Args:
+            model_name:
+                The Depth Anything V3 model name. Supported names are
+                ``"depth-anything-v3/da3mono-large"``, ``"depth-anything/DA3MONO-LARGE"``,
+                and ``"da3mono-large"``.
```
liopeer: I would aim for a single name to start with. We can still later introduce aliases to make things easier for the users. Question is, do we want to keep the keep the naming scheme of other models, and call this one `"dinov2/depthanythingv3-mono-large"` or something like this?

src/lightly_train/_task_models/dinov2_dav3_depth_estimation/task_model.py
```diff
@@ -0,0 +1,303 @@
...
+class DepthAnythingV3MonocularDepthEstimation(TaskModel):
+    """Depth Anything V3 monocular relative-depth inference model."""
+
+    model_suffix = "depth-estimation"
+
+    def __init__(
+        self,
+        *,
+        model_name: str = "depth-anything-v3/da3mono-large",
+        image_size: int = 504,
```
liopeer: In the other models I think we use tuples for this. We should keep it consistent if possible. It's probably also important for batch prediction to have equally-sized images across both dimensions.

src/lightly_train/_task_models/dinov2_dav3_relative_depth_estimation/task_model.py
```diff
@@ -0,0 +1,303 @@
...
+        self.model_name = config["canonical_name"]
+        self.image_size = image_size
+        self.image_normalize = (
+            _DEFAULT_IMAGE_NORMALIZE if image_normalize is None else image_normalize
+        )
+
+        # Reuse the official Depth Anything V3 input pipeline so that resizing and
+        # normalization match the reference implementation exactly. The mono model
+        # bounds the longest side and rounds both sides to a patch multiple, then
+        # applies the official ImageNet normalization.
```
liopeer: If it uses ImageNet anyway, then why have the normalization separate in self.image_normalize?

src/lightly_train/_task_models/dinov2_dav3_depth_estimation/task_model.py
```diff
@@ -0,0 +1,303 @@
...
+            model_args=backbone_model_args,
+            load_weights=load_weights,
+        )
+        self.decoder = DPT(
+            dim_in=int(self.backbone.embed_dim),
+            patch_size=patch_size,
+            output_dim=int(net_args.get("output_dim", 1)),
+            features=int(net_args.get("features", 256)),
+            out_channels=tuple(net_args.get("out_channels", (256, 512, 1024, 1024))),
+            use_sky_head=bool(net_args.get("use_sky_head", True)),
```
liopeer: Don't use `.get` here, but instead access the key directly. Now we basically have "default" values in 2 places.

src/lightly_train/_task_models/dinov2_dav3_relative_depth_estimation/task_model.py
```diff
@@ -0,0 +1,303 @@
...
+            to a multiple of the patch size). This matches the official ``Prediction``
+            resolution. Larger values correspond to farther scene content.
+        """
+        self._track_inference()
+        if self.training:
+            self.eval()
+
+        x = self._preprocess_image(image)
+        out = self.forward(x.unsqueeze(0))
+        return out["depth"][0, 0]
```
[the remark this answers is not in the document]
liopeer: Probably agree with this.

src/lightly_train/_task_models/dinov2_dav3_depth_estimation/task_model.py
```diff
@@ -0,0 +1,303 @@
...
+
+    def forward(self, x: Tensor) -> dict[str, Tensor]:
+        """Run depth inference on a preprocessed batch with shape ``(B, 3, H, W)``."""
+        if x.ndim != 4:
+            raise ValueError(
+                f"Expected input shape (B, C, H, W), got {tuple(x.shape)}."
+            )
+        feats = self._extract_features(x)
+        out = self.decoder(feats=feats, H=x.shape[-2], W=x.shape[-1])
+        _set_mono_sky_regions_to_max_depth(out)
```
liopeer: This sounds like postprocessing to me. Do we want this in the `forward` pass or only in the `predict` method?

src/lightly_train/_task_models/dinov2_dav3_relative_depth_estimation/task_model.py
```diff
@@ -0,0 +1,303 @@
...
+        intermediate = self.backbone.get_intermediate_layers(
+            x,
+            n=self.out_layers,
+            reshape=False,
+            return_class_token=True,
+            norm=True,
+        )
+        return [patch_tokens for patch_tokens, _class_token in intermediate]
+
+    def _preprocess_image(self, image: PathLike | PILImage | Tensor) -> Tensor:
```
liopeer, line 228: Did you try using torchvision transforms instead? Does it not work? I would be very surprised if it didn't. Already now we're using albumentations (cv2) for training and torchvision for inference (in the other models).

src/lightly_train/_task_models/dinov2_dav3_depth_estimation/task_model.py
```diff
@@ -0,0 +1,303 @@
...
+            [proc_input],
+            process_res=self.image_size,
+            process_res_method=self.process_res_method,
+            num_workers=1,
+        )
+        first_param = next(self.parameters())
+        return batch[0].to(device=first_param.device, dtype=first_param.dtype)
+
+
+def _set_mono_sky_regions_to_max_depth(out: dict[str, Tensor]) -> None:
```
liopeer: This is definitely postprocessing, so it should not be inside `forward`.

src/lightly_train/_task_models/dinov2_dav3_depth_estimation/task_model.py
```diff
@@ -0,0 +1,303 @@
...
+        )
+        non_sky_depth = non_sky_depth[idx]
+    non_sky_max = torch.quantile(non_sky_depth, 0.99)
+
+    depth = out["depth"].clone()
+    depth[~non_sky_mask] = non_sky_max
+    out["depth"] = depth
+
+
+def _as_input_processor_image(image: PathLike | PILImage | Tensor) -> np.ndarray:
```
liopeer: prefer typing as `NDArray[<np.dtype>]`

src/lightly_train/_task_models/dinov2_dav3_relative_depth_estimation/task_model.py
mrpositron: My goal was to mirror the existing implementation as faithfully as possible, minimizing the output diff. Introducing different preprocessing is possible, but it will likely introduce some deviations in the diff.

src/lightly_train/_task_models/depth_estimation_components/head_utils.py
```diff
@@ -0,0 +1,168 @@
...
+    align_corners: bool = True,
+) -> Tensor:
+    """
+    Safe interpolation implementation to avoid INT_MAX overflow in torch.nn.functional.interpolate.
+    """
+    if size is None:
+        assert scale_factor is not None, "Either size or scale_factor must be provided."
+        size = (int(x.shape[-2] * scale_factor), int(x.shape[-1] * scale_factor))
+
+    INT_MAX = 1610612736
```
liopeer, line 157: Should probably use this in `linear_segmentation` as well. There we run into exactly this problem when we try to go for batch sizes > 16. 🤔

src/lightly_train/_task_models/depth_estimation_components/input_processor.py
```diff
@@ -0,0 +1,523 @@
...
+      2) Boundary resize (upper/lower bound, preserving aspect ratio)
+      3) Enforce divisibility by PATCH_SIZE:
+         - "*resize" methods: each dimension is rounded to nearest multiple
+           (may up/downscale a few px)
+         - "*crop"   methods: each dimension is floored to nearest multiple via center crop
+      4) Convert to tensor and apply ImageNet normalization
+      5) Stack into (1, N, 3, H, W)
+
+    Parallelization:
+      - Each image is processed independently in a worker.
```
liopeer: This is a bit of a problem. Preprocessing should not spawn workers, that should be done by a dataloader which wraps a torch dataset. But no need to address in this PR.

src/lightly_train/_task_models/depth_estimation_components/input_processor.py
```diff
@@ -0,0 +1,523 @@
...
+        *,
+        num_workers: int = 8,
+        print_progress: bool = False,
+        sequential: bool | None = None,
+        desc: str | None = "Preprocess",
+    ) -> tuple[torch.Tensor, torch.Tensor | None, torch.Tensor | None]:
+        """
+        Returns:
+            (tensor, extrinsics_list, intrinsics_list)
+            tensor shape: (1, N, 3, H, W)
```
liopeer, line 66: Not entirely clear to me what it returns if extrinsics/intrinsics are not provided. But those are anyway only required for metric depth I assume.

src/lightly_train/_task_models/depth_estimation_components/input_processor.py
```diff
@@ -0,0 +1,523 @@
...
+        arr = cv2.resize(np.asarray(img), (new_w, new_h), interpolation=interpolation)
+        return Image.fromarray(arr)
+
+
+# Backward compatibility alias
+InputAdapter = InputProcessor
+
+
+# ===========================
+# Minimal test runner (parallel execution)
```
liopeer: do we need this?

src/lightly_train/_task_models/depth_estimation_components/parallel_utils.py
```diff
@@ -0,0 +1,105 @@
+#
```
liopeer: I would try to get rid of this entirely (and hardcode `num_workers=0`).

liopeer (commented):
I think the main issue I currently see is the separation of preprocessing, model forward pass and postprocessing, which we'll need to address in order to fit the standards of the other task models a bit better. But the rest is looking really nice!

src/lightly_train/_task_models/dinov2_dav3_depth_estimation/task_model.py
mrpositron: `image_size: int = 504` is a longest-side upper bound. The DA3 InputProcessor (`upper_bound_resize`) scales the longest side to 504, scales the other side by the same factor, and rounds both to a patch multiple, no padding. Output (H, W) therefore varies with aspect ratio.

mrpositron: I will use better name for it.

src/lightly_train/_task_models/dinov2_dav3_relative_depth_estimation/task_model.py
mrpositron: Agreed 👍. I am looking into it.

src/lightly_train/_task_models/dinov2_dav3_relative_depth_estimation/dpt.py
```diff
@@ -0,0 +1,501 @@
...
+                *ln_seq,
+                nn.ReLU(inplace=True),
+                nn.Conv2d(head_features_2, 1, kernel_size=1, stride=1, padding=0),
+            )
+
+    # -------------------------------------------------------------------------
+    # Public forward
+    # -------------------------------------------------------------------------
+    def forward(
+        self, feats: list[torch.Tensor], H: int, W: int
```
liopeer, line 186: In follow-up: import classes directly, i.e. use `Module` instead of `nn.Module`, `Tensor` instead of `torch.Tensor`.

dev_tools/depth_anything_3_licenseheader.tmpl
```diff
@@ -0,0 +1,13 @@
+Copyright (c) 2025 ByteDance Ltd. and/or its affiliates
```
liopeer, line 1: Are you planning on using it inside the `depth_estimation_components` sub-package? Because for this you would need to change the Makefile.

pyproject.toml
```diff
@@ -190,6 +190,7 @@ ignore_missing_imports = true
 module = [
     "lightly_train._models.dinov2_vit.dinov2_vit_src.*",
     "lightly_train._models.dinov3.dinov3_src.*",
+    "lightly_train._task_models.depth_estimation_components.*",
```
liopeer, line 194: If we're not treating it as "vendored code", i.e. copied code from the repo that we use with minimal changes, we should eventually make it fully typed.

src/lightly_train/_task_models/depth_estimation_components/input_processor.py
```diff
@@ -0,0 +1,367 @@
...
+        self,
+        *,
+        image: list[np.ndarray | Image.Image | str],
+        exts_list: list[np.ndarray | None] | None,
+        ixts_list: list[np.ndarray | None] | None,
+        process_res: int,
+        process_res_method: str,
+    ):
+        results = [
+            self._process_one(
```
liopeer, line 125: Eventually we will want to split this into "batchable" and "non-batchable" transforms as well.

src/lightly_train/_task_models/dinov2_dav3_relative_depth_estimation/task_model.py
```diff
@@ -0,0 +1,328 @@
...
+    DPT,
+)
+from lightly_train._task_models.task_model import TaskModel
+from lightly_train.types import PathLike
+
+logger = logging.getLogger(__name__)
+
+_MODEL_CONFIGS: dict[str, dict[str, Any]] = {
+    "dinov2/dav3-relative-large": {
+        "canonical_name": "dinov2/dav3-relative-large",
```
liopeer, line 38: This looks a bit redundant, but no urgent need to address.

src/lightly_train/_task_models/dinov2_dav3_relative_depth_estimation/task_model.py
```diff
@@ -0,0 +1,328 @@
...
+            "use_sky_head": True,
+        },
+    }
+}
+
+
+class DepthAnythingV3RelativeDepthEstimation(TaskModel):
+    """Depth Anything V3 relative-depth inference model."""
+
+    model_suffix = "dav3_relative_large"
```
liopeer, line 59: Once we have metric depth postprocessors, we'll need to rename this.

liopeer (approved):
I think it would be mergeable. You can treat all my comments as suggestions that should flow into follow-up issues.

src/lightly_train/_task_models/dinov2_dav3_relative_depth_estimation/task_model.py
mrpositron: My idea is to name it `dav3_metric_large`

pyproject.toml
mrpositron: Agreed 👍. Was not sure about it.

PR #764, opened 2026-06-03
Benchmark command for object detection

simonschoelly:
## What has changed and why?

This PR adds a `benchmark_object_detection` command that allow one to benchmark speed and accuracy of the object detection task models with PyTorch, ONNX and TensorRT.

Current output:
```
                                                                        Benchmark Report — eval_run0                                                                         

Run Config                                                                                                                                                                   

 • Model: dinov3/vitt16-ltdetr                                                                                                                                               
 • Model class: DINOv3LTDETRObjectDetection                                                                                                                                  
 • Backend: onnx (cuda), fp16                                                                                                                                                
 • Dataset: COCO 2017 (5000/5000 images)                                                                                                                                     
 • Batch Size: 8                                                                                                                                                             
 • Warmup Steps: 0                                                                                                                                                           
 • Steps: all                                                                                                                                                                

Device Info                                                                                                                                                                  

 • GPU: NVIDIA GeForce RTX 3090                                                                                                                                              
 • GPU Memory: 23.6 GB                                                                                                                                                       
 • CUDA: 12.8                                                                                                                                                                
 • cuDNN: 91002                                                                                                                                                              
 • CPU Threads: 16                                                                                                                                                           
 • RAM: 62.7 GB                                                                                                                                                              

Performance Metrics                                                                                                                                                          

                      
 Metric         Value 
 ──────────────────── 
 mAP@0.5:0.95  0.4906 
 mAP@0.50      0.6697 
 mAP@0.75      0.5265 
 mAP (small)   0.2317 
 mAP (medium)  0.4628 
 mAP (large)   0.6607 
 mAR@1         0.3765 
 mAR@10        0.6275 
 mAR@100       0.6915 
 mAR (small)   0.4599 
 mAR (medium)  0.6884 
 mAR (large)   0.8364 
                      

Throughput & Latency                                                                                                                                                         

                                                             
                          min     max    mean  median    std 
 ─────────────────────────────────────────────────────────── 
 Throughput (img/s)     13.74   45.13   43.23   43.98   2.60 
 Throughput (batch/s)    1.72    5.64    5.40    5.50   0.33 
 Latency (ms/img)       22.16   72.80   23.28   22.74   2.59 
 Latency (ms/batch)    177.25  582.37  186.22  181.92  20.73 
                                                                                                                          
```

## How has it been tested?

Some tests were written and I also run some benchmarks manually.

Would wait with update until the feature is mature enough.

- [ ] Yes

Would wait with update until the feature is mature enough.

- [ ] Yes


src/lightly_train/_commands/benchmark_backends.py
```diff
@@ -0,0 +1,357 @@
...
+
+        if hasattr(self.model, "deploy"):
+            self.model.deploy()  # type: ignore[operator]
+        else:
+            self.model.eval()
+
+        if backend_args.compile:
+            self.model.forward_backend = torch.compile(self.model.forward_backend)  # type: ignore[method-assign]
+
+    @override
```
liopeer, line 92: Nit: I don't think you need the override. Also you can just use `...` instead of `pass` in above abstractmethod.

uv.lock
```diff
@@ -386,9 +386,6 @@ conflicts = [[
     { package = "lightly-train", group = "pinned-torch-minimal" },
 ]]
 
-[options]
```
liopeer, line 389: Can you generate it with `make lock`, so that we keep the `--exclude-newer`? You're also free to upgrade the date in the Makefile, but we should always lock with an `--exclude-newer` date.

src/lightly_train/_commands/benchmark_types.py
```diff
@@ -0,0 +1,322 @@
...
+    COCOObjectDetectionDataArgs,
+)
+from lightly_train._data.yolo_object_detection_dataset import (
+    YOLOObjectDetectionDataArgs,
+)
+from lightly_train._task_models.task_model import TaskModel
+from lightly_train.types import PathLike
+
+
+class ObjectDetectionPrediction(TypedDict):
```
liopeer, line 31: Do we not have this anywhere already?

src/lightly_train/_commands/benchmark_types.py
```diff
@@ -0,0 +1,322 @@
...
+    """Result of a benchmark run."""
+
+    out: str
+    model_name: str | None
+    model_class: str
+    backend_args: BenchmarkBackendArgs
+    device_info: Annotated[DeviceInfo, Field(discriminator="device_type")]
+    dataset_name: str
+    dataset_size: int
+    num_images: int
```
liopeer, line 103: What is the difference between those 2?

src/lightly_train/_commands/benchmark_types.py
```diff
@@ -0,0 +1,322 @@
...
+
+        # Footer.
+        results_path = Path(self.out) / "benchmark_results.json"
+        lines.append("---")
+        lines.append(f"Full report: `{results_path}`")
+
+        return "\n".join(lines)
+
+    def print(self) -> None:
+        from rich.console import Console
```
liopeer: If we're adding `rich` to core dependencies, prefer module-level import.

src/lightly_train/_commands/benchmark_types.py
```diff
@@ -0,0 +1,322 @@
...
+        console = Console()
+        console.print(Markdown(self.to_markdown()))
+
+
+class TorchBackendArgs(PydanticConfig):
+    """Backend arguments for PyTorch inference."""
+
+    format: Literal["torch"] = "torch"
+    compile: bool = False
+    precision: Literal["fp32", "fp16", "bf16"] = "fp32"
```
liopeer: I think here we should use the same terminology as in the other commands. See e.g. the `train_task` command. And in the current implementation we use anyway `torch.amp.autocast`, which is not really `bf16` or `fp16`, but instead would be `bf16-mixed` or `fp16-mixed`.

src/lightly_train/_commands/benchmark_types.py
```diff
@@ -0,0 +1,322 @@
...
+class BenchmarkObjectDetectionConfig(PydanticConfig):
+    out: PathLike
+    dataset_name: str
+    data: Annotated[
+        Union[YOLOObjectDetectionDataArgs, COCOObjectDetectionDataArgs],
+        Field(discriminator="format"),
+    ]
+    model: TaskModel | PathLike
+    batch_size: int = Field(default=1, ge=1)
+    threshold: float = 0.0
```
liopeer: Shouldn't we use a higher value? Or are the values that you report with threshold 0.0?

src/lightly_train/_task_models/picodet_object_detection/task_model.py
```diff
@@ -468,6 +473,69 @@ def forward(
...
+            ``scores`` keys.
+        """
+        boxes_xyxy, obj_logits, cls_logits = raw_outputs
+        model_h, model_w = self.image_size
+
+        scores = torch.sigmoid(obj_logits)
+        internal_labels = cls_logits.argmax(dim=-1)
+        labels = self.internal_class_to_class[internal_labels]
+
+        out: list[dict[str, Tensor]] = []
```
liopeer, line 519: Note: Should eventually refactor to use a shared postprocessor class between picodet and LT-DETR.

src/lightly_train/_commands/benchmark_task.py
```diff
@@ -0,0 +1,595 @@
...
+
+
+def benchmark_object_detection(
+    *,
+    out: PathLike,
+    dataset_name: str,
+    data: dict[str, Any] | str,
+    model: TaskModel | PathLike,
+    batch_size: int = 1,
+    threshold: float = 0.0,
```
liopeer, line 84: Would only have defaults in one place. Now we have defaults here and in the pydantic class.

src/lightly_train/_commands/benchmark_task.py
```diff
@@ -0,0 +1,595 @@
...
+    args = {k: v for k, v in locals().items() if v is not None}
+    config = validate.pydantic_model_validate(BenchmarkObjectDetectionConfig, args)
+    return _benchmark_object_detection_from_config(config=config)
+
+
+def _benchmark_object_detection_from_config(
+    config: BenchmarkObjectDetectionConfig,
+) -> BenchmarkResult:
+    # Set up output directory.
+    out_dir = _get_out_dir(out=config.out, overwrite=config.overwrite)
```
liopeer, line 146: Can't we re-use this from other tasks?

src/lightly_train/_commands/benchmark_task.py
```diff
@@ -0,0 +1,595 @@
...
+    model: TaskModel
+    if isinstance(config.model, TaskModel):
+        model = config.model
+    else:
+        model = task_model_helpers.load_model(model=config.model)
+
+    backend_args = config.backend_args
+
+    # Build val transform args from the model.
+    transform_args = _build_val_transform_args(model=model)
```
liopeer, line 158: Consider using/adapting `get_transform_args` from `train_task_helpers.py` instead of completely reconstructing everything.

src/lightly_train/_commands/benchmark_task.py
```diff
@@ -0,0 +1,595 @@
...
+
+    # Set up validation data.
+    data_args = config.data
+    num_workers = common_helpers.get_num_workers(
+        num_workers=config.num_workers, num_devices_per_node=1
+    )
+    val_dataloader = _create_val_dataloader(
```
liopeer, line 165: Here we maybe can also re-use something existing.

src/lightly_train/_commands/benchmark_backends.py
```diff
@@ -0,0 +1,357 @@
...
+        )
+        time_predict = time.perf_counter() - start_predict
+
+        # postprocess
+        outputs = dict(zip(self.output_names, raw_outputs))
+        labels = torch.from_numpy(outputs["labels"])
+        boxes_unscaled = torch.from_numpy(outputs["boxes"])
+        scores = torch.from_numpy(outputs["scores"])
+
+        # The ONNX forward() rescales boxes to the model input size when
```
liopeer, line 237: We should definitely use a postprocessor class here.

src/lightly_train/_commands/benchmark_backends.py
```diff
@@ -0,0 +1,357 @@
...
+        self.stream.synchronize()
+        time_predict = time.perf_counter() - start_predict
+
+        # Postprocess.
+        labels_batch = outputs["labels"].cpu()
+        boxes_batch = outputs["boxes"].cpu()
+        scores_batch = outputs["scores"].cpu()
+
+        results: list[ObjectDetectionPrediction] = []
+        for i in range(len(metadata)):
```
liopeer: Here as well!

src/lightly_train/_commands/benchmark_task.py
```diff
@@ -0,0 +1,595 @@
...
+    summary_path.write_text(result.to_markdown() + "\n", encoding="utf-8")
+
+    result.print()
+
+    return result
+
+
+def _make_val_bbox_params() -> BboxParams:
+    """Create standard YOLO bbox params matching training val transforms."""
+    return BboxParams(
```
liopeer, line 332: If we re-use from the existing ValTransformArgs we don't need to re-define this.

uv.lock
simonschoelly: This is weird, I could have sworn that I used `make lock`. Did it now again and the `--exclude-newer` is now there.

src/lightly_train/_commands/benchmark_types.py
simonschoelly: The issue is that in the training loop we don't seem to use any threshold. So in order to get the same mAP I set it here to 0.0 by default. Otherwise it seems to be 0.6 in our predict methods. Not for the ONNX methods though were we don't have a threshold.

src/lightly_train/_commands/benchmark_task.py
simonschoelly: The issue here is that similar methods in the training loop use Fabric.

src/lightly_train/_commands/benchmark_task.py
simonschoelly: The issue is currently that `get_transform_args` works on `TrainClass`.

uv.lock
liopeer: Sometimes the agents run `uv run` without `--frozen` and then it generates a new lock. Might have been the issue.

src/lightly_train/_commands/benchmark_backends.py
simonschoelly: I dont't have an opinion on that. Do you prefer `...`?

src/lightly_train/_commands/benchmark_types.py
simonschoelly: No we don't. Probably should though. For example `postprocess` for `DINOv3LTDETRObjectDetection` returns `list[dict[str, Tensor]]` instead of something like `list[ObjectDetectionPrediction]`.

There is also a bit of inconsistencies if it is called `boxes` or `bboxes`.

Should I try to refactor it in this PR or should we do it somewhere else?

src/lightly_train/_commands/benchmark_types.py
simonschoelly: Yes I can do that. But I saw from Lightning we have this:
```
_PRECISION_INPUT_STR_ALIAS_CONVERSION = {"64": "64-true", "32": "32-true", "16": "16-mixed", "bf16": "bf16-mixed"}
```

src/lightly_train/_commands/benchmark_backends.py
simonschoelly: Can you elaborate what you exactly mean here?

src/lightly_train/_commands/benchmark_task.py
simonschoelly: See above why we cannot use ValTransformArgs.

src/lightly_train/_commands/benchmark_types.py
simonschoelly: The difference is that `dataset_size` is the number of images in the dataset. And `num_images` is the number of images that we actually use during the benchmark.
For the benchmark command we have
* `drop_last=True`. This is in difference to the valiation loop - but we have to make sure the batches always have the same size for the ONNX models.
* If `steps` is set, then `num_images = num_batches * num_images`.

Do you think the names are confusing and I should come up with better ones?

src/lightly_train/_commands/benchmark_task.py
simonschoelly: We seem to have default in both places for things like `TrainTaskConfig`. What would you prefer?

Thinking about it, maybe in the future we can just use a library that generates the functions from the Pydantic model.

src/lightly_train/_commands/benchmark_task.py
simonschoelly: The closest one is `common_helpers.get_out_dir`. But it made for the training loop, which is reflected in the error message and in checking the rank.

src/lightly_train/_commands/benchmark_backends.py
liopeer: Modern Python generally uses `...` afaik, but I don't care much. :) Also the `@override` comment is absolutely optional, which is why I prepended `Nit`.

src/lightly_train/_commands/benchmark_types.py
liopeer: No, not in this one, but we urgently should. 😅

src/lightly_train/_commands/benchmark_types.py
liopeer: Makes sense. I currently don't have a better idea either, so I would suggest leaving it.

src/lightly_train/_commands/benchmark_types.py
liopeer: That's really misleading imo. We should only use `_PRECISION_INPUT` from lightning. But not all values (unless we actually support them). If we only support mixed precision or full precision, we should only manually specify those, i.e. fp32, fp16-mixed and bf16-mixed. I would check all out train commands and use the same terminology.

src/lightly_train/_commands/benchmark_types.py
liopeer: Fair enough, then let's leave it.

src/lightly_train/_commands/benchmark_task.py
liopeer: Then that's really badly done in `TrainTaskConfig` if you ask me. 🙃 I would generally prefer to only have a single place where defaults are preferred.

src/lightly_train/_commands/benchmark_backends.py
liopeer: We currently have the "rescaling" of the bboxes implemented multiple times I think. Put it into a helper function. Eventually we will refactor this into a proper class for postprocessing that can be used across the whole codebase.

src/lightly_train/_commands/benchmark_backends.py
liopeer: This is the other location I mean.

src/lightly_train/_commands/benchmark_backends.py
simonschoelly: Done

src/lightly_train/_commands/benchmark_types.py
simonschoelly: Done

src/lightly_train/_commands/benchmark_task.py
simonschoelly: Did put the defaults now to `benchmark_object_detection`.

src/lightly_train/_commands/benchmark_backends.py
```diff
@@ -0,0 +1,372 @@
...
+        # Create CUDA stream for async execution on the target device.
+        self.stream = torch.cuda.Stream(device=self.device)  # type: ignore[no-untyped-call]
+
+        self.trt = trt
+
+    @override
+    def run_batch(
+        self, batch: ObjectDetectionBatch
+    ) -> tuple[list[ObjectDetectionPrediction], float]:
+        import numpy as np
```
liopeer, line 324: Can be moved to top of module in a later cleanup.

src/lightly_train/_commands/benchmark_types.py
```diff
@@ -0,0 +1,321 @@
...
+    def print(self) -> None:
+        console = Console()
+        console.print(Markdown(self.to_markdown()))
+
+
+class TorchBackendArgs(PydanticConfig):
+    """Backend arguments for PyTorch inference."""
+
+    format: Literal["torch"] = "torch"
+    compile: bool = False
```
liopeer, line 241: Is this deliberately on `False`?

liopeer (approved):
Let's merge it!

src/lightly_train/_commands/benchmark_types.py
simonschoelly: I don't remember why I set it to `False`. Perhaps it makes sense as it is slightly more stable?

PR #773, opened 2026-06-11
Add tiling utilities for SAHI support for instance segmentation

yutong-xiang-97:
## What has changed and why?

This is part 1 for adding SAHI to instance segmentation. It includes 
- extending the tiling utilities (`tiling_utils.py`) toward SAHI-style inference for EoMT instance segmentation
- fixing a correctness bug in how tile predictions are combined with the global-view prediction. A tile overlapping a different-label global most strongly — but also overlapping an above-threshold same-label global — was incorrectly kept, leaking a duplicate detection.
- the corresponding unit tests

More specifically,
- in `tile_image`:
  - added a `padding_mode` (`"resize"` default, preserving prior behavior, or `"pad"`) for instance segmentation.
  - rewrote the tile-stepping loop around an explicit `_tile_starts` helper that snaps the final tile to `size - tile_size`, eliminating the duplicate edge tiles the old `range(...)` + clamp produced
- Renamed `combine_predictions_tiles_and_global` to `combine_object_detection_tiles`
- Added `combine_instance_segmentation_tiles` (with mask-based `_mask_iou` and class-aware mask NMS helpers) for stitching tiled instance-segmentation masks.

## How has it been tested?

- Unit tests.

- [ ] Yes

- [ ] Yes


gabrielfruet (commented):
@codex review

tests/_task_models/object_detection_components/test_tiling_utils.py
```diff
@@ -0,0 +1,585 @@
...
+    assert tiles.shape == (1, 3, 16, 16)
+    torch.testing.assert_close(coordinates, torch.tensor([[0, 0]], device=image.device))
+    torch.testing.assert_close(tiles[0, :, :8, :10], image)
+    assert torch.all(tiles[0, :, 8:, :] == 0)
+    assert torch.all(tiles[0, :, :, 10:] == 0)
+
+
+def test_tile_image__appends_last_tile_for_non_divisible_size() -> None:
+    # 30 is not a multiple of the tile size, so the last tile in each dimension
+    # must be snapped back to size - tile_size (14) to stay within the image.
```
gabrielfruet, line 91: My impression is that, by this current test, the first tile is snapped back, since the second one starts at 14.

i.e. 

first one start at 0 ends at 14,        size **14**.
second one starts at 14 and ends at 30, size **16**.

so the first one is snapped back instead of the last one.

src/lightly_train/_task_models/object_detection_components/tiling_utils.py
```diff
@@ -83,7 +106,79 @@ def tile_image(
...
+        ious[i] = torch.where(
+            union > 0,
+            intersection.float() / union.float(),
+            torch.zeros_like(union, dtype=torch.float),
+        )
+
+    return ious
+
+
+def combine_object_detection_tiles(
```
gabrielfruet, line 181: If I understood there is a bug fix here also. Maybe we could update the changelog regarding this.

gabrielfruet (approved):

tests/_task_models/object_detection_components/test_tiling_utils.py
yutong-xiang-97: For a 30px image, 16px tile, and step 16, `_tile_starts(size=30, tile_size=16, step=16)` returns `[0, 14]`. Those are start coordinates, so the tiles are `[0:16]` and `[14:30]`.
So the last tile is the one snapped back to size - tile_size = 14; the first tile is unchanged.`

src/lightly_train/_task_models/object_detection_components/tiling_utils.py
yutong-xiang-97: This function is not used elsewhere and I believe it only creates a duplicate so no significant impact.

PR #776, opened 2026-06-12
Add Metric Depth Estimation (DAv3)

mrpositron:
## What has changed and why?

Adds metric depth estimation for Depth Anything V3. The new `DepthAnythingV3MetricDepthEstimation` task model (`dinov2/dav3-metric-large`) reuses the relative model's backbone and DPT decoder, and scales the output to metric units via `metric_depth = focal * output / 300` when camera intrinsics are given. `dpt.py` is moved to the shared `depth_estimation_components/` package, and `convert_checkpoint.py` now also converts the official `DA3METRIC-LARGE` weights.

## How has it been tested?

NYUv2 Benchmark:
| Model | a1 ↑ | a2 ↑ | a3 ↑ | abs_rel ↓ | log_10 ↓ | rmse ↓ | rmse_log ↓ | silog ↓ | sq_rel ↓ |
|---|---|:-:|:-:|:-:|:-:|:-:|:-:|:-:|:-:|
| Official Implementation| 0.948 | 0.984 | 0.992 | 0.08 | 0.036 | 0.342 | 0.116 | 9.112 | 0.047 |
| LightlyTrain Port | 0.948 | 0.984 | 0.992 | 0.08 | 0.036 | 0.342 | 0.116 | 9.103 | 0.046 |

- [ ] Yes

- [ ] Yes


mrpositron:
@codex

tests/_task_models/depth_anything_v3_depth_estimation/test_task_model.py
```diff
@@ -21,18 +21,11 @@
 @pytest.fixture()
 def tiny_model_args() -> dict[str, Any]:
     return {
-        "backbone_name": "custom",
```
liopeer, line 24: Deleted because they're anyway the defaults? If you do unrelated things, best to quickly leave a comment.

src/lightly_train/_task_models/dinov2_dav3_metric_depth_estimation/task_model.py
```diff
@@ -0,0 +1,473 @@
...
+class DepthAnythingV3MetricDepthEstimation(TaskModel):
+    """Depth Anything V3 metric-depth inference model."""
+
+    model_suffix = "dav3_metric_large"
+
+    def __init__(
+        self,
+        *,
+        model_name: str = "dinov2/dav3-metric-large",
+        process_resolution: int = 504,
```
liopeer: This is probably in the other model as well, but shouldn't we call it `image_size` to comply with the other task models?

src/lightly_train/_task_models/dinov2_dav3_metric_depth_estimation/task_model.py
```diff
@@ -0,0 +1,473 @@
...
+    def is_supported_model(cls, model: str) -> bool:
+        try:
+            cls.parse_model_name(model_name=model)
+        except ValueError:
+            return False
+        else:
+            return True
+
+    @classmethod
+    def parse_model_name(cls, model_name: str) -> str:
```
liopeer: If this is not required by the `TaskModel` class I would leave it away. We can collapse it completely into `is_supported_model` and keep things tidy.

src/lightly_train/_task_models/dinov2_dav3_metric_depth_estimation/task_model.py
```diff
@@ -0,0 +1,473 @@
...
+        raise ValueError(
+            f"Model name '{model_name}' is not supported. Available models are: "
+            f"{cls.list_model_names()}."
+        )
+
+    @torch.no_grad()
+    def predict(
+        self,
+        image: PathLike | PILImage | Tensor,
+        intrinsics: Tensor | None = None,
```
liopeer: If `intrinsics` can be none, are we then not exactly back to relative depth? In that case I don't fully understand why we should have 2 task models.

src/lightly_train/_task_models/dinov2_dav3_metric_depth_estimation/task_model.py
```diff
@@ -0,0 +1,473 @@
...
+
+import torch
+import torch.nn.functional as F
+from PIL.Image import Image as PILImage
+from torch import Tensor
+
+from lightly_train._data import cache, download, file_helpers
+from lightly_train._env import Env
+from lightly_train._models.dinov2_vit.dinov2_vit_package import DINOV2_VIT_PACKAGE
+from lightly_train._task_models.depth_estimation_components import image_utils
```
liopeer: Consider import as `depth_image_utils`.

src/lightly_train/_task_models/dinov2_dav3_metric_depth_estimation/task_model.py
```diff
@@ -0,0 +1,473 @@
...
+                )
+            tensors.append(x)
+            metadata.append(meta)
+        batch = self.preprocess_batch(tensors)
+        raw = self.forward(batch)
+        return self.postprocess(raw, metadata)
+
+    def preprocess_image(
+        self, image: PathLike | PILImage | Tensor
+    ) -> tuple[Tensor, dict[str, Any]]:
```
liopeer, line 256: You can type this more narrowly as `dict[str, int]`.

tests/_task_models/dinov2_dav3_metric_depth_estimation/test_task_model.py
```diff
@@ -0,0 +1,114 @@
...
+    return {
+        "out_layers": (0, 1, 2, 3),
+        "image_size": 56,
+        "patch_size": 14,
+        "features": 16,
+        "out_channels": (8, 16, 32, 32),
+    }
+
+
+def test_task_model__predict_returns_original_resolution(
```
liopeer: You could group these under `TestDepthAnythingV3MetricDepthEstimation`. Then the tests would be called: `test_predict__returns_original_resolution` etc.

src/lightly_train/_task_models/dinov2_dav3_metric_depth_estimation/task_model.py
mrpositron: We could call it `image_size`, but it is not equivalent to `image_size` in the other task models. There, `image_size` defines both height and width; here, it defines the size of only one side.

src/lightly_train/_task_models/dinov2_dav3_metric_depth_estimation/task_model.py
mrpositron: Nope, the model output is not relative depth. These are separate models (from the relative depth) with different outputs. I will remove the `None` option from intrinsics.

src/lightly_train/_task_models/dinov2_dav3_metric_depth_estimation/task_model.py
mrpositron: The dict is heterogeneous. `preprocess_image` returns `orig_h`/`orig_w` (int), but predict adds a float `focal`.

src/lightly_train/_task_models/dinov2_dav3_metric_depth_estimation/task_model.py
mrpositron: Added `# type: ignore[override]` to `predict` and `predict_batch`, keeping `intrinsics` as a required parameter.

src/lightly_train/_task_models/dinov2_dav3_metric_depth_estimation/task_model.py
mrpositron: Renamed `process_resolution` -> `inference_size`, moved into _MODEL_CONFIGS (fixed per-model, not user-facing)

liopeer: I still think we should have named it `image_size`, but let's address this later when we also move the stuff to torchvision. The slightly annoying thing is that we then have to convert+upload the checkpoints again, but it's not crazy much work. Feel free to resolve.

liopeer (approved):
LGTM!

PR #779, opened 2026-06-15
Add Depth Estimation (DAv2) to LightlyTrain 

mrpositron:
## What has changed and why?

Adds relative depth estimation with Depth Anything V2, a faithful port of the official implementation.

- New **DepthAnythingV2RelativeDepthEstimation** task model with three variants: `dav2-relative-small`/`base`/`large`. Reuses the shared DPT decoder (no sky head) and register-free DINOv2 backbone so the architecture matches the official checkpoint. `predict`/`predict_batch` return depth at the original input resolution.
- `process_image_dav2` preprocessing: aspect-preserving cubic resize reproducing cv2.INTER_CUBIC with no OpenCV dependency.
- Hosted weights aren't available yet, pass a local weights=<converted .pt> from the convert script for now.

## How has it been tested?

- Outputs validated against the official implementation

- [ ] Yes

- [ ] Yes


liopeer (approved):
LGTM!

PR #787, opened 2026-06-18
Add EdgeCrafter Components for Instance Segmentation

yutong-xiang-97:
## What has changed and why?

Add from EdgeCrafter
- instance segmentation head: https://github.com/Intellindust-AI-Lab/EdgeCrafter/blob/main/ecdetseg/engine/edgecrafter/segmentation_head.py
- decoder: https://github.com/Intellindust-AI-Lab/EdgeCrafter/blob/main/ecdetseg/engine/edgecrafter/decoder.py
- postprocessor: https://github.com/Intellindust-AI-Lab/EdgeCrafter/blob/main/ecdetseg/engine/edgecrafter/postprocessor.py

to `instance_segmentation_components`.

There will be a refactoring to use a shared `ltdetr_components` and make inheritance nicer between some of the classes

## How has it been tested?

N/A

- [ ] Yes

- [ ] Yes


yutong-xiang-97:
@claude review once

src/lightly_train/_task_models/instance_segmentation_components/edgecrafter_postprocessor.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+class EdgeCrafterInstanceSegmentationPostProcessor(RTDETRPostProcessor):
+    """Postprocessor for LTDETR instance segmentation outputs."""
+
+    def forward(  # type: ignore[override]
+        self,
+        outputs: dict[str, Tensor],
+        orig_target_sizes: Tensor,
+    ) -> (
```
gabrielfruet, line 41: Not something you should address. But I think in the future we should, at least in the main touch points (main outputs of the model, inputs of post processors), use TypedDicts, dataclasses or TensorDict.

src/lightly_train/_task_models/object_detection_components/dfine_decoder.py
```diff
@@ -840,6 +873,10 @@ def _get_encoder_input(self, feats: List[torch.Tensor]):
         # [b, l, c]
         return torch.concat(feat_flatten, 1), spatial_shapes
 
+    def _get_encoder_input(self, feats: List[torch.Tensor]):
```
gabrielfruet, line 876: Type hint missing here

gabrielfruet (approved):

gabrielfruet (approved):

src/lightly_train/_task_models/instance_segmentation_components/edgecrafter_postprocessor.py
yutong-xiang-97: This was inherited from the original EdgeCrafter code so we try to minimize things that we add unless absolutely necessary.
But you are right, we need to maintain these code so it would make sense at one point.

src/lightly_train/_task_models/object_detection_components/dfine_decoder.py
yutong-xiang-97: Hmm the original code does not have any type hints and we only add these type hints with our own small refactor. Let's leave it for now.

PR #790, opened 2026-06-19
Dependency group for tensorrt and ONNX

liopeer:
## What has changed and why?
- pins dependencies for tensorrt-cu13 and onnxruntime-gpu
- prework for running benchmarks in CI

Note: This is a dependency group, i.e. this is not exposed/distributed to the users and purely internal.

## How has it been tested?
- installs on compute node

- [ ] Yes

- [ ] Yes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,6 +84,10 @@ dist = [
     "twine>=6.1.0",
     "wheel>=0.45.1",
 ]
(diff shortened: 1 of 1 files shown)


liopeer (commented):
/review

yutong-xiang-97 (approved):

PR #791, opened 2026-06-19
Refactor LTDETR task model

yutong-xiang-97:
## What has changed and why?

Refactor LTDETR so that both object detection and instance segmentation can inherit the same task model

## How has it been tested?

N/A

- [ ] Yes

- [ ] Yes


liopeer (commented):
LGTM!

liopeer (approved):

PR #794, opened 2026-06-19
Add benchmark command to docs

liopeer:
## What has changed and why?
- document benchmark command
- small changes

## How has it been tested?

(Delete this: Please describe the tests that you ran to verify your changes. Provide
instructions so we can reproduce. Please also list any relevant details for your test
configuration.)

- [ ] Yes

- [ ] Yes


src/lightly_train/_commands/benchmark_task.py
```diff
@@ -101,7 +101,10 @@ def benchmark_object_detection(
             Human-readable name for the dataset (e.g. ``"COCO 2017"``).
             Included in the benchmark report.
         data:
-            Dataset configuration dictionary (same format as train_object_detection).
+            The dataset configuration or path to a YAML file with the configuration
+            (same format as train_object_detection). The ``format`` defaults to
+            ``"yolo"`` if not specified. See the documentation for more information:
```
simonschoelly: I left the default out here on purpose. I had previously only added a default for the train task to keep it consistent with what we had before.

As you know seem to have merged the duplicate code - perhaps we can keep in the default but not expose it in the docs?

docs/source/object_detection.md
```diff
@@ -799,3 +815,150 @@ when exporting to TensorRT.
...
+Builds a TensorRT engine for fast GPU inference (see
+[Exporting a Checkpoint to TensorRT](#object-detection-tensorrt)).
+
+```python
+result = lightly_train.benchmark_object_detection(
+    ...,
+    backend_args={
+        "format": "tensorrt",
+        "precision": "fp16",         # One of "fp32", "fp16".
+        # "export_args": {...},      # Optional, forwarded to model.export_onnx().
```
simonschoelly: This is actually forwarded to `model.export_tensorrt`.

src/lightly_train/_commands/benchmark_task.py
liopeer: I would like to have the fallback to `"yolo"` as the default (so that things behave the same way between training and benchmarking). But I agree that it's actually not really desirable to have this default. So I am fine with not documenting it.

docs/source/object_detection.md
liopeer: True!

yutong-xiang-97 (approved):

PR #795, opened 2026-06-21
Add task_model.py for LTDETR Instance Segmentation

yutong-xiang-97:
## What has changed and why?

Add `task_model.py` for LTDETR Instance Segmentation. 

Note that we don't include `export_onnx` and `export_tensorrt` for now.

Also
- reordered the methods in LTDETR object detection model
- remove `mask_spatial_level` and use fixed `0` to follow the original EdgeCrafter design

## How has it been tested?

N/A

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_task_models/dinov3_ltdetr_instance_segmentation/task_model.py
+++ b/src/lightly_train/_task_models/dinov3_ltdetr_instance_segmentation/task_model.py
@@ -11,6 +11,7 @@ from collections.abc import Sequence
 from typing import Any, Literal
 
 import torch
+import torch.nn.functional as F
 from torch import Tensor
 
 from lightly_train._task_models.dinov3_ltdetr.task_model import (
@@ -65,7 +66,6 @@ class DINOv3LTDETRInstanceSegmentation(_DINOv3LTDETRBase):
         return EdgeCrafterInstanceSegmentationTransformer(  # type: ignore[no-untyped-call]
             **decoder_config,
             eval_spatial_size=self.image_size,
-            mask_spatial_level=0,
         )
 
     def build_postprocessor(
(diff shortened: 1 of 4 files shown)


src/lightly_train/_task_models/dinov3_ltdetr_instance_segmentation/task_model.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def build_decoder(
+        self, config: _DINOv3LTDETRConfig
+    ) -> EdgeCrafterInstanceSegmentationTransformer:
+        decoder_config = config.dfine_transformer.model_dump()
+        decoder_config.update({"num_classes": len(self.classes)})
+        return EdgeCrafterInstanceSegmentationTransformer(  # type: ignore[no-untyped-call]
+            **decoder_config,
+            eval_spatial_size=self.image_size,
+        )
```
mrpositron, line 69: Not high priority: object detection defaults to `RTDETRTransformerv2`, which resolves feature strides automatically. This instance segmentation path uses D-FINE, where the strides are hardcoded and seem to assume `patch_size=16`.

Should we either resolve the D-FINE strides as well, or raise an error for unsupported patch sizes like `patch_size=14`?

mrpositron (approved):
LGTM 🚀

src/lightly_train/_task_models/dinov3_ltdetr_instance_segmentation/task_model.py
yutong-xiang-97: `patch_size=14` is only for DINOv2 backbone. With ECViT we only have `patch_size=16`

I think it is fine for now.

PR #814, opened 2026-06-26
feat: integrate HF DebugUnderflowOverflow into fine-tuning

gabrielfruet:
## What has changed and why?

- New `src/lightly_train/_debug/` module wraps HuggingFace `DebugUnderflowOverflow` for the Fabric manual loop used by fine-tuning (`train_*` task APIs).
- `train_*` configs accept `debug_args={"underflow_overflow": {...}}`; raises early when combined with active `torch.compile`.
- Monitor is attached after Fabric setup, advanced with `set_step(step)` each iteration, and closed on the happy path + via `atexit` for exception safety.
- Per-rank report: `out/debug/underflow_overflow_rank{rank}.log`; `train.log` keeps a concise pointer.

Closes TRN-2255.

## How has it been tested?

- `pytest tests/_debug/test_underflow_overflow.py` → 9/9 pass (leaned: removed redundant config permutation and lifecycle micro-tests).
- `ruff check tests/_debug/test_underflow_overflow.py` → clean.

- [ ] Not needed (follow-up PR per user instruction)


src/lightly_train/_debug/underflow_overflow.py
```diff
@@ -0,0 +1,262 @@
...
+from __future__ import annotations
+
+import contextlib
+import logging
+from pathlib import Path
+from typing import Any
+
+import torch.nn as nn
+from torch.utils.hooks import RemovableHandle
+from transformers.debug_utils import DebugUnderflowOverflow
```
liopeer: How much effort would it be to vendor this from Transformers and not import it? If you look at all the other places where the `transformers` dependency is used across the codebase you'll see that we would like to get rid of it, so it would be best to not make us more dependent.

src/lightly_train/_commands/train_task.py
liopeer, line 1195: Are you deliberately not documenting it?

src/lightly_train/_commands/train_task.py
```diff
@@ -1385,6 +1399,14 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
             train_model_cls=train_model_cls,
             torch_compile_args=config.torch_compile_args,
         )
+        config.debug_args = get_debug_args(debug_args=config.debug_args)
```
liopeer: Please use module imports for function, i.e. `debug_args.debug_helpers`.

src/lightly_train/_commands/train_task.py
```diff
@@ -1385,6 +1399,14 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
...
+        # torch.compile and DebugUnderflowOverflow are mutually exclusive. Forward
+        # hooks and per-tensor reductions used for debugging do not interact well
+        # with compiled graphs.
+        check_compile_conflict(
```
liopeer: Same here.

src/lightly_train/_commands/train_task.py
```diff
@@ -1481,6 +1503,22 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
...
+        underflow_overflow_monitor: UnderflowOverflowMonitor | None = None
+        if config.debug_args.is_underflow_overflow_enabled():
+            assert config.debug_args.underflow_overflow is not None
+            underflow_overflow_monitor = UnderflowOverflowMonitor(
+                model=train_model,
+                debug_args=config.debug_args.underflow_overflow,
+                out_dir=out_dir,
+                global_rank=fabric.global_rank,
+            )
+            atexit.register(underflow_overflow_monitor.close)
```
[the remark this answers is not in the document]
liopeer: I don't quite understand why you are not using the monitor as a context manager to begin with, which should make things cleaner.

src/lightly_train/_debug/debug_args.py
```diff
@@ -0,0 +1,54 @@
...
+    forward hooks on all model modules and reports the absolute min/max of every weight,
+    input, and output. When a NaN or inf is detected, the last ``max_frames_to_save``
+    forward frames are dumped so that the module where values first exploded can be
+    identified. Training is aborted as soon as a NaN or inf is detected.
+    """
+
+    enabled: bool = False
+    max_frames_to_save: int = Field(default=21, ge=1)
+    trace_batch_nums: list[int] = Field(default_factory=list)
+    abort_after_batch_num: int | None = Field(default=None, ge=0)
```
liopeer: Would be great to fully document the arguments. To me it's not clear what `trace_batch_nums` and `abort_after_batch_num` are for.

src/lightly_train/_debug/underflow_overflow.py
liopeer: If it's much work it's not important. If it's only copying a single file + licencse + notice, we should probably consider it.

src/lightly_train/_debug/underflow_overflow.py
```diff
@@ -0,0 +1,262 @@
...
+    Inherits upstream's hook-registration, batch-numbering bookkeeping, and
+    tensor-analysis machinery; only overrides the hook entry point to (a) drive
+    the batch header explicitly and (b) redirect all upstream ``print()`` output
+    to a caller-supplied log file.
+    """
+
+    def __init__(
+        self,
+        model: nn.Module,
+        log_file: Any,
```
liopeer: Use narrower typing.

src/lightly_train/_debug/underflow_overflow.py
```diff
@@ -0,0 +1,262 @@
...
+            )
+        except BaseException:
+            # ``super().__init__`` registers the hooks via ``model.apply`` and
+            # ``analyse_model`` can raise on malformed modules; if registration
+            # got partway, remove the handles we captured so far before letting
+            # the exception propagate.
+            self.detach_hooks()
+            raise
+
+    def _register_forward_hook(self, module: nn.Module) -> None:  # type: ignore[override]
```
liopeer: Consider wrapping in `@override` from `typing_extensions` to make it clear to the typechecker that this has to match the signature of the superclass. Maybe you can then also drop the type ignore.

liopeer: Also, use `Module` instead of `nn.Module`.

src/lightly_train/_debug/underflow_overflow.py
```diff
@@ -0,0 +1,262 @@
...
+
+        Iterates the captured :class:`RemovableHandle` list (see
+        :meth:`_register_forward_hook`). Safe to call when no hooks are
+        registered (empty list) or when called more than once.
+        """
+        for handle in self._handles:
+            handle.remove()
+        self._handles.clear()
+
+    def forward_hook(self, module: Any, input: Any, output: Any) -> None:  # type: ignore[override]
```
liopeer: Again, use `@override` from `typing_extensions`.

liopeer: Also try to use narrower typing. `module` is easy to type, `input` and `output` might indeed be difficult to properly type.

tests/_debug/test_underflow_overflow.py
```diff
@@ -0,0 +1,213 @@
...
+        created.append(monitor)
+        return monitor
+
+    yield _make
+
+    for monitor in created:
+        monitor.close()
+
+
+def test_get_debug_args__parses_user_config() -> None:
```
liopeer: Best to separate the tests into individual modules with the same names as the original modules. Otherwise it's hard to find tests later.

liopeer (changes requested):

src/lightly_train/_debug/underflow_overflow.py
gabrielfruet: I think it's only copying one file, it's not much work.

src/lightly_train/_commands/train_task.py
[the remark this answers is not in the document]
gabrielfruet: I was afraid of modifying the main function too much. If that's not a problem, then it's good to have it like a context manager

src/lightly_train/_commands/train_task.py
```diff
@@ -1481,375 +1557,415 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
...
-            model_class_path=train_model.get_task_model().class_path,
-            model_init_args=train_model.get_task_model().init_args,
-        )
+        # Attach underflow/overflow debugger after Fabric setup so that forward hooks
+        # are registered on the device-placed model. Disabled by default. We use the
+        # monitor as a context manager so the log file is closed and hooks are
+        # detached even when an exception (including the ValueError the monitor
+        # itself raises on overflow) is caught inside train_task(...). The previous
+        # atexit-based cleanup leaked the log file and forward hooks when callers
+        # caught the ValueError around the call site.
```
liopeer: No need to mention anything about atexit since we never merged this.

src/lightly_train/_commands/train_task.py
```diff
@@ -1481,375 +1557,415 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
...
+            monitor_ctx = UnderflowOverflowMonitor(
+                model=train_model,
+                debug_args=config.debug_args.underflow_overflow,
+                out_dir=out_dir,
+                global_rank=fabric.global_rank,
+            )
+        else:
+            monitor_ctx = contextlib.nullcontext(None)
 
-        if config.resume_interrupted and checkpoint_path is not None:
```
liopeer, line 1526: I am assuming that everything starting from here is the same, just indented.

liopeer (approved):
LGTM! One small comment to address, but feel free to afterwards resolve on your own and merge.

gabrielfruet:
Fixed — dropped the atexit sentence, kept the context-manager rationale.

src/lightly_train/_commands/train_task.py
gabrielfruet: Yes, and this line added `underflow_overflow_monitor.set_step(step=step)`

PR #815, opened 2026-06-26
Hotfix for LT-DETR loading

liopeer:
## What has changed and why?
- hotfix for `load_model` that fixes incompatibility issues with old checkpoints

## How has it been tested?
- with a local script

## Did you update [CHANGELOG.md](../CHANGELOG.md)?

- [x] Yes
- [ ] Not needed (internal change)

## Did you update the documentation?

- [ ] Yes
- [x] Not needed (internal change without effects for user)

PR #811, opened 2026-06-25
feat: log gradient norm during training

gabrielfruet:
## What has changed and why?

Logs the total gradient norm (`gradient_norm`) during finetuning (task path). `TrainModel.clip_gradients` now returns the pre-clipping total norm; the train loop logs it as `gradient_norm` (shown in the console as `grad_norm` and written to all configured loggers: JSONL, TensorBoard, Weights & Biases, MLflow). When gradient clipping is disabled, the norm is computed via an unbounded (`inf`) norm — a single pass over the parameters.

Closes TRN-2254.

Scope: finetuning/task path only. The pretrain path (`_methods/*` `configure_gradient_clipping`) is out of scope for a follow-up.

## How has it been tested?

- New `tests/_task_models/test_train_model.py` — verifies the returned norm matches an independent L2 computation, for both enabled (`gradient_clip_val > 0`) and disabled clipping; no mocking.
- New `log_step` tests in `tests/_commands/test_train_task_helpers.py` — `grad_norm` shown when present, omitted when `None`.
- `ruff format`/`check` and `mypy` clean on changed files.
- Regression sweep (`tests/_task_models` + `test_train_task_helpers` + `test_train_helpers`): 323 passed, 4 skipped, 1 xfailed.
- End-to-end: confirmed `gradient_norm` written to the TensorBoard scalar and shown in `train.log` over a short CPU run.

## Did you update [CHANGELOG.md](../CHANGELOG.md)?

- [x] Yes

## Did you update the documentation?

- [x] Yes — added a "Logged metrics" section to `docs/source/settings/train_settings.md`.

PR #817, opened 2026-06-26
Revert "Consolidate Depth Estimation Tasks" (#807)

mrpositron:
Reverts #807 (merge commit `61608ff7`), which consolidated the four depth estimation task models into a single config-driven `DepthAnythingDepthEstimation`. That change is breaking, so this restores the previous four separate task models.

## Details

- #807 was not the tip of `main` — #811 and #815 landed after it. #815 also touched `task_model_helpers.py`, but in different functions (`init_model_from_checkpoint`, download-hash table), so this revert only undoes #807's change in `_raise_unknown_checkpoint_error` and leaves #815's hotfix fully intact.
- The work from #807 is preserved on branch `restore/807` (points at the original merge commit) so it can be reworked into a non-breaking change later.

## Verification

- Imports of all four restored task models succeed.
- `tests/_task_models/dinov2_dav{2,3}_{relative,metric}_depth_estimation/` — 14 passed.
- No dangling references to the removed `depth_estimation.task_model` module.


liopeer (approved):
LGTM!

PR #819, opened 2026-06-26
Configs for LTDETR OD

liopeer:
## What has changed and why?
- create registry of configs similar to how it was already done for linear segmentation

## How has it been tested?
- covered by existing unit tests

## Did you update [CHANGELOG.md](../CHANGELOG.md)?

- [ ] Yes
- [x] Not needed (internal change)

## Did you update the documentation?

- [ ] Yes
- [x] Not needed (internal change without effects for user)

PR #820, opened 2026-06-26
Rename everything in LT-DETR from DINOv3 to Generic

liopeer:
## What has changed and why?
- stacks on top of #819 
- rename DINOv3 LT-DETR model to generic LT-DETR; same for transforms and all other classes; same for filenames
- 🚨 **does not make the "old" DINOv3 TaskModel the default yet for DINOv2-based models** 🚨 

## How has it been tested?
- existing unit tests

- [ ] Yes

- [ ] Yes


src/lightly_train/_task_models/ltdetr_object_detection/task_model.py
```diff
@@ -101,7 +101,7 @@ def _resolve_transformer_config(
     return config_factory()
 
 
-class DINOv3LTDETRObjectDetection(_DINOv3LTDETRBase):
+class LTDETRObjectDetection(_DINOv3LTDETRBase):
```
gabrielfruet, line 104: Shouldn't this be renamed  also? `DINOv3LTDETRBase`

liopeer: I don't want to rename it just yet, because it's also used by instance segmentation.

gabrielfruet (approved):

PR #826, opened 2026-06-29
feat: add NaN capture and replay for fine-tuning

gabrielfruet:
## What has changed and why?

- New `NaNCapture` debug tool for fine-tuning (`train_*` task APIs): when a
  NaN/Inf is detected in parameter gradients (scanned after gradient
  accumulation, before `clip_gradients`/`optimizer.step`), it captures
  reproducible state and halts training.
- Capture is a single self-contained file
  `out_dir/debug/nan_capture/rank{R}/nan_capture.pt` holding the model state
  dict, the `TrainModel` class path + init kwargs (for reconstruction), the
  step's microbatches, and torch/CUDA RNG state. The standard
  `checkpoints/last.ckpt` is **not** touched, so `resume_interrupted` is
  unaffected.
- Adds `load_nan_capture(dir)` + `NaNCaptureState.replay()` for zero-setup
  replay: reconstructs the model, restores RNG, and re-runs the triggering
  forward+backward (mirrors the training loop; stops before the optimizer step)
  to reproduce the NaN deterministically in a notebook/REPL.
- Wired into `_commands/train_task.py` alongside the existing underflow/overflow
  monitor from #814. Enable with `debug_args={"nancapture": {"enabled": True}}`.

Replay is debug-only — the training loop carries no replay flag.

Closes TRN-2256.

## How has it been tested?

- `pytest tests/_debug/test_nan_capture.py` → 11/11 pass (config, monitor grad
  scan + capture payload, buffer clone/detach/reset, replay roundtrip +
  reproduction).
- `make format` → clean (no unintended changes).
- `mypy` on the 4 changed files → clean.

- [x] Yes

- [ ] Not needed (debugging API; docs deferred, matching #814)


src/lightly_train/_commands/train_task.py
```diff
@@ -1574,7 +1582,26 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
...
+            )
+        else:
+            nancapture_ctx = contextlib.nullcontext(None)
+
+        # Compose both debug monitors at the same scope so the training-loop
+        # body stays at a single indent level. ExitStack teardown closes them in
+        # LIFO order after `timer.stop()`/`Training completed.` log below.
+        with contextlib.ExitStack() as debug_stack:
+            underflow_overflow_monitor = debug_stack.enter_context(monitor_ctx)
+            nancapture_monitor = debug_stack.enter_context(nancapture_ctx)
```
liopeer: Nit: I personally prefer

```suggestion
        with (
            monitor_ctx as underflow_overflow_monitor,
            nancapture_ctx as nancapture_monitor,
        ):
```

src/lightly_train/_debug/nan_capture.py
```diff
@@ -0,0 +1,457 @@
+#
+# Copyright (c) Lightly AG and affiliates.
```
liopeer, line 2: Was this not vendored from somewhere?

liopeer (approved):
LGTM! Adressing the comments is optional.

src/lightly_train/_commands/train_task.py
gabrielfruet: My agent said that it wasn't compatible with 3.8 but I tested locally and doesn't seem to be the case. I also prefer that way.

src/lightly_train/_debug/nan_capture.py
gabrielfruet: I think we can just add from where we took (a medium article). But not strictly necessary I think, since it has no licesing.

PR #830, opened 2026-06-30
Completely Reproduce DINOv2 LT-DETR in Generic One

liopeer:
## What has changed and why?
- stacks on top of #831 
- completely reproduces settings of DINOv2 LT-DETR inside the generic TrainModel class

After this PR:
- make sure that all checkpoints (DINOv3 or DINOv2 are re-routed to the new class)
- delete all DINOv2-specific code
- manually do a lot of e2e tests to make sure that the whole migration worked
- some cleanups (probably another PR)

## How has it been tested?
- existing + extra unit tests

- [ ] Yes

- [ ] Yes


src/lightly_train/_task_models/ltdetr_object_detection/train_model.py
```diff
@@ -263,6 +282,74 @@ def effective_losses(self) -> list[str]:
...
+    default_batch_size: ClassVar[int] = 16
+    default_steps: ClassVar[int] = (
+        100_000 // 16 * 72
+    )  # TODO (Lionel, 10/25): Adjust default steps.
+
+    decoder_name: Literal["rtdetrv2", "dfine"] = "rtdetrv2"
+
+    # DINOv2 ViT backbones are always patch-14 (vits14/vitb14/vitl14/vitg14), matching
+    # LTDETR_MODEL_REGISTRY's dinov2 backbone_args.
+    patch_size: int = 14
```
liopeer, line 297: In a clean-up I will set these to `"auto"` and resolve them from the registry.

src/lightly_train/_task_models/ltdetr_object_detection/transforms.py
```diff
@@ -427,3 +427,343 @@ class LTDETRObjectDetectionTrainTransform(ObjectDetectionTransform):
 
 class LTDETRObjectDetectionValTransform(ObjectDetectionTransform):
     transform_args_cls = LTDETRObjectDetectionValTransformArgs
+
+
+# TODO (Lionel, 06/26): Remove all the `v2` naming once the DINOv2 LT-DETR models are
+# completely migrated to the generic LTDETR pipeline.
+class DINOv2LTDETRObjectDetectionRandomPhotometricDistortArgsV2(
```
liopeer, line 434: For now just keeping an exact copy of the original augmentation pipeline here. In a follow-up I can check if it makes sense to bring the pipelines closer together. However, DINOv2 uses patch size 14, and the others use 16, so the pipelines will never be 100% the same.

tests/_task_models/ltdetr_object_detection/test_task_model.py
```diff
@@ -30,6 +30,9 @@
     DFINETransformerConfig,
     RTDETRTransformerv2Config,
 )
+from lightly_train._task_models.ltdetr_object_detection.dino_vit_wrapper import (
```
yutong-xiang-97, line 33: Not part of this PR, but I would like to highlight that `test_task_model.py` actually builds the train model for lots of tests. I think either they should go to `test_train_model.py` if they make sense or should be dropped. `test_task_model.py` should focus on testing task model behavior instead.

yutong-xiang-97 (approved):

tests/_task_models/ltdetr_object_detection/test_task_model.py
liopeer: Really good point. Let me address this in the cleanup.

PR #831, opened 2026-07-01
Reduce Double-Truth for LT-DETR Checkpoints and Aliases

liopeer:
## What has changed and why?
- we had two places that defined model strings, for LT-DETR we reduce this with this PR
- In order to make this more proper I actually revert the changes introduced in #791 that concern the object detection model. We can potentially later revisit those changes.

## How has it been tested?
- existing unit tests (functionality should not change)

- [ ] Yes

- [ ] Yes


src/lightly_train/_task_models/ltdetr_object_detection/task_model.py
```diff
@@ -268,6 +267,34 @@ def __init__(
         if self.backbone_freeze:
             self.freeze_backbone()
 
+    @override
```
yutong-xiang-97, line 271: Is it necessary to add a `@override` decorator here? If so, why not also for `list_model_names`?

src/lightly_train/_task_models/ltdetr_object_detection/task_model.py
```diff
@@ -10,7 +10,7 @@
 import logging
 from collections.abc import Sequence
 from copy import deepcopy
-from typing import Any, Callable, Literal, Union, cast
+from typing import Any, Callable, Literal, Self, Union, cast, override
```
yutong-xiang-97: From Claude Code:


from typing import Any, Callable, Literal, Self, Union, cast, override — typing.Self exists only on Python 3.11+ and typing.override only on 3.12+. This repo declares requires-python = ">=3.8" and CI runs the minimal suite on MINIMAL_PYTHON_VERSION := 3.8. Importing this module raises ImportError on every interpreter below 3.12, so any LTDETR train/predict/load path (and test collection) fails. The rest of the codebase imports both symbols from typing_extensions for exactly this reason

src/lightly_train/_configs/model_registry.py
```diff
@@ -7,27 +7,47 @@
...
+    sha256: str
+
+
+@dataclass(frozen=True)
+class ModelAlias:
+    name: str
+    downloadable_checkpoint: DownloadableCheckpoint
+
+
+AliasT = str | ModelAlias
```
yutong-xiang-97: From Claude Code

AliasT = str | ModelAlias is a module-level runtime assignment (not an annotation), so from __future__ import annotations does not defer it. PEP 604 X | Y at runtime requires Python 3.10+; on 3.8/3.9 this raises TypeError at import time. model_registry is imported broadly via the config layer, so this breaks import on 3.8/3.9. Codebase convention for runtime aliases is Union[...] (see types.py:220 PathLike = Union[str, Path]).
Fix direction: AliasT = Union[str, ModelAlias] (import Union), or make it AliasT: TypeAlias = "str | ModelAlias"

src/lightly_train/_task_models/ltdetr_object_detection/config.py
```diff
@@ -723,7 +807,7 @@ class DINOv3ViTLarge(LTDETRBaseConfig.ViTLarge):
             default_factory=lambda: {"patch_size": 16}
         )
 
-    @LTDETR_MODEL_REGISTRY.register("dinov2/vits14-ltdetr")
+    @LTDETR_MODEL_REGISTRY.register("dinov2/_vittest14-ltdetr", "dinov2/vits14-ltdetr")
```
yutong-xiang-97, line 810: This would expose `_vittest14-ltdetr` to the public via calling `list_model_names()` and `list_aliases()`.

src/lightly_train/_configs/model_registry.py
liopeer: Thx, fixing that and the one above.

src/lightly_train/_task_models/ltdetr_object_detection/task_model.py
liopeer: Same here.

src/lightly_train/_task_models/ltdetr_object_detection/task_model.py
liopeer: It's generally good to have `@override`, because then mypy would realize if your implementation had a different signature from the one of the base class (useful for refactorings of the base class). But this was probably added before I decided to completely drop the `_DINOv2LTDETRBase` base class. I'll check for all of them if they're still useful.

yutong-xiang-97 (approved):

PR #834, opened 2026-07-03
Refactor EoMT segmentation transforms into a shared package

yutong-xiang-97:
## What has changed and why?

This is part of the refactoring for adding LTDETR instance segmentation transforms because using `instance_segmentation_transform.py` dedicated for EoMT is not feasible anymore.

In light of this, we consolidate the three EoMT segmentation transform modules into a new `_transforms/eomt_transforms/` with the following changes:

- Moving the classes
  - `instance_segmentation_transform.py` -> `eomt_transforms/instance_segmentation.py`
  - `panoptic_segmentation_transform.py` -> `eomt_transforms/panoptic_segmentation.py`
  - `semantic_segmentation_transform.py` -> `eomt_transforms/semantic_segmentation.py`
- Prefixing the classes with `EoMT` (e.g. `SemanticSegmentationTransform` → `EoMTSemanticSegmentationTransform`).
- The duplicated `resolve_incompatible` and transform-list-building code (previously copy-pasted across all three modules) is factored into `eomt_transforms/utils.py` as `_resolve_incompatible_eomt` and `_build_transforms_eomt`.

This way, we can make the transform pipeline fully owned by the specific model architectures now that it is the biggest differentiator. It is helpful for maximizing the reusability of pipeline code that serves the same architecture while allowing flexibility among different tasks the architecture supports.

In the future we could also push the transforms forward to `_task_models/` after refactoring all the EoMT models there.

## How has it been tested?

Parity verified by AI.

- [ ] Yes

- [ ] Yes


src/lightly_train/_transforms/semantic_segmentation_transform.py
```diff
@@ -1,280 +0,0 @@
...
-        self.transform = Compose(transform, additional_targets={"mask": "mask"})
-
-    def __call__(
-        self, input: SemanticSegmentationTransformInput
-    ) -> SemanticSegmentationTransformOutput:
-        transformed = self.transform(image=input["image"], mask=input["mask"])
-        return {"image": transformed["image"], "mask": transformed["mask"]}
-
-
-class SemanticSegmentationCollateFunction(TaskCollateFunction):
```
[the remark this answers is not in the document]
yutong-xiang-97: Does not affect loading the existing checkpoints.

src/lightly_train/_task_models/linear_semantic_segmentation/transforms.py
```diff
@@ -11,9 +11,9 @@
 
 from pydantic import Field
 
-from lightly_train._transforms.semantic_segmentation_transform import (
```
liopeer, line 14: I find it a bit weird that we now base ourselves off of an EoMT-specific transform for linear segmentation/multi-head segmentation (because they have nothing to do with EoMT). But this is anyway code we rarely touch, so I can live with it.

src/lightly_train/_transforms/eomt_transforms/__init__.py
```diff
@@ -0,0 +1,46 @@
+#
+# Copyright (c) Lightly AG and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+#
+from lightly_train._transforms.eomt_transforms.instance_segmentation import (
```
liopeer: We usually leave `__init_` files empty for internal code and prefer verbose  import paths.

src/lightly_train/_transforms/eomt_transforms/instance_segmentation.py
```diff
@@ -0,0 +1,176 @@
...
+from albumentations import BboxParams, Compose
+from torch import Tensor
+
+from lightly_train._transforms.eomt_transforms.utils import (
+    _build_transforms_eomt,
+    _resolve_incompatible_eomt,
+)
+from lightly_train._transforms.task_transform import (
+    TaskCollateFunction,
+    TaskTransform,
```
liopeer, line 23: I am not checking this entire module. I assume it's just moving + renaming from generic to `EoMT_`. Would generally be better to quickly split moving and renaming into 2 PRs.

liopeer (approved):
The design looks okay. But I did not go through every line of code, I am assuming that basically everything is just moving + renaming.

src/lightly_train/_task_models/linear_semantic_segmentation/transforms.py
yutong-xiang-97: I also find it a bit weird but I can live with it as well. Linear segmentation is mostly used for benchmarking than actual production training.

src/lightly_train/_transforms/eomt_transforms/instance_segmentation.py
yutong-xiang-97: I would keep them into one PR to make merging faster.

PR #835, opened 2026-07-03
Refactor Object Detection Transforms

yutong-xiang-97:
## What has changed and why?

Refactored the LT-DETR object detection transform implementation into a shared `ltdetr_transforms` package with separate components for step-aware control, object detection, and utility modules. Updates all the relevant task model call sites to use the new structure while preserving the existing object detection behavior.

The new structure is split as follows:

- `ltdetr_transforms/components.py` contains the reusable runtime pieces for LT-DETR-style transforms: step activation tracking, dataloader reinitialization checks, cached sample transform construction, and mosaic activation handling.
- `ltdetr_transforms/object_detection.py` contains the object-detection transform and collate-function implementation, including input/output types, transform args contract, bbox normalization, degenerate-box filtering, mixup/copyblend/scale-jitter handling, and batch transform wiring.
- `ltdetr_transforms/utils.py` contains helper functions for image-size resolution, activation schedules, albumentations transform ordering, bbox conversion/filtering, and normalization.

The LT-DETR task-model transform modules now define task-specific defaults close to the task models, while reusing the object-detection transform/collate implementation and the shared components.

Also refactored the unit test accordingly.

## How has it been tested?

Parity verified by AI.

- [ ] Yes

- [ ] Yes


src/lightly_train/_transforms/ltdetr_transforms/base.py
```diff
@@ -0,0 +1,385 @@
...
+# Shared LT-DETR augmentation presets. These carry the default augmentation
+# hyperparameters used by both object detection and instance segmentation (the two
+# tasks share the same augmentation pipeline, matching the upstream ECDet/ECSeg configs).
+class LTDETRRandomPhotometricDistortArgs(RandomPhotometricDistortArgs):
+    prob: float = 0.5
+
+    brightness: tuple[float, float] = (0.875, 1.125)
+    contrast: tuple[float, float] = (0.5, 1.5)
+    saturation: tuple[float, float] = (0.5, 1.5)
+    hue: tuple[float, float] = (-0.05, 0.05)
```
liopeer: I think the `_transforms` subpackage should have as few default values. It's okay to define the base classes here (i.e. the contract), but the values/implementations should optimally be implemented right next to the task models.

src/lightly_train/_commands/benchmark_task.py
```diff
@@ -419,7 +419,7 @@ def _create_val_dataloader(
     val_dataset_args = data_args.get_val_args()
     dataset_cls = val_dataset_args.get_dataset_cls()
     image_info = list(val_dataset_args.list_image_info())
-    transform = ObjectDetectionTransform(transform_args=transform_args)
+    transform = LTDETRObjectDetectionTransform(transform_args=transform_args)
```
liopeer, line 422: I think this is also supposed to work with PicoDet, so using LTDETRObjectDetectionTransform to it is really confusing.

src/lightly_train/_task_models/picodet_object_detection/transforms.py
```diff
@@ -72,7 +72,7 @@ class PicoDetScaleJitterArgs(ScaleJitterArgs):
     divisible_by: int | None = None
 
 
-class PicoDetObjectDetectionTrainTransformArgs(ObjectDetectionTransformArgs):
+class PicoDetObjectDetectionTrainTransformArgs(LTDETRObjectDetectionTransformArgs):
```
liopeer, line 75: This is really confusing that PicoDet inherits from the LT-DETR transform.

src/lightly_train/_transforms/oriented_object_detection_transform.py
```diff
@@ -49,7 +49,7 @@ class OrientedObjectDetectionTransformOutput(TaskTransformOutput):
     class_labels: NotRequired[Tensor]
 
 
-class OrientedObjectDetectionTransformArgs(ObjectDetectionTransformArgs):
```
liopeer, line 52: Also confusing that OBB transform (which does not carry the LT-DETR name), inherits from LT-DETR.

liopeer (changes requested):
I think in terms of hierarchy we need to discuss, because some hierarchy decisions are difficult to follow (e.g. PicoDetTransform inheriting from LTDETRTransform). We also now have imo to many levels of inheritance (we have the base transform, then LT-DETR transform and the specialized ones on top. One level is okay, but two are imo too much. Instead of inheriting anything from the first level, that first level should be completely abstract (protocols or ABCs).

src/lightly_train/_task_models/picodet_object_detection/transforms.py
yutong-xiang-97: This shouldn't be the case in the very beginning, or is it the case in the original PicoDet paper, but we somehow decided that PicoDet shares the same transform args as LTDETR. Theoretically, we should have separate classes, but I will only include them in a follow-up PR to not bloat this one.

src/lightly_train/_commands/benchmark_task.py
yutong-xiang-97: Will fix in a follow-up as stated. But tbh I would really not care too much about PicoDet since it has serious issues to be fixed before being actually usable anyway.

yutong-xiang-97:
Instead of using three-level inheritance with the shared logic as the intermediate child (refused bequest issue of refactoring), the common step-aware control logic is moved into reusable components in `components.py` (favor composition over inheritance).

The new structure is split as follows:

- `ltdetr_transforms/components.py` contains the reusable runtime pieces for LT-DETR-style transforms: step activation tracking, dataloader reinitialization checks, cached sample transform construction, and mosaic activation handling.
- `ltdetr_transforms/object_detection.py` contains the object-detection transform and collate-function implementation, including input/output types, transform args contract, bbox normalization, degenerate-box filtering, mixup/copyblend/scale-jitter handling, and batch transform wiring.
- `ltdetr_transforms/utils.py` contains helper functions for image-size resolution, activation schedules, albumentations transform ordering, bbox conversion/filtering, and normalization.

The commit also fixes oriented object detection after removing the old LT-DETR base module. 

Note that PicoDet still inherits from the LT-DETR object detection transform stack, which is confusing and should be cleaned up in a follow-up PR.

src/lightly_train/_commands/benchmark_task.py
liopeer: I am not worried about PicoDet. But the benchmark command should not depend on something LT-DETR-specific.

src/lightly_train/_task_models/picodet_object_detection/transforms.py
liopeer: No, but before it was a "generic" object detection transform (at least in naming). Now with the renaming it's really confusing.

src/lightly_train/_transforms/ltdetr_transforms/components.py
```diff
@@ -0,0 +1,232 @@
...
+from albumentations import Compose
+
+from lightly_train._transforms.ltdetr_transforms.utils import (
+    build_ltdetr_sample_transform_parts,
+    is_step_start_or_stop_configured,
+    ordered_ltdetr_sample_transforms,
+)
+from lightly_train._transforms.mosaic import MosaicTransform
+
+if TYPE_CHECKING:
```
liopeer: Why would we need this? Is there a circular dependency otherwise?

src/lightly_train/_data/object_detection_dataset.py
```diff
@@ -35,7 +35,7 @@ class ObjectDetectionDataset(TaskDataset):
     dataset_args: COCOObjectDetectionDatasetArgs | YOLOObjectDetectionDatasetArgs  # type: ignore[assignment]
 
     batch_collate_fn_cls: ClassVar[type[TaskCollateFunction]] = (
-        ObjectDetectionCollateFunction
+        LTDETRObjectDetectionCollateFunction
```
liopeer, line 38: There should probably still exist a super thin protocol/ABC class `ObjectDetectionCollateFunction`. After all it's the object detection dataset args and not the `LTDETRObjectDetectionDatasetArgs`. Can also be done in a follow-up, but it would make naming much less confusing.

src/lightly_train/_transforms/object_detection_transform.py
```diff
@@ -1,727 +0,0 @@
-#
```
liopeer, line 1: We can keep `ObjectDetectionTransform` as a super lightweight protocol as well.

liopeer (approved):
Already quite a bit clearer. Would encourage to fix the inconsistencies in a follow-up, but would first focus on actually implementing the instance segmentation stuff.

src/lightly_train/_task_models/picodet_object_detection/transforms.py
yutong-xiang-97: In a follow-up PR I would create a new class called `PicoDetObjectDetectionTransformArgs` and not subclass from LTDETR at all. I would not assume a generic object detection transform unless for a protocol. But now I don't see an absolute need to have one though.

src/lightly_train/_commands/benchmark_task.py
yutong-xiang-97: There can be a way to make `benchmark_task.py` agnostic to the transforms and collate functions, but it will be in a separate PR in case needed.

liopeer: That's fine. I just want to avoid that we do a refactor, but in the end leave things more complicated than they were initially.

src/lightly_train/_data/object_detection_dataset.py
yutong-xiang-97: There is also a way to avoid this which can be in a follow-up PR, but a simple protocol or ABC won't work for now because the implementation currently involves class constructing or value assignment.

src/lightly_train/_transforms/ltdetr_transforms/components.py
yutong-xiang-97: I used a protocol transform_args.py to solve it since we in any case need it for instance segmentation.

PR #836, opened 2026-07-03
Add Instance Segmentation Transforms

yutong-xiang-97:
## What has changed and why?

This PR stacks on top of #835.

It adds LT-DETR instance segmentation transform support on top of the object detection transform refactor:

- add instance segmentation dataset mask handling
- add LT-DETR instance segmentation transform wiring
- add a reusable `ltdetr_transforms` instance segmentation transform module
- add focused transform coverage for the new instance segmentation path

## How has it been tested?

Unit tests.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/train_task.py
+++ b/src/lightly_train/_commands/train_task.py
@@ -1497,14 +1497,14 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
             model_name=config.model,
             val_transform_args=val_transform_args,
         )
-        train_collate_fn = train_dataset.batch_collate_fn_cls(
-            split="train", transform_args=train_transform_args
-        )
-        val_collate_fn = val_dataset.batch_collate_fn_cls(
-            split="val", transform_args=val_transform_args
-        )
         train_dataset.set_transform(train_transform)
         val_dataset.set_transform(val_transform)
+        train_collate_fn = train_dataset.get_batch_collate_fn_cls()(
+            split="train", transform_args=train_transform_args
+        )
+        val_collate_fn = val_dataset.get_batch_collate_fn_cls()(
+            split="val", transform_args=val_transform_args
+        )
         train_dataloader.collate_fn = train_collate_fn  # type: ignore[assignment]
         val_dataloader.collate_fn = val_collate_fn  # type: ignore[assignment]
         helpers.disable_persistent_workers_for_step_aware_transforms(
--- a/src/lightly_train/_transforms/ltdetr_transforms/instance_segmentation.py
+++ b/src/lightly_train/_transforms/ltdetr_transforms/instance_segmentation.py
@@ -11,23 +11,42 @@ from typing import Literal
 
 import numpy as np
 import torch
-from albumentations import Compose
+from albumentations import BboxParams, Compose
 from albumentations.pytorch.transforms import ToTensorV2
+from pydantic import ConfigDict
 from torch import Tensor
 
-from lightly_train._transforms.ltdetr_transforms.base import (
-    LTDETRTransformArgs,
-    _LTDETRCollateFunction,
-    _LTDETRTransform,
+from lightly_train._transforms.ltdetr_transforms.components import (
+    StepActivationTracker,
+    StepScheduledCompose,
 )
 from lightly_train._transforms.ltdetr_transforms.utils import (
     filter_degenerate_yolo_boxes,
 )
 from lightly_train._transforms.task_transform import (
+    TaskCollateFunction,
+    TaskTransform,
+    TaskTransformArgs,
     TaskTransformInput,
     TaskTransformOutput,
 )
+from lightly_train._transforms.transform import (
+    ChannelDropArgs,
+    CopyBlendArgs,
+    MixUpArgs,
+    MosaicArgs,
+    NormalizeArgs,
+    RandomFlipArgs,
+    RandomIoUCropArgs,
+    RandomPhotometricDistortArgs,
+    RandomRotate90Args,
+    RandomRotationArgs,
+    RandomZoomOutArgs,
+    ResizeArgs,
+    ScaleJitterArgs,
+)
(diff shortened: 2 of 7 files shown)


src/lightly_train/_transforms/ltdetr_transforms/instance_segmentation.py
```diff
@@ -0,0 +1,232 @@
...
+        image = input["image"]
+        bboxes = input["bboxes"]
+        class_labels = input["class_labels"]
+        binary_masks = input["binary_masks"]
+        indices = np.arange(len(bboxes), dtype=np.int64)
+
+        if self._should_apply_mosaic():
+            image, bboxes, class_labels, binary_masks_opt = self.mosaic(  # type: ignore[misc]
+                image, bboxes, class_labels, binary_masks
+            )
```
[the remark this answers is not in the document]
yutong-xiang-97: I raised an error instead.

src/lightly_train/_transforms/ltdetr_transforms/utils.py
```diff
@@ -134,13 +129,13 @@ def resolve_ltdetr_step_schedule_for_augmentation(
                 f"step_stop ({step_schedule.step_stop}) <= 0. "
                 "Disabling scale_jitter for this run."
             )
-            args.scale_jitter = None
```
gabrielfruet, line 137: Any reason why using seattr here instead of just setting the attribute?

src/lightly_train/_commands/train_task.py
```diff
@@ -1497,14 +1497,14 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
             model_name=config.model,
             val_transform_args=val_transform_args,
         )
-        train_collate_fn = train_dataset.batch_collate_fn_cls(
+        train_dataset.set_transform(train_transform)
+        val_dataset.set_transform(val_transform)
+        train_collate_fn = train_dataset.get_batch_collate_fn_cls()(
```
gabrielfruet, line 1502: Why we have this method instead of direct access to the batch_collate_fn_cls?

src/lightly_train/_transforms/ltdetr_transforms/instance_segmentation.py
```diff
@@ -0,0 +1,333 @@
...
+    scale_jitter: ScaleJitterArgs | None = None
+    resize: ResizeArgs | None
+    bbox_params: BboxParams | None
+    normalize: NormalizeArgs | Literal["auto"] | None
+
+    # Necessary for BboxParams, which are not serializable by pydantic.
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+
+class _MaskAwareStepScheduledCompose(StepScheduledCompose):
```
gabrielfruet, line 101: Why we don't just make `StepScheduledCompose` MaskAware?

gabrielfruet (commented):
In general LGTM, I just have some questions

src/lightly_train/_commands/train_task.py
yutong-xiang-97: Good point. It is because instance segmentation now should select the collate classes based on the model architecture, but I noticed that this particular implementations assumes EoMT as the default. I will make `get_batch_collate_fn_cls()` return EoMT or LT-DETR purely based on `self.transform` but not remove it.

src/lightly_train/_transforms/ltdetr_transforms/utils.py
yutong-xiang-97: In utils.py, `args` is typed as `LTDETRTransformArgs`, and that is a `Protocol `exposing `scale_jitter` as a read-only `@property`. This will make mypy reject `
args.scale_jitter = None` because the protocol does not define a setter. Instead, `setattr(args, "scale_jitter", None)` sidesteps that static typing limitation while still doing the runtime mutation on the concrete Pydantic model.

src/lightly_train/_transforms/ltdetr_transforms/instance_segmentation.py
yutong-xiang-97: `StepScheduledCompose` is currently the shared LT-DETR scheduling/cache component. Object detection uses it with image + bbox + class labels only, while instance segmentation must pass the masks through the compose and `ToTensorV2` must convert image and masks together.
Making the parent mask-aware would either add mask-specific behavior to the object-detection path, or require flags/conditionals inside the shared component. Neither makes sense for me.

gabrielfruet: Based on the class implementation, we would only need to add ToTensorV2() to the pipeline, which seems very reasonable to me. ToTensorV2() shouldn't affect the object detection path I think

yutong-xiang-97: Nope adding `ToTensorV2` to object detection would actually change output contract of the transform.

Object detection’s `StepScheduledCompose` returns numpy-style Albumentations outputs, and the object-detection collate keeps working with it while it applies batch-level augmentations. Tensor conversion happens only at the end of the collate path. If the parent became mask-aware by always adding `ToTensorV2`, object detection would start receiving torch tensors from the sample transform. That has a few consequences. Most notably, you will receive CHW tensors instead of HWC numpy arrays; also, the shape and dtype assumptions in mixup/copyblend/scale-jitter could change;

For instance segmentation, on the other hand, appending `ToTensorV2` at sample level is specifically needed because Albumentations converts image and masks together in one compose call, and the instance transform then indexes/stacks the returned masks as tensors.

gabrielfruet (approved):

src/lightly_train/_transforms/ltdetr_transforms/instance_segmentation.py
gabrielfruet: Makes sense!

PR #837, opened 2026-07-05
Add: dinov31 DINOv2 post-train method with PaKA dense-relational loss

IgorSusmelj:
### TL:DR

- Uses the new loss added to LightlySSL [Patch-level Kernel Alignment for Self-Supervised Dense Representation Learning
](https://arxiv.org/html/2509.05606v1)
- It works similar to Gram anchoring in DINOv3 with a few nice improvements. It not only enforces patch relation from the teacher but also slightly improves the representativeness. It's more robust to hyper parameters and bad teachers.
- Note, the final model quality heavily depends on the quality oft he pretraining. See DINOv31 as a post training method to improve the backbone for dense downstream tasks (object detection and segmentation). And only use this method for special data domains (where distillation does not work). 

### Extended information

DINOv31 continues a DINOv2-pretrained backbone with the full DINOv2 objective (DINO + iBOT + KoLeo) plus an auxiliary Patch Kernel Alignment (PaKA / CKA) loss that aligns the relational structure of student and teacher dense patch tokens. It is a thin subclass of DINOv2 and does not modify it; the PaKA term is added on top and the DINOv2 path is left byte-identical.

- DINOv31 method: PaKA projection heads (reusing dinov2 `_build_mlp`), ROI- aligned cross-view CKA on clean-teacher globals vs high-overlap student locals, EMA teacher head update, lenient checkpoint load (tolerates missing `*_paka_head*` keys), teacher-temperature warmup cap for short post-trains, and an `on_fit_start` override that suppresses DINOv2's 125k-step warning.
- DINOv31 transform: view layout [global0, global1, dino_local.., clean_global0/1, paka_local..] with geometry-tracked crops (clean globals reuse the global crop geometry; paka locals are true small locals nested inside alternating globals).
- Shared geometry plumbing (opt-in via `record_geometry`): TrackedRandomResized Crop / TrackedHorizontalFlip / TrackedVerticalFlip, ViewTransform records an 8-element geometry tensor, optional `geometries` on the transform output, dataset item and batch, and ImageDataset propagates it only when every view records one.
- Lightly SSL source pinned to the PatchKernelAlignmentLoss merge commit (lightly-ai/lightly#1953) until the loss is released.

> I'll first merge this PR and two more before updating the docs and changelog

## How has it been tested?

- Unit tests + extensive testing

Pretraining of DINOv2 for 300 epochs on COCO train followed by 25 epochs of posttraining with DINOv31 and a ViT-S backbone gave:
- 1.40% better linear segmentation on ADE20k (28.9% vs 27.5%)
- 0.55 % better full finetuning on ADE20k (41.6% vs 41.1%)

> Note: you could probably train longer and get better results. My goal was simply to check if the method works on a hard task (improve dense features solely based on self-supervised pretraining on COCO)

DINOv31 training loss curves (batch size 54, 25 epochs, 4x4090 GPUs, posttraining took 7h, 518x518 resolution):
image

Pretraining was also done on COCO (same dataset as posttraining) for 300 epochs. Took 15h. 224x224 pixels.

- [ ] Yes

- [ ] Yes


docs/source/pretrain_distill/methods/dinov31.md
```diff
@@ -0,0 +1,63 @@
+(methods-dinov31)=
+
+# DINOv31
+
+DINOv31 post-trains ("continues") a DINOv2-pretrained backbone with the full DINOv2
+objective (DINO + iBOT + KoLeo) plus an auxiliary Patch Kernel Alignment (PaKA / CKA)
+loss that aligns the relational structure of student and teacher dense patch tokens. It
+is a thin subclass of {ref}`methods-dinov2` and is aimed at improving dense
```
liopeer:
```suggestion
 is aimed at improving dense
```
This is an implementation detail that I would not mention in the user-facing documentation.

docs/source/pretrain_distill/methods/dinov31.md
```diff
@@ -0,0 +1,63 @@
...
+```python
+import lightly_train
+
+if __name__ == "__main__":
+    lightly_train.pretrain(
+        out="out/my_experiment",
+        data="my_data_dir",
+        model="dinov2/vits14",
+        method="dinov31",
+        checkpoint="dinov2_vits14.ckpt",  # DINOv2 init; paka_head keys tolerated
```
liopeer: The checkpoint is the one with optimizer state etc. The pure weights are in `exported/exported_best.pt`. Should we not use that one? See [https://docs.lightly.ai/train/stable/pretrain_distill/index.html#out](https://docs.lightly.ai/train/stable/pretrain_distill/index.html#out).

src/lightly_train/_transforms/view_transform.py
```diff
@@ -53,11 +56,12 @@ class ViewTransformArgs(PydanticConfig):
     gaussian_blur: GaussianBlurArgs | None
     solarize: SolarizeArgs | None
     normalize: NormalizeArgs
+    # Record per-view crop/flip geometry and attach it to the output view as a
+    # "geometry" tensor, used by dense-relational losses (e.g. dinov31).
+    record_geometry: bool = False
```
liopeer: Nit: We generally prefer keeping `ViewTransformArgs` abstract and setting the values only inside the `_methods` subpackage, so that the whole method configuration is close together.

uv.lock
```diff
@@ -386,9 +386,6 @@ conflicts = [[
     { package = "lightly-train", group = "pinned-torch-minimal" },
 ]]
 
-[options]
-exclude-newer = "2026-05-18T22:00:00Z"
```
liopeer: Would be better to lock with `make lock`, so that we keep the `exclude-newer` flag, which keeps CI stable.

pyproject.toml
```diff
@@ -262,6 +262,12 @@ conflicts = [
 ]
 
 [tool.uv.sources]
+# LightlySSL from the PatchKernelAlignmentLoss merge commit
+# (https://github.com/lightly-ai/lightly/pull/1953).
+# TODO(Igor, 07/2026): Remove this git override once a lightly release contains
+# PatchKernelAlignmentLoss and roi_resample_to_grid, so the plain "lightly"
+# dependency on PyPI suffices.
+lightly = { git = "https://github.com/lightly-ai/lightly.git", rev = "f03d012df6486e885dac5d6c2e14780b6455d9cb" }
```
liopeer, line 271: This does not help with distribution on PyPi, only for local development and CI etc. Is this deliberate? Otherwise you'd instead have to instead directly use `lightly @ git+https://github.com/lightly-ai/lightly.git@f03d012...` in the `dependencies` further above.

src/lightly_train/_commands/train_helpers.py
```diff
@@ -421,18 +421,75 @@ def load_checkpoint(
         )
 
 
+def _interpolate_pos_embed(
```
liopeer: I think pretty much all the changes to this module belong inside `src/lightly_train/_models/dinov2_vit` since they are not generic, but very specific to DINOv2 ViTs. I think you could e.g. register a `load_state_dict` pre-hook on the `DINOv2ViTModelWrapper` or the `DinoVisionTransformer`.

docs/source/pretrain_distill/methods/dinov31.md
IgorSusmelj: I think we need to tackle this in a follow up. Loading from checkpoint has been tested. For exported I did not try.

liopeer: I think you are actually be correct and it indeed requires `checkpoint`, not the exported weights. One of the inconsistencies between the pretraining and fine-tuning pipelines.

liopeer (approved):
LGTM!

PR #838, opened 2026-07-05
Add: adamw8bit optimizer support for lower memory usage

IgorSusmelj:
## What has changed and why?

  - Adds a new 8-bit AdamW optimizer option (`optim="adamw8bit"`), which stores the optimizer's internal state in a much smaller numeric format. The dependency (`bitsandbytes`) stays optional.
  - In my tests I was able to increase the batch size by 2-3x compared to regular adam

## How has it been tested?

- new unit tests
- trained a model with adam8bit vs adam (only tested with DINOv31)

- [ ] Yes

- [ ] Yes


src/lightly_train/_optim/adamw8bit_args.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def get_optimizer(self, params: ParamsT, lr_scale: float) -> TorchOptimizer:
+        try:
+            from bitsandbytes.optim import (  # type: ignore[attr-defined]
+                AdamW8bit,
+            )
+        except ImportError as err:  # pragma: no cover
+            raise ImportError(
+                "optim_type='adamw8bit' requires the 'bitsandbytes' package. "
+                "Install it with `pip install bitsandbytes`."
```
liopeer, line 46:
```suggestion
                "Install it with `pip install lightly-train[bitsandbytes]`."
```

src/lightly_train/_methods/dinov31/dinov31.py
```diff
@@ -58,15 +58,32 @@
...
 from lightly_train._methods.method import TrainingStepResult
+from lightly_train._optim.adamw8bit_args import AdamW8bitArgs
+from lightly_train._optim.optimizer_args import OptimizerArgs
+from lightly_train._optim.optimizer_type import OptimizerType
 from lightly_train._optim.trainable_modules import TrainableModules
 from lightly_train._torch_helpers import update_momentum
 from lightly_train.types import Batch
 
 
+class DINOv31AdamW8bitArgs(DINOv2AdamWViTArgs, AdamW8bitArgs):
```
liopeer, line 78: Nit: Would prefer keeping it single-inheritance (only `AdamW8bitArgs`) and copying these few values over from DINOv2.

liopeer (approved):
LGTM with optional comments!

src/lightly_train/_methods/dinov31/dinov31.py
IgorSusmelj: I decided against it for now to avoid duplicating the or adjusting the other parameters like lr etc.) But we can look into this in a follow up.

PR #846, opened 2026-07-08
Refactor Image Processing for Depth Estimation

mrpositron:
## What has changed and why?

Consolidated the depth-estimation image preprocessing. Previously there were two parallel per-image functions (`process_image_dav2` / `process_image_dav3`) backed by a large amount of custom resize code written to reproduce OpenCV's `INTER_AREA`/`INTER_CUBIC` output pixel-for-pixel. That cv2-bit-exactness goal has been abandoned, so the custom machinery is no longer needed.

- **Single preprocessing path**: `process_image_dav2` + `process_image_dav3` → one `process_image`. Dropped the cv2-parity resize (`_area_weights`,` _cubic_weights`, `_dav2_target_size`, etc.); resizing now uses `torchvision.transforms.v2.functional.resize` (bilinear + antialias), matching every other task model.
- **Resize method is now a call-time parameter** on `predict` / `predict_batch` (`process_res_method`, default `"square_resize"`) instead of a fixed per-model setting. Removed the now-dead `preprocess` config key and `_PROCESS_RES_METHOD_DAV3` constant.
- Added `square_resize` (resize to `image_size × image_size`) alongside upper_bound_resize and lower_bound_resize; all three now work for both DAv2 and DAv3.

Note: the default processing geometry changes (was upper-bound for DAv3, lower-bound for DAv2; now `square_resize` for both). Callers can pass `process_res_method=` to select the old behavior.

## How has it been tested?

- Rewrote `test_image_utils.py`: removed the cv2-parity tests, added coverage for all three resize methods, dtype/channel handling, and process_batch.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_task_models/depth_estimation_components/image_utils.py
+++ b/src/lightly_train/_task_models/depth_estimation_components/image_utils.py
@@ -16,14 +16,20 @@
 
 """Image preprocessing for the Depth Anything depth-estimation models.
 
-Split into a per-image stage (one function per model family) and a shared batch stage,
-so the per-image stage can be baked into an export. Output order matches input order.
+Split into a per-image stage (``process_image``) and a shared batch stage
+(``process_batch``), so the per-image stage can be baked into an export. Output order
+matches input order.
 
-Per-image stage, each returning a float RGB tensor in [0, 255] of shape ``(3, H, W)``:
-  - ``process_image_dav3``: boundary resize, then round each side to a multiple of
-    ``PATCH_SIZE``. Float32, rounding to the integer grid after every resize.
-  - ``process_image_dav2``: a single cubic resize so the shorter side reaches the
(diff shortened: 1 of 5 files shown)


src/lightly_train/_task_models/depth_estimation_components/image_utils.py
```diff
@@ -47,26 +53,31 @@
...
 
 
-def process_image_dav3(
+def process_image(
     img: Tensor,
     *,
-    process_res: int = 504,
-    process_res_method: str = "upper_bound_resize",
+    process_res: int,
+    process_res_method: str = "square_resize",
```
liopeer: Would type with a Literal. Same thing in the other file.

liopeer (approved):
LGTM, with a small optional remark!

PR #845, opened 2026-07-07
Remove legacy DINOv2 LT-DETR model

liopeer:
## What has changed and why?
- completely remove the legacy implementation
- re-route everything towards the generic model

## How has it been tested?
- unit tests

## Did you update [CHANGELOG.md](../CHANGELOG.md)?

- [x] Yes
- [ ] Not needed (internal change)

## Did you update the documentation?

- [x] Yes
- [ ] Not needed (internal change without effects for user)

PR #850, opened 2026-07-09
Fix normalization with legacy LT-DETR models

liopeer:
## What has changed and why?
- stacks on top of #845 
- old checkpoints from LT-DETR used [0,1] normalization, but at some point we changed the fallback to ImageNet, which caused degraded performance in the benchmarks
- this PR fixes the fallback

**Note**: Does not influence new checkpoints trained with ImageNet normalization, since those will include `normalize_args` inside the checkpoint arguments.

## How has it been tested?
- Ran the OD benchmark command for all public `-coco` checkpoints, and they all got the expected mAP number

- [x] Yes

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,9 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- Fixed an issue with legacy LT-DETR checkpoints that used a [0,1] normalization instead
+  of the now-default ImageNet normalization.
+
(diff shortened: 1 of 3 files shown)


gabrielfruet (approved):
LGTM! Why the name of the PR is fix for ConvNext models and we don't have mention/branching/testing for the convnext case?

liopeer:
@gabrielfruet I changed the title. The issue was with all legacy checkpoints, but I encountered it first with the convnext models, which is why i called it that way.

PR #852, opened 2026-07-09
Refactor YAML loading for datasets

liopeer:
## What has changed and why?
- yaml loading behaviour was untested and paths were not resolved with respect to the yaml file, but with respect to cwd
- also there were a lot of inconsistencies about yaml loading between docstrings, docs and the actual implementations

## How has it been tested?
- unit tests

- [x] Yes

- [x] Yes


src/lightly_train/_commands/benchmark_task.py
yutong-xiang-97, line 145: `resolve_data_paths` is only called in `_train_task_from_config`. `BenchmarkObjectDetectionConfig` loads the YAML (so `data_config_file` gets populated) but never calls `resolve_data_paths`, so relative paths in a benchmark data YAML still resolve relative to **cwd**, not the config file.

src/lightly_train/_commands/data_helpers.py
```diff
@@ -32,22 +34,56 @@ def load_data_yaml_if_path(value: Any, data_annotation: Any) -> Any:
             ``cls.model_fields["data"].annotation``.
     """
     if isinstance(value, (str, Path)):
+        data_config_file = Path(value).resolve()
```
yutong-xiang-97, line 37: Remote YAML paths supported by `fsspec`, such as `s3://...` are converted with `Path.resolve()`. This corrupts them into local paths, so relative dataset paths resolve incorrectly.

yutong-xiang-97 (approved):

PR #855, opened 2026-07-10
Fix uncalled `Tensor.is_floating_point` in ONNX export verification

yutong-xiang-97:
## What has changed and why?

The ONNX export verification in the task models checked `tensor.is_floating_point` (the method object) instead of calling it `tensor.is_floating_point()`. A bound method is always truthy, so the check always took the float branch:

- Integer outputs (e.g. `labels`) were force-cast to float and compared with `torch.testing.assert_close` (tolerance-based) instead of the intended exact-match `assert_most_equal` path — the discrete-output check was dead code.

This fixes the call in all 9 task models that have an `export_onnx` verification block (EoMT semantic/instance/panoptic for DINOv2 and DINOv3, PicoDet and LTDETR object detection, and image classification). The correct `dtype.is_floating_point` usages (a genuine property) were left untouched.

## How has it been tested?

Static change only. Verified there are no remaining uncalled `Tensor.is_floating_point` references via grep, and that the correct `dtype.is_floating_point` property usages were not touched. No behavioral change other than the verification path now selecting the correct comparison for integer vs float outputs.

- [x] Yes

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- Fix ONNX export verification for task models: `Tensor.is_floating_point` was
+  referenced without calling it, so the always-truthy bound method forced integer
+  outputs (e.g. labels) through the float comparison path instead of the intended
+  exact-match check.
+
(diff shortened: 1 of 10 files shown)


liopeer (approved):
LGTM!

PR #857, opened 2026-07-10
Add model registry for eomt instance segmentation

liopeer:
## What has changed and why?
- adds `config.py` files with model registries to the DINOv2 and DINOv3 EoMT instance segmentation models

## How has it been tested?
- unit tests

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_task_models/dinov2_eomt_instance_segmentation/config.py
+++ b/src/lightly_train/_task_models/dinov2_eomt_instance_segmentation/config.py
@@ -1,0 +1,66 @@
+#
+# Copyright (c) Lightly AG and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+#
+from __future__ import annotations
+
+from lightly_train._configs.config import ConfigsNamespace, PydanticConfig
+from lightly_train._configs.model_registry import ModelRegistry
+
+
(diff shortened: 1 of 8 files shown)


src/lightly_train/_task_models/dinov2_eomt_instance_segmentation/config.py
```diff
@@ -0,0 +1,66 @@
...
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+#
+from __future__ import annotations
+
+from lightly_train._configs.config import ConfigsNamespace, PydanticConfig
+from lightly_train._configs.model_registry import ModelRegistry
+
+
+class EoMTInstanceSegmentationConfig(PydanticConfig):
```
yutong-xiang-97, line 14: Will DINOv2 and DINOv3 in the end be merged into one single module?

yutong-xiang-97 (approved):
LGTM. Left a question.

src/lightly_train/_task_models/dinov2_eomt_instance_segmentation/config.py
liopeer: No, I don't think we should do that. A lot of effort but very little reward imo. Should only be considered once we want to support more backbones than DINOv2/DINOv3.

PR #859, opened 2026-07-10
Model registry for EoMT Semantic Segmentation

liopeer:
## What has changed and why?
- adds model registry for semantic segmentation eomt models

**Note**: Breaking change since we remove the option to pass the patch size via the training command. I don't think anyone was using this and it was a really poor architectural choice to have the patch size in the `train_model.py`, because it has 0 influence on the training recip.

## How has it been tested?
- unit tests

- [x] Yes

- [x] Yes

### DIFF
--- a/src/lightly_train/_task_models/dinov3_eomt_semantic_segmentation/task_model.py
+++ b/src/lightly_train/_task_models/dinov3_eomt_semantic_segmentation/task_model.py
@@ -114,8 +114,10 @@ class DINOv3EoMTSemanticSegmentation(TaskModel):
             locals(), ignore_args={"backbone_weights", "backbone_url", "load_weights"}
         )
         config = self._get_config(model_name=model_name)
-        parsed_name = self._resolve_model_name(model_name=model_name)
-        self.model_name = parsed_name["model_name"]
+        package_name, backbone_name = package_helpers.parse_model_name(
+            config.backbone_name
+        )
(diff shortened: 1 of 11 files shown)


src/lightly_train/_task_models/dinov3_eomt_semantic_segmentation/task_model.py
```diff
@@ -109,7 +113,8 @@ def __init__(
         super().__init__(
             locals(), ignore_args={"backbone_weights", "backbone_url", "load_weights"}
         )
-        parsed_name = self.parse_model_name(model_name=model_name)
+        config = self._get_config(model_name=model_name)
+        parsed_name = self._resolve_model_name(model_name=model_name)
```
mrpositron: `_resolve_model_name` re-invokes `_get_config`, so the registry lookup runs twice per construction. Have `_resolve_model_name` take the already-resolved config, or derive `parsed_name` from config directly. DINOv2 avoids this because it only calls `_resolve_model_name`.

mrpositron (commented):
The PR review marks `CHANGELOG.md` as updated, but this PR does not include any changes to it.

src/lightly_train/_task_models/dinov3_eomt_semantic_segmentation/task_model.py
liopeer: I made it now much simpler, with fewer methods.

mrpositron (approved):
LGTM 🚀

PR #861, opened 2026-07-11
Add ONNX and TensorRT export for LT-DETR instance segmentation

yutong-xiang-97:
## What has changed and why?

Add ONNX and TensorRT export support for LT-DETR instance segmentation.

Note: it now exports both `images` and `orig_target_size` inputs so predicted boxes and masks can be scaled to the caller-provided image size - something listed as TODO in object detection. The TensorRT helper now configures matching optimization profiles for additional inputs with dynamic batch dimensions, such as `orig_target_size`, while rejecting unsupported dynamic non-batch dimensions.

## How has it been tested?

1. Unit tests covering
- static and dynamic batch export
- FP16
- simplification
- opset selection (ranging from 16 to 20 on CI)
- model aliases and non-square inputs
- prediction parity

3. A vibe-coded verification script that checks:
- Both inputs have a dynamic batch dimension.
- TensorRT accepts configured batch sizes, resolves output shapes, and produces finite outputs with the expected dimensions.
- TensorRT and ONNX Runtime produce compatible detections for identical random inputs.
Detections are aligned order-independently using labels, boxes, and scores.
- Boxes and scores are numerically close after alignment.
- Labels match exactly.
- Raw mask-logit error statistics are reported.
- Thresholded masks are compared using pixel agreement, IoU, and agreement away from the zero-logit boundary.

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_export/tensorrt_helpers.py
+++ b/src/lightly_train/_export/tensorrt_helpers.py
@@ -50,6 +50,9 @@ def export_tensorrt(
     This loads the ONNX file, parses it with TensorRT, infers the static input
     shape (C, H, W) from the `"images"` input, and creates an engine with a
     dynamic batch dimension in the range `[min_batchsize, opt_batchsize, max_batchsize]`.
+    Any additional inputs (e.g. `"orig_target_size"`) that have a dynamic batch
+    dimension get a matching optimization profile over the same batch range; a
+    dynamic non-batch dimension on such an input is not supported and raises.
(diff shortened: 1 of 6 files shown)


liopeer (approved):
LGTM! In the future we should not have `orig_target_size` in the exported model anyway. The model graph should only forward the image and yield class logits and box coordinates, working on this in #860, but this will take some time (and likely a design doc as well).

PR #866, opened 2026-07-14
Use LT-DETRv2 models in task training tests

yutong-xiang-97:
## What has changed and why?

Use lightweight, non-pretrained LT-DETRv2 presets in the object-detection and instance-segmentation training smoke tests.

This adds internal ECViT-Tiny-backed model aliases for both tasks, keeps those aliases out of the public model-name listings, and lets the ECViT package resolve an internal non-pretrained preset without downloading weights. The task training tests now exercise the LT-DETRv2 object-detection and instance-segmentation paths directly instead of using unrelated DINOv3/EoMT models.

## How has it been tested?

- `pytest tests/_commands/test_train_task.py::test_train_object_detection_yolo tests/_commands/test_train_task.py::test_train_instance_segmentation -q` (`2 passed`)
- `make static-checks`

- [ ] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_models/ecvit/ecvit_package.py
+++ b/src/lightly_train/_models/ecvit/ecvit_package.py
@@ -15,7 +15,6 @@ import torch
 
 from lightly_train._env import Env
 from lightly_train._models.ecvit.ecvit import (
-    ECVIT_PRESETS,
     ECVIT_PRETRAINED_URLS,
     ECViTModelWrapper,
 )
@@ -121,11 +120,6 @@ class EdgeCrafterPackage(Package):
         model_name = cls.parse_model_name(model_name=model_name)
         model_info = MODEL_NAME_TO_INFO[model_name]
         preset_name = model_info["preset_name"]
-        if preset_name not in ECVIT_PRESETS:
-            raise ValueError(
-                f"Unknown ECViT preset: '{preset_name}'. Available presets: "
-                f"{list(ECVIT_PRESETS)}."
-            )
 
         weights_path: PathLike | None = None
         if load_weights and model_info["default_weights"] is not None:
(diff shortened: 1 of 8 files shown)


src/lightly_train/_models/ecvit/ecvit_package.py
```diff
@@ -107,18 +118,19 @@ def get_model(
                 f"num_input_channels={num_input_channels}."
             )
 
-        preset_name = cls.parse_model_name(model_name=model_name)
+        model_name = cls.parse_model_name(model_name=model_name)
+        model_info = MODEL_NAME_TO_INFO[model_name]
+        preset_name = model_info["preset_name"]
         if preset_name not in ECVIT_PRESETS:
```
liopeer: Now we're basically have 2 truths for which presets exist. I would consider getting rid of the `ECVIT_PRESETS` import and only use `MODEL_NAME_TO_INFO`, but add a unit test to make sure that they never diverge.

src/lightly_train/_task_models/ltdetr_instance_segmentation/task_model.py
```diff
@@ -182,7 +182,11 @@ def is_supported_model(cls, model: str) -> bool:
 
     @classmethod
     def list_model_names(cls) -> list[str]:
-        return list(LTDETR_SEG_MODEL_REGISTRY.list_aliases())
+        return [
```
liopeer: This is not a public method, so I would prefer keeping it the same way as with the other models. Also, this is again basically a "string parsing logic", which is exactly what the model registry wants to do away with. If we want some models to be not listed by `list_model_names`, we should imo extend the `ModelRegistry` base class with some attributes/logic to handle that.

liopeer: Same for the object detection model.

liopeer (changes requested):

liopeer (approved):

PR #869, opened 2026-07-15
feat(ltdetr): enforce 4 px min bbox guard in train transforms (TRN-2312)

gabrielfruet:
## What has changed and why?

Enforce a minimum bbox size of **4 pixels (w and h)** on the LT-DETR training codepath so that degenerate targets don't destabilize training.

The guard is a new `filter_normalized_cxcywh_min_size(bboxes, *, image_size, min_size_px)` helper in `ltdetr_transforms/utils.py`. It returns a positional boolean keep-mask for YOLO `(cx, cy, w, h)` boxes whose pixel-side width/height falls below `min_size_px` (disabled when `min_size_px <= 0`, which returns an all-`True` mask).

It's wired in via a `min_bbox_size_px: float` field on the transform args, defaulting to `0.0` on the shared base classes (`LTDETRObjectDetectionTransformArgs`, `LTDETRInstanceSegmentationTransformArgs`) and only set to `4.0` on the LT-DETR *training* transform-args subclasses — validation targets, PicoDet targets, and DINOv2 val variants are left unfiltered so eval metrics aren't affected.

Filtering happens exclusively in the collate functions (`LTDETRObjectDetectionCollateFunction`, `LTDETRInstanceSegmentationCollateFunction`), after all batch-level augmentations (mixup, copyblend, scale-jitter, `ToTensorV2`) are applied and images are in their final tensor form — that's the last point a box's true final size is known, since a later batch-level scale-jitter can shrink a box that already passed an earlier per-sample check. The per-sample `Transform.__call__` methods no longer filter at all. Because collate operates on per-sample lists of tensors, the boolean mask is applied positionally to bboxes, classes, and (for instance segmentation) masks in lockstep — no index bookkeeping needed.

This landed after an earlier attempt that put the guard in the matcher/criterion (via `sanitize_boxes_cxcywh_normalized`, threading `image_size` through `criterion.forward`) and through `train_model.py`. That approach was reverted in favor of filtering at the transform/collate level, which doesn't require plumbing image size into the loss/matcher code at all. It also earlier reverted an Albumentations `BboxParams` min-size change, since Albumentations applies that per-transform against the intermediate (pre-resize) image size — the wrong reference for a post-resize guard. A subsequent review pass (moving all filtering into collate) also removed a mask re-indexing bug in the instance-segmentation transform, where a second filter call re-indexed already-compacted masks using stale indices.

Fixes TRN-2312.

## How has it been tested?

Unit tests:

- `tests/_transforms/ltdetr_transforms/test_utils.py`: `TestFilterNormalizedCxcywhMinSize` covers the boolean-mask interface (drops sub-minimum boxes, `min_size_px<=0` keeps all, empty input, wrong-shape input raises).
- `tests/_transforms/ltdetr_transforms/test_object_detection.py`: `TestObjectDetectionTransformBboxFilter` asserts the transform no longer filters; `TestObjectDetectionCollateBboxFilter` covers filtering after resize and after scale-jitter; `TestMinBboxSizePxDefaults` asserts the base/train/val defaults.
- `tests/_transforms/ltdetr_transforms/test_instance_segmentation.py`: mirrors the above, plus `TestLTDETRInstanceSegmentationCollateBboxFilter` verifying boxes *and masks* stay aligned after collate-level filtering.

Local runs:

```
uv run --frozen pytest tests/_transforms/ltdetr_transforms \
                       tests/_task_models/ltdetr_object_detection \
                       tests/_task_models/ltdetr_instance_segmentation \
                       tests/_task_models/picodet_object_detection
uv run --frozen mypy src/lightly_train/_transforms/ltdetr_transforms
```

`make format` ran clean.


src/lightly_train/_transforms/ltdetr_transforms/instance_segmentation.py
```diff
@@ -93,6 +94,7 @@ class LTDETRInstanceSegmentationTransformArgs(TaskTransformArgs):
     resize: ResizeArgs | None
     bbox_params: BboxParams | None
     normalize: NormalizeArgs | Literal["auto"] | None
+    min_bbox_size_px: float = 4.0
```
yutong-xiang-97: `min_bbox_size_px=4.0` is defined on the shared base transform arguments, so it also filters validation targets and PicoDet targets—not just LT-DETR training. Filtering validation ground truth changes evaluation metrics. I would default the shared base to 0.0, then set 4.0 explicitly only on the LT-DETR training argument classes.

src/lightly_train/_transforms/ltdetr_transforms/instance_segmentation.py
```diff
@@ -234,9 +242,24 @@ def __call__(
         else:
             binary_masks_out = torch.stack(kept_masks).to(dtype=torch.int)
 
+        # Drop boxes whose final pixel-side width/height is below the configured
+        # threshold; re-index masks with the same filtered indices so the three
+        # arrays stay aligned.
+        bboxes_out, class_labels_out, indices_out = filter_boxes_below_min_size(
```
yutong-xiang-97: Instance-segmentation mask indexing can become incorrect or raise `IndexError`. Here, `indices_out` still contains indices into Albumentations’ original mask list, while `binary_masks_out` has already been compacted. Using the former to index the latter is only safe when Albumentations has not previously removed any boxes.

The minimum-size filter should return a positional boolean mask, or filtering should happen before `binary_masks_out` is constructed. Then apply the same positional mask to boxes, labels, and masks.

See my suggestions in `utils.py` for the interface design of the helper function.

src/lightly_train/_transforms/ltdetr_transforms/utils.py
```diff
@@ -303,3 +303,51 @@ def filter_degenerate_yolo_boxes(
...
+def filter_boxes_below_min_size(
+    bboxes: NDArrayBBoxes,
+    class_labels: NDArrayClasses,
+    image_size: tuple[int, int],
+    min_size_px: float = ...,
+    indices: None = ...,
+) -> tuple[NDArrayBBoxes, NDArrayClasses, None]: ...
+
+
+def filter_boxes_below_min_size(
```
yutong-xiang-97: This only applies to normalized cxcywh bboxes in YOLO format, I would use something like `filter_normalized_cxcywh_min_size` to denote this should this be used elsewhere later on.

Also with the changes mentioned above, the interface can be simplified to something like:

```python
def filter_normalized_cxcywh_min_size(
    bboxes: Tensor,
    *,
    image_size: tuple[int, int],
    min_size_px: float,
) -> Tensor:
    """Return boxes whose normalized width and height meet a pixel threshold."""
    if bboxes.ndim != 2 or bboxes.shape[1] != 4:
        raise ValueError(f"Expected bboxes with shape (N, 4), got {bboxes.shape}.")

    if min_size_px <= 0.0:
        return torch.ones(
            len(bboxes),
            dtype=torch.bool,
            device=bboxes.device,
        )

    height, width = image_size
    width_px = bboxes[:, 2] * width
    height_px = bboxes[:, 3] * height

    return (width_px >= min_size_px) & (height_px >= min_size_px)
```

This way you also don't need to consider mask logic for instance segmentation and can thus apply the same logic to instance segmentation collate function.

src/lightly_train/_transforms/ltdetr_transforms/object_detection.py
```diff
@@ -96,6 +97,7 @@ class LTDETRObjectDetectionTransformArgs(TaskTransformArgs):
     resize: ResizeArgs | None
     bbox_params: BboxParams | None
     normalize: NormalizeArgs | Literal["auto"] | None
+    min_bbox_size_px: float = 4.0
```
yutong-xiang-97: Same here.

src/lightly_train/_transforms/ltdetr_transforms/object_detection.py
```diff
@@ -158,8 +160,23 @@ def __call__(
             transformed["bboxes"], transformed["class_labels"]
         )
 
+        # Drop boxes whose final pixel-side width/height is below the configured
```
yutong-xiang-97: I really do not think the `transform.__call__` is the right places to do the filter. This way, a later scale jitter can shrink a previously valid 5-pixel box to 2 pixels and you have to do another filtering in the collate function anyway. 

See below for my suggestions to do all the filtering only in the collate function's `__call__`.

src/lightly_train/_transforms/ltdetr_transforms/object_detection.py
```diff
@@ -343,6 +360,20 @@ def __call__(self, batch: list[ObjectDetectionDatasetItem]) -> ObjectDetectionBa
             if isinstance(item["class_labels"], list):
                 item["class_labels"] = np.array(item["class_labels"])
 
+            # Drop boxes made too small by the batch-level transforms (mixup,
```
yutong-xiang-97: Instead of doing a second filtering, I would do all the filtering only in the collate function's `__call__`:

```python
# add these lines right before return
height, width = image.shape[-2:]
for index, (item_bboxes, item_classes) in enumerate(zip(bboxes, classes)):
    keep = filter_normalized_cxcywh_min_size(
        item_bboxes,
        image_size=(height, width),
        min_size_px=self.transform_args.min_bbox_size_px,
    )
    bboxes[index] = item_bboxes[keep]
    classes[index] = item_classes[keep]

return ObjectDetectionBatch(...)
```

This way you also don't need `indices` and function overload in the helper function which simplifies the code.

src/lightly_train/_transforms/ltdetr_transforms/instance_segmentation.py
yutong-xiang-97, line 325: Here you can add something like:

```python
filtered_bboxes: list[Tensor] = []
filtered_classes: list[Tensor] = []
filtered_masks: list[Tensor] = []

for image, item_bboxes, item_classes, item_masks in zip(
    images,
    bboxes,
    classes,
    masks,
):
    height, width = image.shape[-2:]
    keep = normalized_cxcywh_min_size_keep_mask(
        item_bboxes,
        image_size=(height, width),
        min_size_px=self.transform_args.min_bbox_size_px,
    )

    filtered_bboxes.append(item_bboxes[keep])
    filtered_classes.append(item_classes[keep])
    filtered_masks.append(item_masks[keep])

bboxes = filtered_bboxes
classes = filtered_classes
masks = filtered_masks
```

to remove the filtering from per-sample transform.

yutong-xiang-97, line 347: Also just realized that it is not using `InstanceSegmentationBatch`. Sorry for that!

yutong-xiang-97 (changes requested):
Generally in a right direction. However, I think the placement of the filtering can be instead moved to the collate function to simplify the logic.

src/lightly_train/_transforms/ltdetr_transforms/object_detection.py
gabrielfruet: Makes much more sense!

yutong-xiang-97 (approved):

PR #880, opened 2026-07-17
Fixing Linear Semantic Segmentation Bugs

mrpositron:
## What has changed and why?

Fixes two bugs in linear semantic segmentation with DINOv3 backbones:
1. **DINOv3 backbone crash on model creation.** 

DINOv3 builders hardcode `drop_path_rate` internally and re-splat `**kwargs`, so forwarding it raised `got multiple values for keyword argument 'drop_path_rate'`. No DINOv3 builder accepts the argument, so `dinov3_package.py` now drops it before building the model (a no-op for the model).

2. **Out-of-memory errors during validation on wide or tall images.**

Validation previously tiled every image and forwarded all generated crops in a single batch while retaining every crop’s full-resolution logits. Because the number of crops grows with the image aspect ratio, memory usage was unbounded for very wide or tall images.

Validation now:
- Collects crops from all images in the validation batch.
- Processes them in chunks bounded by the dataloader batch size.
- Pools crops across images so forward batches remain well utilized.
- Immediately scatters each crop’s logits into per-image sum/count accumulators.
- Averages overlapping regions without retaining logits for every crop.
- Keeps only the first `viz_max_images` predictions on CPU for visualization.

This bounds forward activation and crop-logit memory while preserving correct reconstruction across overlapping crops.

## How has it been tested?

- Added `test_dinov3_package.py` asserting `drop_path_rate` is dropped and the model builds.
- Added `test_train_model.py` covering the per-image validation path (loss/metrics accumulation, chunked forward, and `viz_max_images` capping).

- [ ] Yes

- [ ] Yes


tests/_task_models/linear_semantic_segmentation/test_train_model.py
```diff
@@ -0,0 +1,193 @@
+#
```
yutong-xiang-97: I would drop `test_train_model.py` altogther. We never had it before and you would need either mocking (not recommended) or real fabric instance to test it properly, which is cumbersome.

yutong-xiang-97 (approved):

PR #878, opened 2026-07-17
Move LT-DETR export to torch dynamo

liopeer:
## What has changed and why?
- move LT-DETR exporting to torch dynamo
- define model I/O spec
- bake licensing and normalization params into ONNX export

## How has it been tested?
- unit tests
- locally ran benchmark command

## Did you update [CHANGELOG.md](../CHANGELOG.md)?

- [ ] Yes
- [x] Not needed (internal change)

## Did you update the documentation?

- [ ] Yes
- [x] Not needed (internal change without effects for user)

PR #888, opened 2026-07-20
Move exports to mixin classes

liopeer:
## What has changed and why?
- stacks on top of #878

## How has it been tested?
- unit tests
- local run of benchmark command across torch, onnx and tensorrt

- [ ] Yes

- [ ] Yes


src/lightly_train/_export/export_onnx.py
```diff
@@ -0,0 +1,253 @@
...
+        )
+        dynamic_shapes = self.model_input_spec.dynamic_shapes(
+            dynamic_batch_size=dynamic_batch_size
+        )
+        with torch.no_grad():
+            example_output = module(**example_inputs)
+        output_names = [field.name for field in fields(example_output)]
+
+        logger.info(f"Exporting ONNX model to '{out}'")
+        torch.onnx.export(
```
gabrielfruet, line 138: Since we have to specify everything again anyway for the `torch.onnx.export`, why don't we just use the `torch.onnx.export` and omit the self.export call?

src/lightly_train/_task_models/task_model.py
```diff
@@ -82,6 +83,15 @@ def class_path(self) -> str:
         """
         return f"{self.__module__}.{self.__class__.__name__}"
 
+    @property
+    def is_deploy_mode(self) -> bool:
+        """Whether the model has been transformed for inference deployment."""
+        return False
+
+    def deploy(self) -> Self:
```
gabrielfruet, line 91: This was already in place and we just specified it in the TaskModlel now, right?

src/lightly_train/_export/export_onnx.py
```diff
@@ -0,0 +1,253 @@
...
+        out: PathLike,
+        *,
+        precision: Literal["fp32", "fp16"] = "fp32",
+        batch_size: int = 1,
+        dynamic_batch_size: bool = True,
+        opset_version: int | None = None,
+        simplify: bool = True,
+        verify: bool = True,
+        format_args: dict[str, Any] | None = None,
+        shape_overrides: dict[str, tuple[int | None, ...]] | None = None,
```
gabrielfruet, line 90: I'm missing the docs here explaining what each parameters does. For example, shape_overrides is not 100% clear to me

src/lightly_train/_task_models/ltdetr_object_detection/task_model.py
```diff
@@ -629,233 +645,33 @@ def export_onnx(
...
-        dimension and is traced with batch size 2; otherwise it uses ``batch_size``.
-        ``num_channels`` optionally overrides the declared image-channel count.
-
-        The exported graph returns raw class logits and normalized ``cxcywh`` boxes.
-        Image preprocessing, top-k selection, thresholding, box rescaling, and SAHI
-        merging are intentionally kept outside the graph.
-
-        Optionally simplifies the exported model in-place using onnxslim and
-        verifies numerical closeness against a float32 CPU reference via ONNX Runtime.
+        """Export raw LT-DETR predictions to ONNX.
```
gabrielfruet: I also feel like a clearer docstring for this method is benefitial.

gabrielfruet (approved):
Approved! Just some questions and minor documentation stuff.

src/lightly_train/_export/export_onnx.py
liopeer: 100% agree. I'll add them as soon as it becomes public (at the moment this is being called from inside `LTDETRObjectDetection.export_onnx`). I am also tbh not fully happy yet with the `shape_overrides`, which is why I did not make it the public-facing one just yet. 

Essentially it's meant to make the image tensor's shape flexible for the export – this was a TODO and currently you can change the number of channels. I think it could make sense that one would want to export the model with a different image size than it was trained, but e.g. overriding the channels does not make sense to me at all, since the model won't work anymore. But let me know your opinions as well on how the public API should look like in the future.

src/lightly_train/_export/export_onnx.py
liopeer: The goal is that `self.export` produces a re-usable `ExportedProgram` that we can use in other pipelines such as MIGraphX or ExecuTorch (I think OpenVINO will also take ONNX as the starting point).

But you are right, the `torch.onnx.export` would be completely enough, I just wanted to demonstrate here that this is compositional. In fact, `torch.onnx.export` would just call `torch.export` underneath if we passed a `nn.Module` instead of an `ExportedProgram`.

src/lightly_train/_task_models/ltdetr_object_detection/task_model.py
liopeer: Thanks for pointing that out. Here we should keep the full docstring for now in fact. I will revert this!

src/lightly_train/_task_models/task_model.py
liopeer: Yes, this is meant for optimizations such as reparameterizations where branches of parameters get merge into single parameters etc. This is already happening in LT-DETR anyway, so I thought it's good to formalize.

PR #889, opened 2026-07-21
Update Docs

mrpositron:
## What has changed and why?

Audits and improves the finetuning task documentation (object detection, instance/semantic/panoptic segmentation, depth estimation, and the object-detection quick start). Two goals: make the docs consistent with the actual public API, and make every argument mentioned in prose clickable so it routes to where it's defined.

**Doc/code fixes**

- Removed the `metric_log_classwise` setting from `train_settings.md`, it doesn't exist in the source. Added a note pointing to the real per-class-metrics toggle, `classwise` under `metric_args`.
- Fixed a broken API cross-reference in `object_detection.md` (`DINOv3LTDETRObjectDetection` → `LTDETRObjectDetection`), which was failing the `--fail-on-warning` build.
- Corrected `semantic_segmentation.md` prose that referred to a `checkpoint` parameter where the code uses `model=`.
- Added the missing `import torch` to the panoptic visualize snippet.
- ...

## How has it been tested?

Links + build: make -C docs docs (sphinx-build -b html --fail-on-warning --keep-going) → build succeeded, confirming every cross-reference resolves and the previously-broken references are gone.

- [ ] Yes

- [x] Yes


docs/source/settings/train_settings.md
```diff
@@ -182,21 +181,9 @@ lightly_train.train_object_detection(
...
-
-lightly_train.train_object_detection(
-    ...,
-    model_args={
-        "metric_log_classwise": True,
-    },
-)
+```{note}
+To log per-class metrics (for example AP or IoU per class), use the
+[`classwise`](#classwise) key of [`metric_args`](#metric_args) rather than `model_args`.
```
liopeer: Should we even mention "rather than `model_args`"? We probably don't want them to use it in that way.

liopeer: Especially now it's part of the `backbone_weights` section, which does not make much sense imo.

docs/source/depth_estimation.md
```diff
@@ -138,9 +138,10 @@ an NVIDIA T4 GPU:
 
 ## Quick Start
 
-Load a model and call `predict` on an image. The image can be a file path, a URL, a PIL
-image, or a `(C, H, W)` tensor. The result is a single `(H, W)` tensor with the same
-height and width as the input image.
+Load a model with {py:func}`load_model <lightly_train.load_model>` and call
```
liopeer, line 141: Amazing! 👏

docs/source/instance_segmentation.md
```diff
@@ -5,7 +5,7 @@
 [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/lightly-ai/lightly-train/blob/main/examples/notebooks/eomt_instance_segmentation.ipynb)
 
 ```{note}
-🔥 LightlyTrain now supports training **DINOv3**-based instance segmentation models
+LightlyTrain now supports training **DINOv3**-based instance segmentation models
```
liopeer, line 8: 😢

liopeer (approved):
LGTM, with a single relevant comment! Feel free to address and resolve yourself.

PR #890, opened 2026-07-21
perf: release train DataLoader workers before validation

yutong-xiang-97:
## What has changed and why?

During task training, two DataLoader worker pools could be alive at the same time throughout every validation phase.

`InfiniteCycleIterator` caches its underlying iterator and keeps it alive across training steps, so the train workers did not shut down before validation. When validation started, `iter(val_dataloader)` spawned a second pool while the train pool was still held. Because the validation loader uses the same `num_workers` as training, the number of live worker processes doubled during validation (for example, 2 GPUs x 8 workers = 16 -> 32).

This PR ensures only one task DataLoader worker pool is alive at a time:

1. **Release train workers before validation**: when `num_workers > 0`, `infinite_train_dataloader.reset()` drops the cached train iterator before validation starts, allowing its workers to shut down before validation workers spawn.
2. **Release validation workers before training resumes**: the validation iterator is scoped with `try/finally` and deleted in the `finally`, so validation workers are released even if validation exits early or raises.
3. **Disallow persistent task workers**: `persistent_workers` passed through `loader_args` is now ignored for both task train and validation dataloaders. Persistent workers are owned by the DataLoader and survive iterator shutdown, so allowing this override would defeat the release-before-validation behavior.

## How has it been tested?

Ran a 2-GPU / `num_workers=4` job under SLURM on a small COCO instance-segmentation subset (`ltdetrv2-seg-s`, 30 steps, validating at steps 10/20/30 plus the final step). A `psutil` monitor sampled the number of live DataLoader worker processes, matched by their `pt_data_worker` process name and isolated to the job by run name, every 0.25 s.

- [x] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/train_task.py
+++ b/src/lightly_train/_commands/train_task.py
@@ -1511,9 +1511,11 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
         )
         train_dataloader.collate_fn = train_collate_fn  # type: ignore[assignment]
         val_dataloader.collate_fn = val_collate_fn  # type: ignore[assignment]
-        helpers.disable_persistent_workers_for_step_aware_transforms(
-            train_dataloader, train_transform, train_collate_fn
-        )
+        # NOTE: persistent_workers is forced to False for both dataloaders at
+        # construction time (see helpers.get_train_dataloader / get_val_dataloader).
+        # Persistent workers are owned by the DataLoader and survive StopIteration, so
+        # they cannot be released between the training and validation phases, which
+        # would keep both worker pools alive at once.
         # Wrap dataloaders with fabric *after* installing the collate function.
         # fabric's _FabricDataLoader snapshots the underlying dataloader; assigning
         # collate_fn on the wrapper would not reach the iterator.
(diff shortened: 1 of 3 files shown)


src/lightly_train/_commands/train_task.py
```diff
@@ -1867,6 +1869,17 @@ def _train_task_from_config(config: TrainTaskConfig) -> None:
...
+                        # Release train DataLoader workers before validation starts so
+                        # the train and validation worker pools never coexist (which
+                        # would multiply the live worker processes per rank).
+                        #
+                        # Training is step-based rather than epoch-based, so resetting
+                        # mid-epoch is acceptable: it re-shuffles from the top and drops
+                        # the current epoch's remaining batches, but we make no
+                        # per-epoch coverage guarantees anyway. The next next() call
+                        # re-creates the iterator and respawns the workers.
+                        infinite_train_dataloader.reset()
```
liopeer, line 1882: Now there is basically no guarantee that if you set `steps=len(dataset) // batch_size` that every sample gets seen once. I am not sure if this is desirable tbh and would prefer a different solution, since most people would expect that to be the case. Also we still log the "epoch".

liopeer (commented):
Tbh, I don't think we can merge it like this. I don't have an idea at the moment either on how to handle this, but if I have a calm minute in the next few days I will try to figure something out. The sampler for the train dataloader needs to somehow be resumable.

yutong-xiang-97:
Note: we would go with improving the batch sampler later on and leave this PR as is atm. (Issue backlogged with a plan on Linear)

PR #893, opened 2026-07-21
Fix ckpt loading for fine-tuning

liopeer:
## What has changed and why?
- our ckpt loading was broken for downloadable models and for LT-DETRv1 models

## How has it been tested?
- unit tests
- tested fine-tuning with `dinov3/vitt16-ltdetr-coco`, `dinov3/vits16-ltdetr-coco`, `dinov2/vits14-noreg-ltdetr-coco` and `ltdetrv2-s-coco` and all gave expected val mAP values after only a single step of fine-tuning

- [x] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -1306,7 +1306,8 @@ def read_model_init_args_from_ckpt(ckpt_path: PathLike) -> dict[str, Any]:
     except (TypeError, RuntimeError, NotImplementedError, AttributeError):
         ckpt = torch.load(p, map_location="cpu", weights_only=False)
 
-    return cast(dict[str, Any], ckpt["model_init_args"])
+    model_init_args: dict[str, Any] = ckpt["model_init_args"]
+    return model_init_args
 
 
 def load_checkpoint(
--- a/src/lightly_train/_commands/train_task_helpers.py
+++ b/src/lightly_train/_commands/train_task_helpers.py
@@ -1317,7 +1317,23 @@ def load_checkpoint(
     checkpoint: PathLike | None,
     task: str,
 ) -> tuple[CheckpointDict | None, Path | None, str, dict[str, Any] | None]:
-    """Build a checkpoint context from the current run configuration.
+    """Resolve model initialization and resume checkpoints.
+
+    Resolution follows these rules, in order:
+
+    1. ``resume_interrupted=True`` restores the full training state from
+       ``out/checkpoints/last.ckpt``. The ``model`` argument is not resolved because
+       the resume checkpoint is authoritative.
+    2. An explicit ``checkpoint`` starts a new fine-tuning run. Its metadata determines
+       the exact architecture, while ``model`` is used to verify the task-model family
+       and as a fallback for legacy checkpoints without model metadata. Only model
+       weights are returned.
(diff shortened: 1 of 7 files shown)


src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -1317,7 +1318,23 @@ def load_checkpoint(
     checkpoint: PathLike | None,
     task: str,
 ) -> tuple[CheckpointDict | None, Path | None, str, dict[str, Any] | None]:
-    """Build a checkpoint context from the current run configuration.
+    """Resolve model initialization and resume checkpoints.
+
+    Resolution follows these rules, in order:
+
+    1. ``resume_interrupted=True`` restores the full training state from
```
gabrielfruet, line 1343: Very clear explanation!

src/lightly_train/_commands/train_task_helpers.py
```diff
@@ -1378,19 +1382,39 @@ def load_checkpoint(
...
         ckpt_path = Path(checkpoint).resolve()
-    elif model_path is not None:
-        ckpt_path = model_path
     else:
-        # No checkpoint to load. Backbone will be initialized from model name.
-        return (None, None, model_name, None)
+        model_path = Path(model).resolve()
+
+        # Case 3: A local path passed as model starts a new fine-tuning run.
+        if model_path.exists():
```
gabrielfruet: I think I also prefer the clarity here instead of collapsing these two into a single if statement

tests/_task_models/test_task_model_helpers.py
```diff
@@ -187,6 +193,39 @@ def test_init_model_from_checkpoint__legacy_dinov2_ltdetr_reroutes_to_generic()
...
     assert isinstance(model, LTDETRObjectDetection)
     assert model.init_args["model_name"] == "dinov3/vitt16-notpretrained-ltdetr"
-    assert model.init_args["decoder_name"] == "rtdetrv2"
+    assert model.init_args["decoder_name"] is None
+    assert isinstance(model.decoder, RTDETRTransformerv2)
+    for name, param in model.state_dict().items():
+        assert torch.equal(param, reference_model.state_dict()[name])
+
+
+def test_init_model_from_checkpoint__missing_decoder_uses_dfine_architecture() -> None:
```
gabrielfruet: The ones that have missing decoder, shouldn't they use the RTDETRv2 decoder? since they probably were old checkpoints

gabrielfruet (approved):
LGTM!

src/lightly_train/_commands/train_task_helpers.py
liopeer: This will be done in the follow up #896

src/lightly_train/_commands/train_task_helpers.py
liopeer: Also took me a while to understand how it worked. Had not worked on this part of the code before.

tests/_task_models/test_task_model_helpers.py
liopeer: Really misleading test, sorry for that. But yes, missing decoder name should indeed resolve to RTDETRv2. The test case is actually already covered by `test_init_model_from_checkpoint__legacy_dinov2_uses_registered_decoder` and the test title is clearer there. The implementation is as follows: If the `decoder_name` is not present, it will just fallback to the architecture that is in the model registry.

PR #895, opened 2026-07-21
Add SAHI inference for LT-DETRv2 instance segmentation

yutong-xiang-97:
## What has changed and why?

Adds Slicing Aided Hyper Inference (SAHI) support to LT-DETRv2 instance segmentation through `predict_sahi`.

The implementation runs a global prediction, processes overlapping tiles in configurable batches, stitches tile masks into the original image coordinates, merges global and local predictions, removes empty masks, and derives bounding boxes from the merged masks. The public Python API documentation now exposes the LT-DETR instance-segmentation prediction and export methods.

This PR is intentionally scoped to adding SAHI inference support. It does not include refactoring of the task model, shared tiling utilities, or related inference paths.

## How has it been tested?

- `make static-checks`
- `pytest tests/_task_models/ltdetr_instance_segmentation/test_task_model.py -q` (40 passed, 1 skipped)

- [ ] Yes
- [x] Not needed (no changelog entry included in this focused PR)

- [x] Yes


liopeer (approved):
Looks correct. In the future we need to make sure that this runs batched (but only after we externalize pre-/postprocessing like we do for OD).

PR #897, opened 2026-07-21
Add benchmark script for inference

liopeer:
## What has changed and why?
- stacks on top of #888 
- adds a new directory for structured benchmarks
- starts the necessary dependency management
- adds basic TensorRT benchmarks – these are very preliminary and for scheduled benchmarks we will have to structure them better, but it's already very usable for debugging

## How has it been tested?
- ran the entire suite on RTX4090

- [ ] Yes

- [ ] Yes


inference_benchmarks/README.md
```diff
@@ -0,0 +1,63 @@
+# Inference Benchmarks
```
gabrielfruet, line 1: I assume this inference_benchmarks is a whole new package right? That is not shipped together with the lightly-train package? 

Because, if not, I would expect it to be under src/, maybe (?)

inference_benchmarks/run_tensorrt_benchmark.py
```diff
@@ -0,0 +1,149 @@
...
+            "images": str(root / "images" / "train2017"),
+        },
+        "val": {
+            "annotations": str(root / "annotations" / "instances_val2017.json"),
+            "images": str(root / "images" / "val2017"),
+        },
+    }
+
+
+def sanitize(model_name: str) -> str:
```
gabrielfruet, line 91: This is meant for some kind of BC?

gabrielfruet (approved):
LGTM! Small questions only.

I assume we should add the ltdetrv2-m,l,x as soon as it arrives. Do you think this is a priority or not? It should be very simple (just adding the new names)

I think we could even create an issue and forward that to a OSS contributor.

inference_benchmarks/README.md
liopeer: I don't want this to be shipped, it is meant to regularly test inference performance in e.g. CI, which is important for refactorings of the export API.

inference_benchmarks/run_tensorrt_benchmark.py
liopeer: The `dinov2/` or `dinov3/` prefixes would create folders if we did not sanitize.

PR #903, opened 2026-07-23
refactor(object-detection)!: rework pre-/postprocessors

liopeer:
## What has changed and why?

This PR refactors the object detection pre- and postprocessors, gives predictions a typed return value, and brings PicoDet up to feature parity with LT-DETR.

### Typed predictions

`LTDETRObjectDetection.predict()`, `predict_batch()`, `predict_sahi()`, `predict_sahi_batch()`, and `PicoDetObjectDetection.predict()` now return an `ObjectDetectionPrediction` instead of a plain `dict`:

```python
results = model.predict("image.jpg")
results.labels  # Class labels, tensor of shape (num_boxes,)
results.bboxes  # Boxes in (xmin, ymin, xmax, ymax), tensor of shape (num_boxes, 4)
results.scores  # Confidence scores, tensor of shape (num_boxes,)
```

Both object detection models now agree on the return type.

### Filtering

Predictions support row filtering, in the style of `supervision.Detections`. A filter always returns a new prediction and does not change the original one:

```python
confident = results[results.scores > 0.8]
people = results[results.labels == 0]
confident.num_detections
```

The filter accepts a boolean mask, an integer tensor, a list of integers, a slice, or a single integer. Row filtering comes from a new `RowIndexableOutput` base class. Other task models can adopt the same API later.

### Backward compatibility

`ObjectDetectionPrediction` is a `Mapping`, so dict-style reads keep working: `results["bboxes"]`, `"bboxes" in results`, `results.keys()`, `dict(results)`, and `**results`.

Two things do change:

- `isinstance(results, dict)` is no longer true. Use `isinstance(results, collections.abc.Mapping)` instead.
- These dict methods raise `AttributeError`: `results["bboxes"] = ...`, `.update()`, `.pop()`, and `.copy()`. Use `results[mask]` to filter, or `dict(results)` for a mutable copy.

`len(results)` returns the number of fields and iteration yields field names, because predictions are also a `Mapping`. Use `results.num_detections` to count detections.

### PicoDet catches up with LT-DETR

`PicoDetObjectDetection` now has `predict_batch()`, `predict_sahi()`, and `predict_sahi_batch()`, matching `LTDETRObjectDetection`. This came with a few behavior changes for PicoDet:

- `predict()` now returns at most `max_detections` results (100 by default) instead of one detection per anchor. For `picodet-s-coco` that was 3598 detections before.
- A detection is kept when `score > threshold` instead of `score >= threshold`.
- The same anchor can now produce detections for more than one class, because top-k selection runs over `(anchor, class)` pairs instead of just anchors.
- `predict()` is now implemented on top of `predict_batch()` and shares the same postprocessor as LT-DETR.
- Training and validation metrics are now computed in original-image pixel coordinates instead of model-input pixels. This matches LT-DETR and what `predict()` returns, but it does change the absolute-area metrics (`map_small`, `map_medium`, `map_large`) for an otherwise identical model.

### SAHI configuration

`ObjectDetectionSahiConfig` moved from the constructors of `ObjectDetectionPreprocessor` and `ObjectDetectionPostprocessor` to the `preprocess_image()` and `postprocess()` calls. Before this change, `predict_sahi_batch()` constructed two new `nn.Module` objects on every call, and it bypassed `self.preprocessor` and `self.postprocessor`. Both modules are now stateless and are reused. This change also removes a duplicate `internal_class_to_class` buffer and a duplicated validation of the metadata.

### SAHI benchmarking

`benchmark_object_detection()` gained a `sahi_args` argument to benchmark Slicing Aided Hyper Inference the same way `predict_sahi()` runs it. It accepts `overlap`, `nms_iou_threshold`, and `global_local_iou_threshold`, defaulting to the same values `predict_sahi()` uses. A non-zero `threshold` is recommended alongside it, since merging tile predictions runs NMS over everything that survives thresholding. The ONNX and TensorRT backends need a dynamic export batch size for this, because tiling makes the number of model input rows differ from batch to batch.

### Shared decoding

The decode math (sigmoid, top-k, box conversion, rescale, class remap) existed in the postprocessor and again in `train_model._decode_predictions_for_metrics`. Both now call one `decode_object_detection_output()` function. The postprocessor gets no new public method.

### Class-aware NMS in SAHI

`combine_object_detection_tiles` now uses `batched_nms` instead of `nms`. Before this change, a high-confidence detection suppressed an overlapping detection of a different class. SAHI can now return more boxes for overlapping objects of different classes. The changelog records this change under `Fixed`.

### Breaking: PicoDet export changes

`PicoDetObjectDetection.export_onnx()` and `export_tensorrt()` now produce the same graph as LT-DETR. The outputs change from `["labels", "boxes", "scores"]` to `["logits", "boxes"]`: `logits` are raw pre-sigmoid class scores, and `boxes` are normalized `cxcywh` boxes relative to the model input (previously boxes were `xyxy` in model-input pixels). The batch dimension is renamed from `"N"` to `"batch_size"`. Class-id remapping, top-k selection, score thresholding, and rescaling to the original image are no longer part of the exported graph; call `model.postprocess(...)` after running the exported model, or replicate `decode_object_detection_output` in your own deployment code. The export now goes through the same TorchDynamo pipeline as LT-DETR, which requires `torch>=2.6`. Existing exported PicoDet ONNX/TensorRT models need to be re-exported, and any custom postprocessing code needs to be updated to match. Separately, `precision="auto"` is no longer accepted; use `"fp32"` (the new default) or `"fp16"`. Exported ONNX models also now carry the same metadata as other task models (`lightly_train_version`, `license_info`, `image_normalize`, `classes`, `model_name`) instead of only `classes`.

### Bug fixes

- `export_tensorrt(precision="fp16")` built its FP16 engine from an FP32 intermediate ONNX model. It now exports that intermediate model in FP16 too.
- The Torch backend of `benchmark_object_detection()` never switched PicoDet models to eval mode, so batch norm stayed in training mode and the reported PicoDet numbers were wrong.
- The Torch backend of `benchmark_object_detection()` stopped its timer before the GPU had actually finished. CUDA kernels are queued asynchronously, so latency was reported far too low and throughput far too high on CUDA devices. The ONNX and TensorRT backends were not affected. CUDA timings from earlier releases are not comparable to what's reported now.
- The Torch backend now decodes raw model outputs in FP32, like the ONNX and TensorRT backends already did. This can slightly change the numbers reported for `precision="fp16-mixed"` and `"bf16-mixed"` runs, which are now comparable to the other backends.

### Shared tiling code

The tiling logic used to live in `object_detection_components/tiling_utils.py`. It moved to a new shared module, `_pre_post_processing/tiling.py`, so it isn't tied to one task model. Alongside it there's a new `_pre_post_processing/instance_segmentation.py` with a class-aware mask NMS helper. It isn't used by any task model yet — it's unit-tested groundwork for adding SAHI-style tiling to instance segmentation later.

## How has it been tested?

- New test: `test_predict_sahi_batch__splits_raw_outputs_per_image` uses two images with different tile counts (2 and 3 rows in one forward pass). Every row carries one detection with its own class, score, and location. The test asserts that each image receives its own rows and that the boxes land in the correct image. The earlier version of this test only produced empty predictions, so it did not detect an off-by-one error in the slicing.
- New tests for `RowIndexableOutput`: boolean mask, index tensor, list, slice, integer index, empty selection, invalid indices, the preserved `Mapping` protocol, and the pytree round trip. One test asserts that a filter never returns a view: the data pointer differs, and a write to the result does not change the original.
- New tests for `ObjectDetectionPrediction`: filter by score, filter by label, new object semantics, `num_detections`, and `to_torchmetrics()`.
- New tests for the error paths: SAHI metadata for more than one image, and empty input to `predict_batch()`.
- New test: `test_predict__matches_predict_batch`, because `predict()` now delegates to `predict_batch()`.
- New tests for the shared `_pre_post_processing/tiling.py` module (`test_tiling.py`) and the new `_pre_post_processing/instance_segmentation.py` mask NMS helper (`test_instance_segmentation.py`).
- Expanded `test_benchmark_task.py` to cover the new `sahi_args` benchmarking path, plus the PicoDet eval-mode and CUDA-timer fixes.
- New/expanded train_model tests for LT-DETR and PicoDet (`test_train_model.py` in both `ltdetr_object_detection` and `picodet_object_detection`) covering the shared decode path and PicoDet's original-image-coordinate metrics.

Commands:

```
pytest tests/_pre_post_processing/ tests/_task_models/test_task_model_io.py \
  tests/_task_models/ltdetr_object_detection/ \
  tests/_task_models/object_detection_components/ \
  tests/_task_models/picodet_object_detection/ \
  tests/_commands/test_benchmark_task.py
# 287 passed, 11 skipped

pytest tests/_commands/test_train_task.py::test_train_object_detection_yolo
# 1 passed

make type-check     # Success: no issues found in 640 source files
make format-check   # passed
```

- [x] Yes

- [x] Yes


src/lightly_train/_pre_post_processing/object_detection.py
```diff
@@ -27,20 +30,56 @@
...
+class ObjectDetectionDecodedBatch(BaseModelOutput):
+    """Dense top-k detections before score filtering or SAHI merging."""
+
+    labels: Tensor
+    bboxes: Tensor
+    scores: Tensor
+
+
+@dataclass
+class ObjectDetectionPrediction(BaseModelOutput):
```
liopeer: Looks like a repetition of `ObjectDetectionDecodedBatch`.

src/lightly_train/_pre_post_processing/object_detection.py
```diff
@@ -27,20 +30,56 @@
...
+    target_sizes: Tensor
+
+
+@dataclass
+class ObjectDetectionSahiPreprocessedBatch(BaseModelOutput):
+    """Flattened global/tile inputs and tensor-only SAHI reconstruction data."""
+
+    images: Tensor
+    target_sizes: Tensor
+    tile_offsets: Tensor
```
liopeer: Would be great to have tensor shapes in the docstring.

src/lightly_train/_commands/benchmark_task.py
```diff
@@ -444,10 +446,28 @@ def _create_val_dataloader(
...
+                )
+            )
+        else:
+            result.append(
+                {key: value.detach().cpu() for key, value in prediction.items()}
+            )
+    return result
+
+
+def _prediction_value(prediction: Any, name: str) -> Tensor:
```
liopeer: I don't think this is needed.

src/lightly_train/_commands/benchmark_task.py
```diff
@@ -444,10 +446,28 @@ def _create_val_dataloader(
     )
 
 
-def _to_cpu(
-    predictions: list[ObjectDetectionPrediction],
-) -> list[ObjectDetectionPrediction]:
-    return [{k: v.detach().cpu() for k, v in p.items()} for p in predictions]  # type: ignore[attr-defined,misc]
+def _to_cpu(predictions: list[Any]) -> list[Any]:
```
liopeer: We should be able to leave this as is.

PR #909, opened 2026-07-24
Add C++ inference recipes for LT-DETR object detection

gabrielfruet:
## What has changed and why?

C++ inference recipes for LT-DETR object detection under `examples/cpp/`, for ONNX Runtime (CUDA EP) and TensorRT (EP), both with zero-copy GPU input allocation. Postprocessing (top-k over query×class, cxcywh→xyxy, class remap) mirrors `ObjectDetectionPostprocessor.decode()` exactly. These are copy-paste-and-edit recipes (hardcoded constants, no CLI/JSON parsing), matching the style of the existing export notebooks. Linked from `object_detection_export.ipynb`.

Known limitation, not fixed here: `model.export_tensorrt(precision="fp32")` on `ltdetrv2-s-coco` currently produces a numerically degraded engine (max score drops from ~0.97 to ~0.32) — TensorRT's weakly-typed builder appears to pick FP16 tactics in the transformer decoder despite `fp32` being requested. `strongly_typed=True` fixes it. Tracking that separately; the TensorRT recipe here is verified correct against a `strongly_typed` engine.

## How has it been tested?

Built and ran both recipes on a GPU node (Docker, ONNX Runtime 1.20.1 + TensorRT 10.13.3.9) against `ltdetrv2-s-coco` on the COCO `000000039769.jpg` sample. ONNX Runtime recipe matches `model.predict(threshold=0.6)` almost exactly (scores within ~0.01, boxes within ~0.1px). TensorRT recipe matches Python's TensorRT backend byte-for-byte given the same engine.

- [x] Yes

- [x] Yes


examples/cpp/cmake/FindTensorRT.cmake
```diff
@@ -0,0 +1,45 @@
...
+find_path(TensorRT_INCLUDE_DIR
+  NAMES NvInfer.h
+  HINTS ${TensorRT_ROOT} ENV TensorRT_ROOT
+  PATH_SUFFIXES include
+)
+
+find_library(TensorRT_LIBRARY
+  NAMES nvinfer
+  HINTS ${TensorRT_ROOT} ENV TensorRT_ROOT
+  PATH_SUFFIXES lib lib64
```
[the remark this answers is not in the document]
gabrielfruet: Fixed — pip tensorrt-cu12 has no C++ headers at all, so docs now require a native/apt install instead.

gabrielfruet:
Verification: compute-02-ubuntu-4x4090 (1 GPU, Docker — CUDA 12.4.1 devel, ONNX Runtime GPU 1.20.1 tarball, TensorRT 10.13.3.9 via apt), `ltdetrv2-s-coco` on COCO `000000039769.jpg`.

**ONNX Runtime recipe** vs `model.predict(threshold=0.6)`: cat 0.967/0.963, couch 0.962, remote 0.915/0.855 — matches within ~0.01 score, boxes within ~0.1px.

**TensorRT recipe**: default `precision="fp32"` export is weakly-typed and picks FP16 tactics in the decoder, so max score collapses to ~0.32 (should be ~0.97). Rebuilt with `strongly_typed=True`: couch score=0.962, box=(0.2, 1.1, 641.9, 474.3) — matches Python's 0.9618 / [1.24, 1.30, 639.52, 474.11]. Repro'd the precision bug independently with a pure-Python TensorRT harness mirroring `TensorRTBackend`, so it's an export-side issue, not the C++ code — this recipe is verified correct against the strongly_typed engine. Not fixing the export heuristic here; will open a separate Linear ticket for it.

No screenshots — verification ran headless on the cluster and the image was dropped afterward.

gabrielfruet:
Correction to my earlier comment — the TensorRT validation claim was overstated.

- `strongly_typed=True` doesn't fix the precision bug, just masks it partially: top scores land at 0.91/0.72 instead of 0.97/0.96, and 3 of 5 ONNX Runtime detections are still missing. Couch happened to land close, which is all the earlier comment checked.
- `fp32_attention_scores=True` (forcing decoder attention to FP32) has no effect. Ruled out.
- Root cause, found by bisecting: marking subsets of the ~1600 intermediate ONNX tensors as outputs and rebuilding. It's RT-DETR's encoder-to-decoder anchor selection block (class-score head -> ReduceMax -> TopK). TensorRT fuses it into one kernel and computes the anchor-ranking scores at reduced precision, so different anchors get picked than in the reference model, corrupting every downstream detection.

Only the ONNX Runtime recipe is verified against `model.predict()`. TensorRT fix tracked in [TRN-2341](https://linear.app/lightly/issue/TRN-2341/tensorrt-export-precision-bug-in-ltdetrv2-s-coco-traced-to-encoder) — not blocking this PR, the recipe code faithfully reproduces whatever engine it's given.

examples/cpp/onnxruntime/main.cpp
```diff
@@ -0,0 +1,186 @@
...
+  Ort::Env env(ORT_LOGGING_LEVEL_WARNING, "od_infer_onnxruntime");
+  Ort::SessionOptions session_options;
+
+  OrtCUDAProviderOptions cuda_provider_options{};
+  cuda_provider_options.device_id = kDeviceId;
+  session_options.AppendExecutionProvider_CUDA(cuda_provider_options);
+
+  Ort::Session session(env, kModelPath, session_options);
+
+  const auto pre = od_common::preprocess_image(kImagePath, kModelWidth, kModelHeight,
```
liopeer: Nit: Would use a different naming here, not just `pre`.

examples/cpp/onnxruntime/main.cpp
```diff
@@ -0,0 +1,186 @@
...
+                                                kNormalize);
+
+  Ort::MemoryInfo cuda_memory_info("Cuda", OrtArenaAllocator, kDeviceId,
```
liopeer: This is not super important, but I would also manually allocate the outputs with `cudaMalloc`. The `OrtArenaAllocator` obscures memory management a bit away, which I think we want to avoid in this example.

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- Add [C++ inference recipes](examples/cpp/README.md) for LT-DETR object detection,
```
liopeer, line 12: You're not using ONNXRuntime with the TensorRT EP, you're directly using TensorRT.

examples/cpp/README.md
```diff
@@ -0,0 +1,91 @@
...
+[`examples/notebooks/object_detection_export.ipynb`](../notebooks/object_detection_export.ipynb),
+just in C++.
+
+Two recipes are provided:
+
+- `onnxruntime/` -- runs an exported ONNX model with ONNX Runtime's CUDA
+  execution provider.
+- `tensorrt/` -- runs a serialized TensorRT engine.
+
+Both allocate their model input directly on the GPU (zero-copy) instead of
```
liopeer: Should be input + output that gets directly allocated.

examples/cpp/README.md
```diff
@@ -0,0 +1,91 @@
...
+On a machine with `lightly-train` installed:
+
+```python
+import lightly_train
+
+model = lightly_train.load_model("ltdetrv2-s-coco")
+print(model.image_size, model.image_normalize, model.classes)
+
+model.export_onnx("model.onnx")
+model.export_tensorrt("model.trt")  # requires the `tensorrt` package, see below
```
liopeer: Move the tensorrt requirement to the top of section, i.e. give some instructions, e.g. install LightlyTrain with TensorRT support (tested with TensorRT 10.x): `pip install lightly-train tensorrt-cu12 # or tensorrt-cu13`

examples/cpp/README.md
```diff
@@ -0,0 +1,91 @@
...
+The printed `image_size` / `image_normalize` / `classes` must match the
+`kModelHeight`/`kModelWidth`/`kNormalize`/`kClassNames` constants at the top
+of `onnxruntime/main.cpp` and `tensorrt/main.cpp` -- update them if you
+export a different checkpoint. TensorRT engines carry no metadata (unlike
+ONNX's embedded `image_normalize`/`classes`), so this hardcoding is required
+for the TensorRT recipe in particular.
+
+## 2. Prerequisites
+
+- CUDA toolkit (matching the driver on your machine).
```
liopeer, line 59: For our own sake we should include a (or two) development container(s) with this feature. If you have never worked with dev containers, you can check my MIGraphX PR: https://github.com/lightly-ai/lightly-train/pull/912/changes (includes a hint in the CONTRIBUTING.md on how to use it).

examples/cpp/README.md
```diff
@@ -0,0 +1,91 @@
...
+- For the ONNX Runtime recipe: the **prebuilt GPU tarball** from the
+  [ONNX Runtime releases page](https://github.com/microsoft/onnxruntime/releases)
+  (e.g. `onnxruntime-linux-x64-gpu-<version>.tgz`) -- the pip `onnxruntime-gpu`
+  wheel does not reliably ship the C++ headers/import library needed here.
+- For the TensorRT recipe: a native TensorRT install (e.g. `libnvinfer-dev`
+  and matching version pins via
+  [NVIDIA's CUDA apt repo](https://developer.nvidia.com/cuda-downloads), or an
+  extracted TensorRT tarball) matching the version pinned in the notebook
+  (`tensorrt-cu12==10.13.3.9`) -- the pip `tensorrt-cu12` wheel does **not**
+  ship the `NvInfer.h` C++ headers needed here, only Python bindings.
```
liopeer, line 70: Would clearly separate the environment setup of export vs. inference.

examples/cpp/README.md
```diff
@@ -0,0 +1,91 @@
...
+`libnvinfer-dev`) already puts headers/libs on the default search path.
+
+## 4. Run
+
+```bash
+./examples/cpp/build/onnxruntime/od_infer_onnxruntime
+./examples/cpp/build/tensorrt/od_infer_tensorrt
+```
+
+Each program reads `image.jpg` (edit `kImagePath` for a different file) and
```
liopeer, line 99: Best to give a download link for an example image, e.g. the one from the quickstart in the docs.

examples/cpp/onnxruntime/main.cpp
```diff
@@ -0,0 +1,186 @@
...
+constexpr const char* kOutputPath = "output.jpg";
+constexpr float kThreshold = 0.6f;
+constexpr int kDeviceId = 0;
+constexpr int kModelHeight = 640;  // must match model.image_size[0] at export time
+constexpr int kModelWidth = 640;   // must match model.image_size[1] at export time
+// Must match the exported model's postprocessor config (LT-DETR "Generic"
+// default is 300, see src/lightly_train/_task_models/ltdetr_object_detection/config.py).
+constexpr int kNumTopQueries = 300;
+
+// Must match model.image_normalize for the exported checkpoint. "ltdetrv2-s-coco"
```
liopeer: You could theoretically read this from the model. It's not important that you do (might actually make it more readable like this), but you could. Same for the image size.

examples/cpp/tensorrt/main.cpp
```diff
@@ -0,0 +1,277 @@
...
+// (src/lightly_train/_commands/benchmark_backends.py::TensorRTBackend),
+// which is the reference this file ports to C++.
+//
+// Export the engine first, on a machine with lightly-train + tensorrt-cu12
+// installed:
+//
+//   import lightly_train
+//   model = lightly_train.load_model("ltdetrv2-s-coco")
+//   print(model.image_size, model.image_normalize, model.classes)
+//   model.export_tensorrt("model.trt")
```
liopeer, line 26: It's fine for now, but we should later use the "proper" fp16 model.

liopeer (changes requested):
Nice work! I would change a few things and I would add a dev container setup, otherwise we will never touch these scripts again, because the setup is so annoying. I've linked a draft in #915, which just needs a bit of cleanup, but is properly working.

liopeer (approved):
LGTM!

PR #912, opened 2026-07-25
MIGraphX Support

liopeer:
## What has changed and why?
- adds MIGraphX export support including
  - example on how to export / run inference
  - dev container setup, since no pip install available

## How has it been tested?
- unit tests inside dev container

TBD

- [x] Yes

Will do later, with the entire *Deploy* section, see #921 
- [ ] Yes


.devcontainer/migraphx/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM rocm/dev-ubuntu-24.04:7.2.4
+
+RUN apt-get update \
+    && DEBIAN_FRONTEND=noninteractive apt-get install --yes --no-install-recommends \
+        migraphx \
+        python3-pip \
+        git \
+    && rm -rf /var/lib/apt/lists/* \
+    && python3 -m pip install --break-system-packages --no-cache-dir \
+        "uv==0.11.32"
```
liopeer, line 10: Don't install uv via pip. Check the dockerfile for TensorRT/ONNXRT on how to install it.

CONTRIBUTING.md
```diff
@@ -9,6 +9,41 @@ git clone https://github.com/lightly-ai/lightly-train.git
 make install-dev
 ```
 
+#### Running MIGraphX Tests
```
liopeer, line 12: Move to the bottom of the file. Only reference it here where the "main" setup happens.

src/lightly_train/_export/export_migraphx.py
```diff
@@ -0,0 +1,252 @@
...
+                "For official installation instructions, see "
+                "https://rocm.docs.amd.com/projects/AMDMIGraphX/en/latest/install.html"
+            )
+
+        onnx_out = Path(out).with_suffix(".onnx")
+        self.export_onnx(
+            out=onnx_out,
+            batch_size=batch_size,
+            dynamic_batch_size=False,
+            opset_version=19,
```
liopeer, line 246: Add TODO to make configurable (opset version) and check whether migraphx supports it. Also other todos to expose simplify and dynamic batching to the user.

src/lightly_train/_export/export_migraphx.py
```diff
@@ -0,0 +1,252 @@
...
+    replaced_resize_outputs: set[str] = set()
+    stale_value_info_names: set[str] = set()
+    for resize in model.graph.node:
+        if resize.op_type != "Resize" or len(resize.input) < 4:
+            continue
+        resize_attributes = {
+            attribute.name: helper.get_attribute_value(attribute)
+            for attribute in resize.attribute
+        }
+        if resize_attributes.get("mode", b"nearest") != b"linear":
```
liopeer, line 76: Check upstream. I think this is an issue with ONNX renaming some parameters.

src/lightly_train/_export/_compile_migraphx.py
```diff
@@ -0,0 +1,42 @@
...
+    args = parser.parse_args()
+    compile_migraphx(
+        onnx_path=args.onnx,
+        out=args.out,
+        precision=args.precision,
+    )
+
+
+if __name__ == "__main__":
+    main()
```
gabrielfruet, line 42: Is this supposed to be a script under the main library code? Shouldn't this be moved somewhere else?

src/lightly_train/_export/export_migraphx.py
```diff
@@ -0,0 +1,252 @@
...
+
+        onnx_out = Path(out).with_suffix(".onnx")
+        self.export_onnx(
+            out=onnx_out,
+            batch_size=batch_size,
+            dynamic_batch_size=False,
+            opset_version=19,
+            simplify=False,
+        )
+        _prepare_onnx_for_migraphx(onnx_out)
```
gabrielfruet, line 249: I really like code like this, where we have very simple and descriptive functions in the method and then the implementation is under the functinos that we call in the method. Reads like english.

src/lightly_train/_export/export_migraphx.py
```diff
@@ -0,0 +1,252 @@
...
+        "lightly_train._export._compile_migraphx",
+        "--onnx",
+        str(onnx_out),
+        "--out",
+        str(out),
+        "--precision",
+        precision,
+    ]
+    try:
+        subprocess.run(command, check=True, env=_migraphx_compile_environment())
```
gabrielfruet, line 199: Any reason to use a subprocess here?

gabrielfruet (commented):
LGTM! Minor questions that may need to be addressed before approval

src/lightly_train/_export/export_migraphx.py
liopeer: The reason is the environment variable that we set above (i.e. `"MIGRAPHX_DISABLE_PASSES"`). It does not properly propagate to the compilation if you just set it in Python – I did not spend too much time investigating why and whether there is really no other workaround without a subprocess though.

This also answers why we need a script to run this and not a standard module.

src/lightly_train/_export/_compile_migraphx.py
liopeer: Answered below.

src/lightly_train/_export/export_migraphx.py
liopeer: I will spend a bit of time again looking into it and maybe find something more elegant (or at least figuring out and documenting why it's implemented this way and at which point we could change it; maybe also raising an issue on MIGraphX's GitHub).

gabrielfruet: I would intentionally not spend to much time on this. 

I personally don't like to put scripts under the library code, let me know your opnion.

Maybe we should have a scripts folder for this.

gabrielfruet (approved):

PR #915, opened 2026-07-27
C++ Dev Container

liopeer:
## What has changed and why?
- stacks on top of #909 
- adds dev container setup

## How has it been tested?
- works on local node

- [ ] Yes

- [ ] Yes


examples/cpp/onnxruntime/main.cpp
```diff
@@ -103,13 +111,18 @@ int main() {
   const auto pre = od_common::preprocess_image(kImagePath, kModelWidth, kModelHeight,
                                                 kNormalize);
 
-  Ort::MemoryInfo cuda_memory_info("Cuda", OrtArenaAllocator, kDeviceId,
+  // OrtDeviceAllocator (not OrtArenaAllocator): this MemoryInfo only
```
gabrielfruet, line 114: I think we don't need to specify that is not OrtArenaAllocator

gabrielfruet (approved):

examples/cpp/onnxruntime/main.cpp
liopeer: Agree, that's AI slop :)

PR #917, opened 2026-07-28
Fix TensorRT GridSample inference for LT-DETR

yutong-xiang-97:
## What has changed and why?

This is PR 2 of a three-PR stack and is based on
`yutong-trn-2329-fix-onnx-verification`.

LT-DETR TensorRT engines could build and run successfully while silently producing
incorrect detections or segmentations because TensorRT compiled the deformable-attention
`GridSample` path incorrectly.

For ONNX export, the affected DFINE and RT-DETRv2 attention modules now replace native
`grid_sample` with equivalent gather-based bilinear interpolation built from simpler
operations such as `Pad`, `Gather`, multiplication, and addition. The resulting ONNX
graphs contain no `GridSample` nodes, so no opset-specific TensorRT parser workaround is
needed. Training and ordinary deployment retain the faster native `grid_sample` path.

The ONNX-only conversion is applied through an explicit runtime-checkable export
protocol. DINOv3 ViT-S/16 FP16 TensorRT export remains strongly typed independently of
the `GridSample` fix, preserving the graph's FP32 attention operations and preventing
FP16 overflow. Other model and precision combinations retain the existing weakly typed
path.

## How has it been tested?

Raw ONNX Runtime versus TensorRT results on the T4 with TensorRT 10.13.3.9:

- Object detection:
  - logits max/mean difference: `7.97e-4` / `1.46e-5`
  - boxes max/mean difference: `5.21e-5` / `3.58e-7`
- Instance segmentation:
  - labels: exact match
  - boxes max/mean difference: `1.05e-2` / `1.48e-4`
  - masks max/mean difference: `1.16e-3` / `2.95e-5`
  - scores max/mean difference: `6.87e-6` / `4.29e-7`

- [x] Yes

- [ ] Yes
- [x] Not needed (the exported model interface is unchanged)

### DIFF
--- a/src/lightly_train/_export/tensorrt_helpers.py
+++ b/src/lightly_train/_export/tensorrt_helpers.py
@@ -154,7 +154,8 @@ def export_tensorrt(
         _debug_mark_all_layers_as_outputs(onnx_out)
 
     logger.info(f"Loading ONNX file from {onnx_out}")
-    onnx_bytes = _fix_grid_sample_modes_for_tensorrt(onnx_out=onnx_out)
+    with open(onnx_out, "rb") as f:
+        onnx_bytes = f.read()
     if not parser.parse(onnx_bytes):
         for error in range(parser.num_errors):
             logger.error(parser.get_error(error))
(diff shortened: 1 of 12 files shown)


src/lightly_train/_task_models/object_detection_components/dfine_decoder.py
```diff
@@ -158,6 +158,18 @@ def _reset_parameters(self):
         init.constant_(self.attention_weights.weight, 0)
         init.constant_(self.attention_weights.bias, 0)
 
+    def convert_to_onnx_export(self) -> None:
```
gabrielfruet, line 161: Nit: I would prefer a protocol for that instead of relyinh on `hasattr`.

src/lightly_train/_export/tensorrt_helpers.py
```diff
@@ -263,6 +263,37 @@ def export_tensorrt(
     logger.info("Engine export complete.")
 
 
+def _fix_grid_sample_modes_for_tensorrt(*, onnx_out: PathLike) -> bytes:
```
gabrielfruet: Since we don't have GridSample anymore anyway at the ONNX, do we need this?

gabrielfruet (approved):
LGTM

PR #923, opened 2026-07-30
Update object detection benchmark plot

gabrielfruet:
## What has changed and why?

Refreshed the object detection benchmark chart (`map_vs_params.png`) with an updated
plot covering the same LT-DETR, LT-DETRv2, RF-DETR, YOLOv11, and YOLOv8 comparison.

## How has it been tested?

Visual diff of the new PNG against the current one; values match, only styling changed.

- [ ] Yes


mrpositron (approved):
LGTM 🚀

Issue #925, opened 2026-08-03
Add multi-scale feature support to all backbone packages


liopeer:
## Description

Dense prediction tasks such as semantic segmentation and object detection need normalized intermediate feature maps from multiple backbone layers or stages. Multi-scale features can also be useful during pretraining and distillation, for example to define objectives or transfer representations at more than one level of the network, e.g. wanted in #628 .

LightlyTrain already defines a public interface for this:

- [`ForwardMultiScaleFeatures`, `MultiScaleFeatureDims`, and `MultiScaleFeatureStrides`](https://github.com/lightly-ai/lightly-train/blob/main/src/lightly_train/_models/model_wrapper.py#L146-L210)
- [`MultiScaleFeatureViT` and `MultiScaleFeatureCNN`](https://github.com/lightly-ai/lightly-train/blob/main/src/lightly_train/_models/model_wrapper.py#L213-L225)
- [`MultiScaleFeaturePackage`](https://github.com/lightly-ai/lightly-train/blob/main/src/lightly_train/_models/package.py#L73-L80)

DINOv2 and DINOv3 provide reference implementations:

- [DINOv2 ViT wrapper](https://github.com/lightly-ai/lightly-train/blob/main/src/lightly_train/_models/dinov2_vit/dinov2_vit.py#L30-L115)
- [DINOv3 ViT wrapper](https://github.com/lightly-ai/lightly-train/blob/main/src/lightly_train/_models/dinov3/dinov3_vit.py#L31-L99)
- [DINOv3 ConvNeXt wrapper](https://github.com/lightly-ai/lightly-train/blob/main/src/lightly_train/_models/dinov3/dinov3_convnext.py#L25-L105)

We would like to extend this interface to the remaining model packages, starting with TIMM. The current [`TIMMModelWrapper`](https://github.com/lightly-ai/lightly-train/blob/main/src/lightly_train/_models/timm/timm.py#L203) already uses TIMM's `forward_intermediates` where available, which should provide a useful starting point.

EdgeCrafter is partially supported: [`ECViTModelWrapper.forward()`](https://github.com/lightly-ai/lightly-train/blob/main/src/lightly_train/_models/ecvit/ecvit.py#L535) already returns a three-level feature pyramid, but its wrapper and package do not yet implement the public multi-scale protocols.

Contributions can address one package at a time; there is no need to implement the whole checklist in a single PR. If you would like to work on one, please leave a comment so work is not duplicated.

## Expected behavior

For each package:

[...]

PR #939, opened 2026-08-11
feat(timm): add multi-scale feature support

### LINKED ISSUE #926: Add multi-scale feature support to TIMM
- [ ] implement multi-scale feature support for the TIMM package, following #925

saud5150:
Closes #926. Part of #925.

## What has changed and why?

`TIMMModelWrapper` implemented `ModelWrapper` only, so timm backbones could not expose
the intermediate feature maps that dense prediction heads need. This adds the
multi-scale protocols to the TIMM package, following the DINOv2/DINOv3 wrappers.

Two wrapper subclasses are added: `TIMMMultiScaleViTModelWrapper` implements
`MultiScaleFeatureViT` and `TIMMMultiScaleCNNModelWrapper` implements
`MultiScaleFeatureCNN`. They are separate classes because the protocols are structural —
a single wrapper carrying both `patch_size()` and `multiscale_feature_strides()` would
satisfy both, and a wrapped ResNet would pass `isinstance(w, MultiScaleFeatureViT)`.
Both delegate to `forward_intermediates(output_fmt="NCHW", intermediates_only=True)`,
with `norm=True` on the ViT path and stage outputs returned as produced on the CNN path,
as agreed in #925.

`get_multiscale_model_wrapper()` picks the wrapper from the feature geometry that timm
reports in `model.feature_info`, rather than from `_TIMM_ARCH_NAME_PREFIXES` as I
proposed in the issue thread. That table classifies the norm type, not the spatial
geometry: `swin_*`, `swinv2_*` and `pvt_v2_*` are `"transformer"` but are hierarchical
and need the CNN protocol, `maxvit_*` and `coat*` are `"hybrid"` and map to neither, and
unknown names fall back to `"transformer"` with a warning. Constant dims and strides
select the ViT path, strictly increasing strides select the CNN path, and anything else
raises. `_TIMM_ARCH_NAME_PREFIXES` is unchanged and still only used by
`architecture_info()`.

Per your fail-fast preference, `ValueError` is raised for models without
`forward_intermediates` (older timm model definitions such as `vgg*`, `densenet*`,
`convit_*`), without `feature_info`, with mixed geometry (`volo_*`, strides
`[8, 16, 16, 16]`), and with a non-square `patch_embed.patch_size` — the last one
otherwise reports a single `patch_size()` for a model whose two axes are strided
differently. Each wrapper re-checks the geometry in its own `__init__` rather than
trusting the factory, so constructing one directly with a mismatched model raises
instead of silently reporting a stem stride as the patch size.

One correctness detail worth flagging: timm returns intermediates in the order the model
produces them and silently drops duplicate indices, so requesting `[3, 0, 2]` yields
stages 0, 2, 3. The spec requires the requested order, so indices are deduplicated and
sorted before the call and the requested order is restored afterwards.

**Open question:** `TIMMPackage` is deliberately *not* yet a `MultiScaleFeaturePackage`.
That protocol types `get_model_wrapper` as
`MultiScaleFeatureViT | MultiScaleFeatureCNN`, but `get_model_wrapper` is also what
`package_helpers.get_wrapped_model` calls for all pretraining, so honoring that
signature means raising for every timm model without `forward_intermediates` and
breaking pretraining for them. It would also make
`LinearSemanticSegmentation.list_model_names()` advertise all 1293 timm models as
`timm/<name>-linear` backbones. Happy to do it in a follow-up (or here) once you have a
preference.

## How has it been tested?

Verified against timm 1.0.28 and torch 2.8.0.

```
uv run --frozen pytest tests/_models/timm -v
59 passed, 2 skipped

uv run --frozen mypy src tests docs/format_code.py inference_benchmarks
Success: no issues found in 638 source files

uv run --frozen ruff format --check . && uv run --frozen ruff check .
662 files already formatted
All checks passed!

uv run --frozen pytest tests
5617 passed, 67 skipped, 1 xfailed
```

35 new tests cover metadata (`multiscale_feature_dims`, `multiscale_feature_strides`,
`patch_size`), output shapes, ordering with a shuffled index list, duplicate, empty,
out-of-range and negative indices, equality with `forward_features()` for the final
layer/stage, and each `ValueError` path. The model matrix is `vit_tiny_patch16_224`,
`resnet18` (extra stem entry in `feature_info`), `convnext_tiny`,
`swin_tiny_patch4_window7_224` (hierarchical transformer), `mobilevit_xxs` (hybrid),
plus `convit_tiny` and `volo_d1_224` for the unsupported paths. ViT coverage includes a
non-224 input, since `TIMMPackage.get_model` forces `dynamic_img_size=True` for
`vit*`/`eva*`/`deit*`.

Before this change the wrappers had no multi-scale methods at all. After,
`get_multiscale_model_wrapper()` resolves:

| Model                          | Wrapper      | `multiscale_feature_dims()` | Patch size / strides |
| ------------------------------ | ------------ | --------------------------- | -------------------- |
| `vit_tiny_patch16_224`         | ViT          | `[192]` × 12                | `patch_size() == 16` |
| `convnext_tiny`                | CNN          | `[96, 192, 384, 768]`       | `[4, 8, 16, 32]`     |
| `swin_tiny_patch4_window7_224` | CNN          | `[96, 192, 384, 768]`       | `[4, 8, 16, 32]`     |
| `convit_tiny`                  | `ValueError` | —                           | —                    |

```
ValueError: Model of type <class 'timm.models.convit.ConVit'> has no
'forward_intermediates' method and therefore does not support multi-scale features.
```

The full suite (`make test`) passes with no regressions.

- [x] Yes

- [ ] Yes
- [x] Not needed (the protocols are internal to `_models` and the public API is
  unchanged)


liopeer (commented):
Looking really solid overall, however this PR surfaced that there is a mismatch with how TIMM handles norms when requesting intermediate features (against how DINOv2/DINOv3/DINOv3ConvNeXt do it – also there it's not 100% coherent). I will quickly have to check what the best way moving forward would be. Maybe an additional hint in the Protocols explaining the "operating principles" or so.

saud5150:
Thanks!

On the norm mismatch — one thing that might save you some digging: passing `norm=True` on the TIMM CNN path wouldn't reproduce the DINOv3 ConvNeXt behaviour. timm only applies it to the last stage (`if norm and feat_idx == last_idx` in ConvNeXt, `i == num_stages - 1` in Swin) and ignores it completely in ResNet, so you'd get a pyramid that's normalized at one level and raw at the others. There's no per-stage hook in `forward_intermediates` equivalent to `DINOv3ConvNeXt.norms[i]`.

The ViT side is already consistent, for what it's worth — DINOv2, DINOv3 and TIMM all apply the final norm to every requested block. It's only the CNN row that differs: DINOv3 ConvNeXt normalizes per stage, TIMM doesn't normalize at all.

If it helps, the principle that describes all of them as they stand today is roughly: *a feature map carries whatever normalization the architecture applies before features leave the backbone* — final norm on every block for ViTs, per-stage norm for CNNs that have one, as produced where there is none. That would just need a sentence in `ForwardMultiScaleFeatures`. The stricter version (everything normalized) would mean wrapper-owned norm layers for TIMM CNNs. Happy to implement either.

On the two Codex comments, so they don't block you:

- The `stem` one is mine — Mixer/ResMLP/gMLP keep the patch embed at `model.stem`, so my non-square guard never fires for them. Fixing it.
- Hiera is real but pre-existing: `TIMMModelWrapper(hiera_tiny_224).forward_features()` already fails on `main` with `AssertionError: normalization of features not supported`, since `_forward_intermediates` forces `norm=True`. My factory does now hand Hiera out as a `MultiScaleFeatureCNN` though, which promises a `forward_features` that doesn't work — so either I reject it in the factory, or stop forcing `norm=True` in the CNN wrapper (fixes Hiera, but changes swin's output). Which would you prefer?

liopeer:
Sorry for the pending reviews @saud5150 btw. I am super swamped on another project, but I think I will have time to do some reviews and merges in the second half of the week!

PR #934, opened 2026-08-07
Fix broken "Monitor training progress" docs link

minutechreview:
Closes #932

## What was wrong

The Features list linked to `pretrain_distill.html#logging`, which returns **404**. `pretrain_distill` is a docs *directory*, not a page — the neighbouring links already use the directory form (e.g. `pretrain_distill/methods/distillation.html`).

Fixed in the two places the issue names — the GitHub README and the docs landing page:

```
- https://docs.lightly.ai/train/stable/pretrain_distill.html#logging
+ https://docs.lightly.ai/train/stable/pretrain_distill/index.html#logging
```

## Verification

- `pretrain_distill.html` → `404`
- `pretrain_distill/index.html` → `200`
- The page really does emit `id="logging"`, so the anchor lands on the Loggers section rather than silently falling back to the top of the page
- Anchor source confirmed at `docs/source/pretrain_distill/index.md:217` (`(logging)=` immediately above `## Loggers`)

## Four more instances of the same broken URL (not fixed here)

Grepping the repo, the same dead `pretrain_distill.html` base appears in four other places. I've left them out of this PR since the issue is specifically about the Features link, but they're all currently 404 and worth a follow-up:

| file | link | note |
|---|---|---|
| `src/lightly_train/_cli.py:57` | `pretrain_distill.html` | → `pretrain_distill/index.html` (verified 200) |
| `src/lightly_train/_commands/train.py:88` | `pretrain_distill.html` | → `pretrain_distill/index.html` (verified 200) |
| `examples/notebooks/distillation.ipynb:384` | `pretrain_distill.html` | → `pretrain_distill/index.html` (verified 200) |
| `examples/notebooks/distillation.ipynb:90` | `pretrain_distill.html#train-data` | **needs a decision** — see below |

The first three are a straight swap. The last one is different: I couldn't find a `(train-data)=` anchor anywhere under `docs/source/`, so that anchor looks dangling independently of the path problem and I didn't want to guess at the intended destination.

Happy to fold any or all of these into this PR, or open a separate one — just say which you'd prefer.

PR #940, opened 2026-08-12
Fix remaining broken pretrain_distill docs links

minutechreview:
Follow-up to #934, picking up the instances I listed there but deliberately left out of that PR.

#934 fixed the two occurrences the issue named (README + docs landing page). The same dead base URL was still present in three more places, all now 404:

| file | before | after |
|---|---|---|
| `src/lightly_train/_cli.py:57` | `pretrain_distill.html` | `pretrain_distill/index.html` |
| `src/lightly_train/_commands/train.py:88` | `pretrain_distill.html` | `pretrain_distill/index.html` |
| `examples/notebooks/distillation.ipynb:384` | `pretrain_distill.html` | `pretrain_distill/index.html` |
| `examples/notebooks/distillation.ipynb:90` | `pretrain_distill.html#train-data` | `pretrain_distill/index.html#pretrain-data` |

## The fourth one

In #934 I flagged the notebook's data-guide link as needing a decision, because its `#train-data` anchor doesn't exist. I've since found the intended target: the Data section in `docs/source/pretrain_distill/index.md` is anchored `(pretrain-data)=` (line 120), not `train-data`. So that link now points at `#pretrain-data`.

If that isn't the section you meant it to reach, say so and I'll change it — but it's the only Data anchor in that page.

## Verification

- `pretrain_distill.html` → **404**; `pretrain_distill/index.html` → **200**
- `id="pretrain-data"` is present on the rendered page; `id="train-data"` is not
- `pretrain_distill/index.html` is already the form used elsewhere in these same files (e.g. the `#resume-training` links at `_cli.py:119`/`133`), so this is consistent rather than a new convention
- `distillation.ipynb` still parses as valid JSON (21 cells)
- `_cli.py` and `_commands/train.py` both compile

After this, `grep -rn "pretrain_distill\.html"` returns nothing across the repo.

### DIFF
--- a/examples/notebooks/distillation.ipynb
+++ b/examples/notebooks/distillation.ipynb
@@ -87,7 +87,7 @@
    "id": "6",
    "metadata": {},
    "source": [
-    "See the [data guide](https://docs.lightly.ai/train/stable/pretrain_distill.html#train-data)\n",
+    "See the [data guide](https://docs.lightly.ai/train/stable/pretrain_distill/index.html#pretrain-data)\n",
     "for more information on supported data formats.\n",
     "\n",
     "In this example, the dataset looks like this:\n",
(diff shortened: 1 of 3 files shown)


liopeer (approved):
Great, thanks!

Issue #941, opened 2026-08-12
EoMT validation uses much more GPU memory than training


liopeer:
## Problem

EoMT models run out of GPU memory in the validation step. Training with the same batch size works.

Issue #936 reports this. A user trains `dinov3/vitt16plus-eomt-inst-coco` with `batch_size=1` on a 12 GB card. Training uses about 2.5 GB. Validation stops with `torch.OutOfMemoryError: Tried to allocate 3.15 GiB`.

We want to lower the memory use. The metric values must stay the same.

## Cause

The validation step resizes the predicted masks to the original image size. This follows the EoMT evaluation protocol. The step resizes all queries together.

Here are the numbers from #936. The model has 200 queries. The original image is 2943x1435 pixels. One float32 tensor of shape `(1, 200, 1435, 2943)` needs 3.15 GiB. This is the exact number in the error message.

`get_labels_masks_scores` then makes four more tensors of the same shape: `masks`, `masks_fp32`, the output of `sigmoid()`, and their product. The peak is near 13 GiB.

Training does not have this problem. Training resizes the masks to the model input size, for example 640x640. That tensor is 10 times smaller. Training also skips this code by default, because `metric_args.train` is `False`.

## Tasks

Please start with task 1. We want to agree on the approach on a small diff first. Task 2 and task 3 can follow in separate pull requests.

### Task 1 — instance segmentation (start here)

Files:

- `src/lightly_train/_task_models/dinov2_eomt_instance_segmentation/train_model.py:403-427`
- `src/lightly_train/_task_models/dinov3_eomt_instance_segmentation/train_model.py:407-434`

Split the query dimension into chunks. For each chunk, resize the mask logits, then call `get_labels_masks_scores`. Keep the results and free the chunk.

This gives the same numbers. Bilinear interpolation works on each channel alone. The score reduction `.sum(2)` works on each query alone. A chunk boundary changes neither.

Pick the chunk size from a byte budget, not from a fixed count. Large images then get small chunks. `linear_semantic_segmentation/train_model.py:212-235` shows the same idea for tiles.

Please do not change `get_labels_masks_scores` in `task_model.py`. That code carries an FP32 and TensorRT warning, and the ONNX export path uses it.

Two more steps are free, and we checked that both keep the results:

- Move the predicted masks to the CPU before the metric update. torchmetrics already calls `.cpu()` on them. The plot code also calls `.cpu()` on them.
[...]

PR #942, opened 2026-08-12
perf(eomt): reduce instance validation mask memory

KrzysiekSko:
## What has changed and why?

Implements Task 1 from #941 for the DINOv2 and DINOv3 EoMT instance-segmentation validation paths.

Validation resized all query mask logits to the original image size at once. `get_labels_masks_scores` then materializes the boolean masks, their float copy, the sigmoid output and the product, so the peak grows with `num_queries * image_h * image_w`. Training does the same work at the model input size and is an order of magnitude smaller.

The queries are now scored in chunks. Bilinear interpolation and the mask score reduction act on every query alone, so a chunk boundary does not change the result.

The chunk size comes from the training footprint rather than from a fixed byte budget:

```python
chunk_size = max(1, num_queries * resize_pixels // image_pixels)
```

Training scores all queries at the model input size, so a chunk covers the same number of mask elements at the original image size. There is no constant to re-tune when the model input size or the query count changes. Images up to the model input size give a single chunk, which leaves them byte-for-byte unchanged.

Completed masks move to the CPU per chunk, so at most one chunk is resident on the device. Labels and scores stay on their original device for the distributed metric sync.

`get_labels_masks_scores` in `task_model.py` is unchanged.

Refs #941. Supersedes #944.

## How has it been tested?

`tests/_task_models/eomt/test_instance_segmentation.py` covers the chunk arithmetic (training parity, single chunk for small images, the clamp to one), equality with the pre-change pipeline for both the multi-chunk and the single-chunk case, the padded-region crop, the output shapes and the mask device.

End-to-end against the real `get_labels_masks_scores` with 60 queries, a 640x640 model input, a (320, 640) crop and an 800x1600 image (chunk size 19, 4 chunks):

| | peak RSS | labels | masks | scores |
| --- | --- | --- | --- | --- |
| before | 1.73 GiB | — | — | — |
| after | 1.01 GiB | identical | identical | identical |

On CPU all three outputs are bit-identical. `labels` and `masks` are elementwise and stay bit-identical on CUDA too; `scores` sums over the mask pixels, and the CUDA reduction split depends on the outer dimension, so it is mathematically identical but may differ in the last ULP. The tests assert exact equality for labels and masks and `assert_close` for scores.

Not run here: the `torch.cuda.max_memory_allocated()` comparison and the before/after validation run on a real checkpoint, since this machine has no GPU. Both are still worth doing before merge.

- [x] Yes

- [ ] Yes

### DIFF
--- a/src/lightly_train/_task_models/dinov2_eomt_instance_segmentation/train_model.py
+++ b/src/lightly_train/_task_models/dinov2_eomt_instance_segmentation/train_model.py
@@ -44,6 +44,9 @@ from lightly_train._task_models.dinov2_eomt_instance_segmentation.transforms imp
     DINOv2EoMTInstanceSegmentationValTransformArgs,
 )
 from lightly_train._task_models.eomt import hooks
+from lightly_train._task_models.eomt.instance_segmentation import (
+    get_instance_segmentation_prediction,
+)
 from lightly_train._task_models.train_model import (
     TaskStepResult,
     TrainModel,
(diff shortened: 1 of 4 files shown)


KrzysiekSko:
Addressed in c21701f and formatted in 6015441.

The chunk budget now uses the larger materialized mask surface by pixel count across `model_image_size` and the final `image_size`, so the first resize cannot bypass the budget when the final image is smaller. I added a regression where `model_image_size=(8, 10)` and `image_size=(4, 5)`; a two-query model-resolution budget produces chunk boundaries `[2, 2, 1]`.

Validation:
- `uv run pytest tests/_task_models/eomt/test_instance_segmentation.py -q`: 3 passed
- focused Ruff check: passed
- focused Ruff format check: passed
- `git diff --check`: passed

liopeer (approved):
LGTM! Thanks for the contribution and sorry that it took quite some time 🚀

Issue #933, opened 2026-08-07
[BUG] DINOv3 LT-DETR: -notpretrained backbone aliases broken or missing for most sizes


liopeer:
### Describe the Bug

You can't reliably turn off backbone pretraining for DINOv3 in LT-DETR object detection (`train_object_detection`). DINOv2 already does this the right way, using dedicated `*-notpretrained` model names. DINOv3 doesn't.

There are 9 DINOv3 backbone sizes for LT-DETR. Only one (`vitt16`) has a `-notpretrained-ltdetr` name at all, and it doesn't work: `dinov3/vitt16-notpretrained-ltdetr` and `dinov3/vitt16-ltdetr` point to the exact same config class. So the "notpretrained" name still loads pretrained weights. The other 8 sizes (ConvNeXt-tiny/small/base/large, ViT-tiny+/small/base/large) have no `-notpretrained-ltdetr` name at all.

DINOv2 gets this right for its main sizes: `dinov2/vits14-notpretrained-ltdetr`, `vitb14-notpretrained-ltdetr`, `vitl14-notpretrained-ltdetr`, and `vitg14-notpretrained-ltdetr` each point to their own config class (e.g. `DINOv2ViTSmallNotPretrained`), which points to a real "no weights" backbone entry (`url=""` in `dinov2_vit_src/configs/__init__.py`). No pretrained weights get downloaded.

There is a workaround for DINOv3 today: pass `model_args={"backbone_args": {"weights": None, "pretrained": False}}` to `train_object_detection`. It works, but it's undocumented and easy to get wrong — passing only one of the two keys either silently loads pretrained weights anyway, or crashes with a `TypeError`. That's not a reasonable way to ask for "just give me a random-init backbone."

### Current State (DINOv3 LT-DETR aliases)

| Size | Pretrained alias | `-notpretrained-ltdetr` status |
|---|---|---|
| ConvNeXt-Tiny | `dinov3/convnext-tiny-ltdetr` | missing |
| ConvNeXt-Small | `dinov3/convnext-small-ltdetr` | missing |
| ConvNeXt-Base | `dinov3/convnext-base-ltdetr` | missing |
| ConvNeXt-Large | `dinov3/convnext-large-ltdetr` | missing |
| ViT-Tiny (vitt16) | `dinov3/vitt16-ltdetr` | **broken** — duplicate of pretrained alias, silently still loads pretrained weights |
[...]

PR #945, opened 2026-08-13
fix(dinov3): add missing notpretrained LT-DETR aliases

KrzysiekSko:
This adds the missing DINOv3 *-notpretrained backbone presets used by LT-DETR object detection, so the corresponding dinov3/...-notpretrained-ltdetr model names resolve cleanly.

Changes:
- add the missing notpretrained DINOv3 backbone entries in DINOv3Package
- add a regression test that the new LT-DETR aliases are listed by LTDETRObjectDetection.list_model_names()

This keeps the change small and focused on the runtime alias gap reported in #933.


liopeer:
This was now already fixed in #962

PR #947, opened 2026-08-14
docs: make quickstart/reference docs yaml-centric with real ready-to-use datasets

liopeer:
Reworks the quickstart and task reference docs to lead with a `config.yaml` instead of an inlined 80-entry class dictionary, and gives every task a real, downloadable example dataset from the new [`coco128_yolo` v0.0.2](https://github.com/lightly-ai/coco128_yolo/releases/tag/v0.0.2) release.

## What changed

**Every task now has a "if you don't have a dataset at hand" starter.** Object detection, instance segmentation (both EoMT and LT-DETRv2), semantic segmentation and panoptic segmentation each get a download snippet plus a minimal `data="<task>/config.yaml"` example, followed by the existing dictionary form for setting things programmatically.

**The 80-class COCO dictionary is gone from the docs.** `quick_start_object_detection.md` and `quick_start_distillation.md` inlined it three times over ~270 lines. Those are now `data="object_detection/config.yaml"`, with a short excerpt of the config showing its shape.

**Image classification** gets a FashionMNIST starter, since the coco128 release ships no classification dataset. It's MIT licensed and comes via `torchvision`, so the docs include a small script to export a subset as class-subfolders.

## On the dataset layout

`v0.0.2` is not a drop-in reshuffle of `v0.0.1`, and the docs had to follow it precisely:

- The 256 images live **once** in a shared top-level `images/` pool. Every labeled task needs `images.zip` **plus** its own task zip, unzipped into the same directory.
- Archives unzip as **siblings into the current directory** — there is no `coco128_yolo/` wrapper folder.
- `object_detection/images` and `instance_segmentation/images` are **symlinks** to `../images` (the YOLO loader derives the labels directory from the images base path, so they must be siblings under one task folder).

This depends on relative `data=` paths resolving against the config file's location rather than the working directory, which needs `lightly-train >= 0.16.3`.

## Verification

Downloaded all five assets from the URLs exactly as written in the docs and loaded each `config.yaml` through the real `*DataArgs` models via `load_data_yaml_if_path` + `resolve_data_paths`:

| Task | Model | train / val |
|---|---|---|
| Object detection | `YOLOObjectDetectionDataArgs` | 128 / 128 |
| Instance segmentation | `YOLOInstanceSegmentationDataArgs` | 100 / 100 |
| Semantic segmentation | `MaskSemanticSegmentationDataArgs` | 100 / 100 |
| Panoptic segmentation | `MaskPanopticSegmentationDataArgs` | 100 / 100 |

(The mask-based tasks cover 100 images per split by design — the upstream panoptic source was missing 28 masks per split, and the loaders skip images without a matching mask.)

Also confirmed: both `images` symlinks survive `unzip` and resolve into the shared pool; every resolved image path exists on disk; `images/train2017` lists 128 images for the pretraining example; the FashionMNIST export script runs verbatim and its `data` dict validates and enumerates 2000 / 500 samples; and `make docs` passes with `--fail-on-warning`.

Model names in the new snippets were checked against the registries — `ltdetrv2-s-coco`, `ltdetrv2-seg-s-coco`, `dinov3/vitl16-eomt-inst-coco`, `dinov3/vitl16-eomt-panoptic-coco`, `dinov2/vitl14-eomt`, `dinov3/vitt16` all resolve.

No training runs were executed as part of this verification — the checks stop at data loading.


docs/source/image_classification.md
```diff
@@ -23,6 +23,67 @@ how to prepare your dataset.
 In multiclass classification, each image is assigned to exactly one class. This is the
 default mode and does not need to be specified explicitly.
 
+If you don't have a dataset at hand,
```
liopeer: Would be better to keep it shorter here.

docs/source/quick_start_distillation.md
```diff
@@ -193,93 +205,7 @@ lightly_train.train_object_detection(
...
-            73: "book",
-            74: "clock",
-            75: "vase",
-            76: "scissors",
-            77: "teddy bear",
-            78: "hair drier",
-            79: "toothbrush",
-        },
-    },
+    data="object_detection/config.yaml",
```
gabrielfruet, line 208: The note about yaml/dict being interchangeable is missing here

docs/source/instance_segmentation/ltdetrv2.md
```diff
@@ -51,6 +51,29 @@ requires a few lines of code using the
 function. See [data](#instance-segmentation-ltdetrv2-data) for more details on how to
 prepare your dataset.
 
+If you don't have a dataset at hand, download a small ready-to-use example dataset:
+
+```bash
+# The shared image pool, used by all tasks.
+wget https://github.com/lightly-ai/coco128_yolo/releases/download/v0.0.2/images.zip && unzip -q images.zip
+# The labels and the config file for this task.
```
gabrielfruet, line 59: I think this wouldn't work on windows because its symlink based

Issue #928, opened 2026-08-03
Finalize EdgeCrafter backbones package implementation


liopeer:
- [ ] adapt the existing feature pyramid to the public interface
- [ ] enable support for multi-channel images following the other packages

lorinczszabolcs:
Hi @liopeer, I addressed the second item of this issue: multi-channel image support, following the other packages.

What I did:
- `EdgeCrafterPackage` now accepts `num_input_channels` and builds the ECViT patch embed for that many channels. It no longer restricts the input to 3 channels.
- Pretrained RGB weights are adapted to the requested channel count on load, through a `load_state_dict` pre-hook that reuses the shared helper the DINOv2, DINOv3, and TIMM packages use.

Note on the stack: this change is stacked on #950 (the first item). GitHub does not let a fork pull request use another fork branch as its base. So this PR is opened against `main` and shows both commits until #950 merges. After #950 is merged I will rebase this branch onto `main` so it contains only the multi-channel commit.

PR #950, opened 2026-08-22
feat(ecvit): add multi-scale feature support

lorinczszabolcs:
Part of #928 and #925.

## What has changed and why?

`ECViTModelWrapper` implemented `ModelWrapper` only, so the EdgeCrafter ECViT backbone could not expose its feature pyramid through the multi-scale interface that dense prediction heads use. This adapts the existing pyramid to that interface (the first item of #928).

- `ECViTModelWrapper` now implements `MultiScaleFeatureCNN`, and `EdgeCrafterPackage` is a `MultiScaleFeaturePackage`.
- `forward_multiscale_features` exposes the existing `(P3, P4, P5)` pyramid from `forward`. The strides are `[8, 16, 32]` and all three levels share the same channel dimension.
- I used the CNN interface, because the pyramid has increasing strides.
- ECViT now also works as a linear semantic segmentation backbone, so I added the linear-seg configs for the four ECViT presets, like the other packages.

The second item of #928 (multi-channel input) is left for a separate change.

## How has it been tested?

New unit tests in `tests/_models/ecvit/test_ecvit.py` cover the feature dimensions and strides, output shapes, output order, invalid indices, the match with `forward_features` for the last level, and the uniform dimensions for the `proj_dim=None` presets. Run them with:

```
uv run --frozen pytest tests/_models/ecvit/
```

- [x] Yes


tests/_models/ecvit/test_ecvit.py
```diff
@@ -272,6 +272,116 @@ def test_model_wrapper_interface(self) -> None:
         assert features["features"].shape == torch.Size([2, 16, 1, 1])
         assert pooled["pooled_features"].shape == torch.Size([2, 16, 1, 1])
 
+    def test_multiscale_feature_dims(self) -> None:
+        model = ECViTModelWrapper(
```
liopeer, line 276: Are we sure this does not download weights?

liopeer (approved):

liopeer (approved):
LGTM! Thanks a lot 🚀

PR #959, opened 2026-08-28
feat(transforms): expose random resized crop ratio

### LINKED ISSUE #958: [FEAT] Expose `ratio` for random resized crop
The `ratio` argument for the random resized crop is not exposed, i.e. can't be wired through which should be fixed for the pretraining pipeline.

ergouziworld:
Closes #958

## What

Expose the random resized crop aspect-ratio range through `transform_args.random_resize.ratio` and pass it through to Albumentations.

- Keep the existing default `(0.75, 4 / 3)`.
- Validate that both ratio bounds are positive and ordered.
- Document the new setting and add coverage for propagation and invalid values.

## Testing

- `python -m compileall -q src tests` (passed)
- Targeted pytest could not run locally because pytest and the project runtime dependencies are not installed; the repository CI will run the full test suite.

### DIFF
--- a/docs/source/settings/pretrain_settings.md
+++ b/docs/source/settings/pretrain_settings.md
@@ -623,10 +623,14 @@ lightly_train.pretrain(
 		"random_resize": {
 			"min_scale": 0.2,
 			"max_scale": 1.0,
+			"ratio": [0.75, 1.3333333333333333],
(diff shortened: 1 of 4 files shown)


liopeer:
Already merged in #970

PR #960, opened 2026-08-28
feat(transforms): expose ratio for random resized crop

### LINKED ISSUE #958: [FEAT] Expose `ratio` for random resized crop

hajar-benhadj:
## Description

Fixes #958.

The `ratio` argument of `RandomResizedCrop` was not exposed: users could configure `random_resize.min_scale`/`max_scale` in `transform_args` but the crop aspect ratio range was always the albumentations default.

## Changes

- `RandomResizeArgs` (src/lightly_train/_transforms/transform.py): add `min_ratio` (default `3/4`) and `max_ratio` (default `4/3`) plus `ratio_as_tuple()`. Defaults match the previous (implicit) albumentations behavior, so this is fully backward compatible.
- `_get_RandomResizedCrop` (src/lightly_train/_transforms/view_transform.py): pass `ratio=args.scale.ratio_as_tuple()` to `RandomResizedCrop` for both albumentations 2.x and <2.x branches.
- No changes needed in the method transforms (SimCLR, DINO, ...): they already forward the full `RandomResizeArgs` object via `scale=transform_args.random_resize`, so `ratio` flows through automatically.
- Tests in tests/_transforms/test_view_transform.py: ratio defaults, ratio/scale pass-through to the constructed transform.
- Docs: document `min_ratio`/`max_ratio` in the `random_resize` section.

## Usage

```python
lightly_train.pretrain(
    ...,
    transform_args={
        "random_resize": {
            "min_scale": 0.2,
            "max_scale": 1.0,
            "min_ratio": 0.75,
            "max_ratio": 1.33,
        },
    },
)
```

## Testing

- `tests/_transforms/test_view_transform.py`: 518 passed (includes 3 new tests)
- `tests/_methods/simclr/test_simclr_transform.py`: 1 passed

### DIFF
--- a/docs/source/settings/pretrain_settings.md
+++ b/docs/source/settings/pretrain_settings.md
@@ -628,6 +628,25 @@ lightly_train.pretrain(
 )
 ```
 
+The aspect ratio range of the crops can be configured with `min_ratio` and
+`max_ratio` and defaults to `(3 / 4, 4 / 3)`.
+
+```python
+import lightly_train
+
+lightly_train.pretrain(
+	...,
+	transform_args={
+		"random_resize": {
+			"min_scale": 0.2,
+			"max_scale": 1.0,
+			"min_ratio": 0.75,
+			"max_ratio": 1.33,
+		},
+	},
+)
+```
+
 #### `random_flip`
 
 Dictionary to configure random flipping augmentation. By default, horizontal flipping is
--- a/src/lightly_train/_transforms/transform.py
+++ b/src/lightly_train/_transforms/transform.py
@@ -47,9 +47,14 @@ class ResizeArgs(PydanticConfig):
 class RandomResizeArgs(PydanticConfig):
     min_scale: float = 0.08
     max_scale: float = 1.0
+    min_ratio: float = 3 / 4
+    max_ratio: float = 4 / 3
 
     def as_tuple(self) -> tuple[float, float]:
         return self.min_scale, self.max_scale
+
+    def ratio_as_tuple(self) -> tuple[float, float]:
+        return self.min_ratio, self.max_ratio
 
 
 class RandomResizedCropArgs(PydanticConfig):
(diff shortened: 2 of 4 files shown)


hajar-benhadj:
I have read the CLA Document and I hereby sign the CLA

liopeer:
Opened #970 as a draft on top of this. It keeps your commit as-is and refactors the config surface underneath it.

The reason for the extra work: `ImageClassificationRandomResizeArgs` inherits `RandomResizeArgs`, so `min_ratio`/`max_ratio` also appear on the classification fine-tuning config — but that path uses a second, duplicated crop builder (`_transforms/image_classification_transform.py`) that only forwards scale. So `transform_args={"random_crop": {"min_ratio": 1.0}}` would validate cleanly and be silently ignored. #970 collapses the two builders into one so the ratio reaches both, renames the args to match the op they configure (`random_resize`/`random_crop` → `random_resized_crop`, with deprecation shims for the old keys), and fixes a separate bug where disabling the crop still center-cropped non-square images.

Happy to take review comments on #970 — the feature itself is unchanged from what you wrote here.

liopeer (approved):
Thx a lot for the contribution! 🙌

hajar-benhadj:
Thank you for the thorough write-up — and for keeping the commit as-is while refactoring the config surface underneath it.

The silent-ignore path you describe (`min_ratio`/`max_ratio` validating cleanly but being dropped by the duplicated crop builder in `_transforms/image_classification_transform.py`) is exactly the kind of edge that is easy to miss; collapsing the two builders into one so the ratio reaches both paths sounds like the right fix. The deprecation shims for the old keys are a nice touch too.

I will take a proper look at #970 and leave review comments there. Thanks again for the careful review and the merge!

PR #967, opened 2026-09-03
feat(image-classification): support class weights in loss

### LINKED ISSUE #966: [FEAT] Support class weights in the loss for image classification
Sub-issue of #965.

## Description

Image classification has no way to deal with imbalanced datasets. The only loss setting today is `label_smoothing`, and both losses are built without any class weighting:

- [`image_classification/train_model.py`](https://github.com/lightly-ai/lightly-train/blob/main/src/lightly_train/_task_models/image_classification/train_model.py#L146-L156)
- [`image_classification_multihead/train_model.py`](https://github.com/lightly-ai/lightly-train/blob/main/src/lightly_train/_task_models/image_classification_multihead/train_model.py#L138-L144)

Both tasks should be covered.

## Expected behavior

Add a `class_weights` argument next to the existing `label_smoothing`, following the general design in #965:

```python
class_weights: Literal["auto"] | dict[str, float] | None = None
```

Used like this:

```python
lightly_train.train_image_classification(
    out="out/my_experiment",
    model="dinov3/vitt16",
    data={...},
    model_args={"class_weights": "auto"},
    # or: model_args={"class_weights": {"cat": 1.0, "dog": 3.5}},
)
```

`"auto"` needs a different formula per classification task, because the two tasks use different losses:

- **multiclass** -> `CrossEntropyLoss(weight=...)`, using inverse class frequency scaled so the mean weight is `1.0`. This balances the classes against each other.
- **multilabel** -> `BCEWithLogitsLoss(pos_weight=...)`, using `num_images_without_class / num_images_with_class`. This is different on purpose: `pos_weight` balances positive against negative examples *within* one class, which is what multilabel training needs. #964 is a multilabel case.

Classes with zero examples must not cause a division by zero, and an unknown class name in the dictionary should raise a clear error.

Care is needed when turning the class names into a weight tensor: classes dropped with `ignore_classes` are renumbered internally, so the weights must follow that internal order.

Please add tests for both tasks and document the new argument in `docs/source/image_classification.md`.

---

If you would like to work on this, please leave a comment so we do not duplicate work.

bferanmi806-sketch:
## Summary

Adds class weighting support for image classification and image classification multihead training.

* adds `class_weights=None`, "auto", or a class-name-to-weight mapping
* uses inverse-frequency weights with `CrossEntropyLoss` for multiclass classification
* uses `pos_weight` with `BCEWithLogitsLoss` for multilabel classification
* computes automatic weights from the training split only
* preserves internal class ordering when `ignore_classes` renumbers classes
* handles zero-count classes without division by zero
* adds focused tests for multiclass, multilabel, ignored classes, manual weights, automatic weights, and multihead training
* documents the new argument

## Testing

* `uv run --frozen pytest tests/_task_models/image_classification/test_class_weights.py`
* focused image-classification dataset/train-model tests
* multiclass and multilabel training tests
* Ruff formatting/checks
* mypy checks

Closes #966


docs/source/image_classification.md
```diff
@@ -389,6 +389,60 @@ LightlyTrain models are supported as backbones. For example:
...
 backbones.
 
+(image-classification-class-weights)=
+
+## Class Weights
+
+Imbalanced datasets can make the model good at common classes and bad at rare ones. Use
+`model_args={"class_weights": ...}` to tell the loss that mistakes on rare classes count
+more. This works for both `train_image_classification` and
+`train_image_classification_multihead`, and for both `multiclass` and `multilabel`
```
liopeer: Multihead is actually not a documented feature, since it's meant for internal benchmarking. We should not expose it here. I think that was a problem from my side in the issue description.

docs/source/image_classification.md
```diff
@@ -389,6 +389,60 @@ LightlyTrain models are supported as backbones. For example:
...
+  `{"cat": 1.0, "dog": 3.5}`. Keys are class names, not class IDs or positions. The
+  mapping must contain exactly the included class names: unknown names and missing
+  classes both raise a clear error.
+
+Formulas:
+
+- Multiclass uses `torch.nn.CrossEntropyLoss(weight=...)`. With `"auto"`, weights are
+  inverse class frequencies rescaled so their mean is `1.0`. A class with no training
+  examples gets a neutral weight of `1.0`.
+- Multilabel uses `torch.nn.BCEWithLogitsLoss(pos_weight=...)`. With `"auto"`, each
```
liopeer: Need to look into what the consequence for the mean weight is here. It's important that this does not influence the overall scaling of the losses.

docs/source/image_classification.md
```diff
@@ -389,6 +389,60 @@ LightlyTrain models are supported as backbones. For example:
...
+  class is handled independently as
+  `pos_weight = num_images_without_class / num_images_with_class`, where an image with
+  several labels counts once for each of those classes. A class with no positive
+  training examples gets a neutral `pos_weight` of `1.0` so it is not suppressed if it
+  appears in validation.
+
+Classes dropped with `ignore_classes` are renumbered internally, but manual weights
+still use the original class names. Only included classes are expected in the
+dictionary, and weights follow the internal class order automatically. For multihead
+training, all heads share the same dataset and the same class-weight criterion.
```
liopeer: Here as well, we should not publicly document the multi-head training. However, this is important documentation, but it should move to a docstring somewhere.

src/lightly_train/_task_models/image_classification_multihead/train_model.py
```diff
@@ -136,12 +155,21 @@ def __init__(
         )
 
         self.criterion: Module
+        # All LR heads share the same dataset, so they share one class-weight
+        # criterion. Weights are computed once, not per head.
+        resolved_weights = image_classification_class_weights.resolve_class_weights(
```
liopeer: We're resolving twice, here and also in the TrainModelArgs. That seems useless.

src/lightly_train/_task_models/image_classification_class_weights.py
```diff
@@ -0,0 +1,240 @@
...
+    the model head and dataset mapping. No sorting is applied.
+    """
+    mapping = label_helpers.get_class_id_to_internal_class_id_mapping(
+        class_ids=classes.keys(),
+        ignore_classes=ignore_classes,
+    )
+    return sorted(mapping, key=lambda class_id: mapping[class_id])
+
+
+def count_training_labels(
```
liopeer: Iterating over all the samples seems like not such a great idea. We're already memory-mapping the dataset into an arrow table and should be able to get this information out of there very quickly.

liopeer (commented):
Thx for the contribution. I found some shortcomings about listing the actual class counts. Addressing them now!

bferanmi806-sketch:
Yeah, you’re right. I checked this more closely: `CrossEntropyLoss(weight=...)` normalizes by the effective sample weights, but `BCEWithLogitsLoss(pos_weight=..., reduction="mean")` still divides by the raw element count, so increasing `pos_weight` also increases the overall loss/gradient scale.

I’m going to change the multilabel path to compute the unreduced BCE and normalize by the effective weights in the batch. That should keep the `neg / pos` relative weighting while avoiding the unintended global rescaling.

I’ll add a regression test for that as well.

bferanmi806-sketch:
Pushed the fix in `a8ffbbc`.

Multilabel BCE now uses an effective-weight normalized mean, so `pos_weight` keeps the intended positive/negative weighting without increasing the overall loss scale.

I also removed the auto-weight cap so `"auto"` follows the exact `negative / positive` ratio from #966, kept validation unweighted, and moved the multihead-specific explanation out of the public docs.

Focused tests are passing (40), along with Ruff, mypy, the changed docs examples, and targeted pre-commit checks.

Thanks for catching the scaling issue  should be ready for another look.

liopeer:
Now looking pretty good. Will do a few last checks tomorrow and then it will be ready to merge!

bferanmi806-sketch:
@liopeer Sounds good, thanks! Also, if there are any other issues or areas in the repo where an extra pair of hands would be useful, feel free to point me at them  happy to help.

bferanmi806-sketch:
Hey @liopeer, I pushed a small Python 3.8 compatibility fix in 6a0d1cd8. The affected tests now pass locally on Python 3.8 (13 tests), along with the focused class-weight tests (24 tests). The new GitHub Actions runs are showing action_required and appear to need maintainer approval. Could you approve them when you get a chance? Thanks!

PR #969, opened 2026-09-10
fix(dino): update teacher and center once per accumulation window

### LINKED ISSUE #968: [Bug] DINO teacher EMA and center update once per microbatch with gradient accumulation
### Summary

With gradient accumulation enabled, DINO updates its teacher EMA and its center on every microbatch rather than once per optimizer step. With four microbatches per step, both updates run four times, and later microbatches in the same effective batch see a different teacher and center than earlier ones.

Changing only the accumulation setting therefore changes the teacher targets and the student gradients.

### Environment Details

**Environment:** LightlyTrain 0.17.0, commit [c1e2b0dc](https://github.com/lightly-ai/lightly-train/blob/c1e2b0dcabfacc3bd3574204c643bd43e6aa6157/src/lightly_train/_methods/dino/dino.py), Lightly commit `e7c81fa123baa39334d1590662813798bab9b826`, run through `lightly_train.pretrain`.

The results below are from the pinned version. Source inspection on 2026-09-10 found the same calculation in [295f04f8](https://github.com/lightly-ai/lightly-train/blob/295f04f8e1208d39119d99589763a61d728819a2/src/lightly_train/_methods/dino/dino.py); the full experiment has not been rerun on that commit.

Separate current-branch regression environment: Ubuntu 26.04 LTS (x86_64), Python 3.11.15, PyTorch 2.13.0+cpu, PyTorch Lightning 2.6.5, and Lightly 1.5.26. LightlyTrain is imported from the source checkout at baseline `295f04f8` plus PR #969, with dependencies installed in a uv-managed virtual environment and `PYTHONPATH=src` for the tests. These are the CPU regression details, not a diagnostic dump from the pinned GPU experiment.

### Steps to Reproduce

The test uses the same 128 ordered synthetic images, deterministic two-view transforms, and the same initial student/teacher checkpoint, and compares:

- one batch of 128 with `accumulate_grad_batches=1`;
- four batches of 32 with `accumulate_grad_batches=4`.

Both perform one SGD update at the same resolved learning rate `0.01`, with teacher and center momentum fixed at `0.9` and gradient clipping at `3.0`.

| Difference between the two settings | Current code (relative L2) | Once-per-window repair (relative L2) |
| --- | ---: | ---: |
| Teacher state after the window | `0.1497115` | `0` |
| Center after the window | `0.6940674` | `0` |
| Student gradient after clipping | `7.45e-4` | `5.94e-8` |

Holding both momentum values at 1 removes the teacher and center differences. The repaired run also matches an independent calculation of the teacher EMA, the DINO loss, and the window-wide center. Reproduced on a second host.

The regression code is in [PR #969](https://github.com/lightly-ai/lightly-train/blob/aa10c27ef71369f19d72aa5620fb99b1c9baeb19/tests/_methods/dino/test_dino.py). From a checkout of that PR with its test dependencies installed:

```bash
python -m pytest -q tests/_methods/dino/test_dino.py
```

The CPU run passed 12 tests. Both native accumulation comparisons fail on the unmodified baseline.

### Expected behavior

With the same logical batch, deterministic views, initial state, and resolved learning rate, gradient accumulation should keep the teacher and center fixed throughout the optimizer window and update them once per window. Their states should agree with the non-accumulated reference.

### Root cause

The teacher and center updates are tied to each training-step call, while Lightning accumulates gradients across several such calls before stepping the optimizer.

Even with the student held fixed, applying the teacher EMA four times changes the weight on the old teacher from `0.9` to `0.9^4 = 0.6561`. Updating the center between microbatches also changes the centering used to build the next microbatch's targets.

### Proposed fix

Reproduce the non-accumulated update order at the window boundary:

1. Update both teacher modules once at the start of the window.
2. Keep that teacher and the previous center fixed for every microbatch in the window.
3. Accumulate teacher-output statistics across the window and update the center once at its end.

The tested repair keeps all four microbatches and follows this order. The teacher EMA and the center need different timing: moving both after `optimizer.step()` would not match the tested non-accumulated path.

gss10282025:
## Summary

Update the DINO teacher once at the start of each gradient accumulation window, keep the teacher and center fixed while its microbatches run, and commit the center at the optimizer boundary. This keeps the update order of the non-accumulated path.

Center statistics are weighted by the number of teacher outputs. Half-precision outputs accumulate in FP32 to avoid overflowing the running sum, then the window mean is converted back before Lightly's usual center update.

Fixes #968. A changelog entry is included.

## Validation

- `python -m pytest -q tests/_methods/dino/test_dino.py`: 12 passed on CPU.
- The native Lightning training tests compare one batch of four with four batches of one across three optimizer steps. They check teacher and center state, including a partial final window; student state is also compared for complete windows. Both cases fail on unmodified `295f04f8`.
- Separate checks cover FP16, BF16, FP32, and FP64 center accumulation against Lightly's full-batch center update, including a sum that would overflow in FP16.
- Mypy, Ruff lint/format checks, Python compilation, and `git diff --check` pass for the changed Python files.

The pinned-version GPU results are in the linked issue; GPU tests were not rerun for this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,9 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- Update the DINO teacher EMA and center once per optimizer step when using gradient
+  accumulation, keeping teacher targets fixed within each accumulation window.
+
 ### Security
 
 ## [0.17.0] - 2026-07-28
--- a/src/lightly_train/_methods/dino/dino.py
+++ b/src/lightly_train/_methods/dino/dino.py
@@ -216,6 +216,45 @@ class DINOSGDArgs(SGDArgs):
     weight_decay: float = 0.0001
 
 
+class _AccumulatingDINOLoss(DINOLoss):  # type: ignore[misc]  # Lightly does not expose a typed DINOLoss.
+    """Keep centering fixed until all microbatches in an optimizer step finish."""
+
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self._teacher_sum: Tensor | None = None
+        self._teacher_count = 0
+        self._teacher_dtype = torch.float32
+
+    @torch.no_grad()
+    def update_center(self, teacher_out: Tensor) -> None:
+        accumulation_dtype = (
+            torch.float32
+            if teacher_out.dtype in (torch.float16, torch.bfloat16)
+            else teacher_out.dtype
+        )
+        teacher_sum = teacher_out.sum(
+            dim=(0, 1), keepdim=True, dtype=accumulation_dtype
+        )
+        if self._teacher_sum is None:
+            self._teacher_sum = teacher_sum
+            self._teacher_dtype = teacher_out.dtype
+        else:
+            self._teacher_sum.add_(teacher_sum)
+        self._teacher_count += teacher_out.shape[0] * teacher_out.shape[1]
+
+    @torch.no_grad()
+    def commit_center(self) -> None:
+        if self._teacher_sum is not None:
+            # Reuse Lightly's center reduction and momentum update with the
+            # mean of the whole local window, including a partial final window.
+            window_mean = (self._teacher_sum / self._teacher_count).to(
+                self._teacher_dtype
+            )
+            super().update_center(window_mean)
+            self._teacher_sum = None
+            self._teacher_count = 0
+
+
 class DINO(Method):
     def __init__(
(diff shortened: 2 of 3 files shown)


liopeer:
Thanks a lot @gss10282025 ! We decided against fixing it here and instead fixing it directly in LightlySSL, release a new version and then bump the minimal LightlySSL version here in the LightlyTrain dependencies. A rough draft is up here: https://github.com/lightly-ai/lightly/pull/2062 . Will still change a bit, but if you later want to make changes to in incorporate that we'd be super happy to receive them!

PR #970, opened 2026-09-14
fix(transforms): honor crop ratio in fine-tuning and don't crop when disabled

liopeer:
## Description

Stacked on #960 — this PR's base is `feat/expose-random-resized-crop-ratio`, a mirror of that PR's head commit in this repo (GitHub cannot use a fork branch as a base directly). **The diff here is only what goes on top of #960.** Once #960 merges, GitHub retargets this PR to `main` automatically.

Two bugs found while reviewing #960, plus the internal cleanup that makes the first one impossible to reintroduce.

### `min_ratio` / `max_ratio` were silently ignored when fine-tuning

`ImageClassificationRandomResizeArgs` inherits `RandomResizeArgs`, so the two new fields from #960 also appear on the classification fine-tuning config. But pretraining and fine-tuning had grown two copies of the crop builder, and the fine-tuning copy (`_transforms/image_classification_transform.py`) only forwarded the scale range. So this validated cleanly and did nothing:

```python
lightly_train.train(..., transform_args={"random_crop": {"min_ratio": 1.0, "max_ratio": 1.0}})
```

Both paths now share `get_random_resized_crop` in `_transforms/random_resized_crop.py`.

### Disabling the crop still cropped

`transform_args={"random_resize": None}` built a `RandomResizedCrop` with `scale=(1.0, 1.0)` and albumentations' default ratio range `(3/4, 4/3)`. A crop covering the full image area at an aspect ratio the image does not have is unsatisfiable, so albumentations ran out of sampling attempts and fell through to its central-crop fallback, which clamps the image to the ratio range. A 200x100 image lost a third of its width with cropping turned off. Verified against albumentations 2.0.8:

```python
old = RandomResizedCrop(size=(64, 64), scale=(1.0, 1.0), ratio=(3/4, 4/3), p=1.0)
# 200x100 image with marked left/right edges
out = old(image=img)["image"]
# edges present: False False
```

It now builds a plain `Resize`. `record_geometry` reports the full image as the crop box, which is what a resized view actually covers — relevant for DINOv3.1, whose PaKA loss consumes that geometry.

### Internal cleanup

- The two arg classes are merged into one `RandomResizedCropArgs` holding all four parameters of the op. The old `RandomResizedCropArgs` was an internal `size` + `scale` wrapper that existed only to carry the `None` one level deeper than it needed to go.
- `ViewTransformArgs` carries the crop size as its own `image_size` field. The `scale: ... | None`, its downstream `assert`, and the in-place mutation of the args object inside `ViewTransform.__init__` are all gone.

## User-facing keys are unchanged

`random_resize` for pretraining and `random_crop` for classification fine-tuning stay exactly as they are. They are documented on separate pages, so the inconsistency is not in front of users day to day, and renaming them now would mean migrating twice — once here behind a deprecation shim, and again during the planned migration of pretraining onto the fine-tuning pipeline. Better to do it once, deliberately, as a breaking change there.

So this PR adds no deprecations and no config churn. The only behaviour changes are the two fixes above.

## Testing

- `tests/_transforms`, `tests/_configs`: 3798 passed
- `tests/_methods`: 213 passed, 3 skipped
- `make format`, `make static-checks`: clean (mypy over 646 files)

New tests: `ratio` reaching the classification transform, and the resize-only path keeping the whole image and reporting full-image geometry.

https://claude.ai/code/session_017jRkyfUYRqdbfVSRxxkG8Y


tests/_transforms/test_random_resized_crop.py
```diff
@@ -0,0 +1,28 @@
...
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+#
+from __future__ import annotations
+
+from lightly_train._transforms.random_resized_crop import get_random_resized_crop
+from lightly_train._transforms.transform import RandomResizedCropArgs
+
+
+class TestGetRandomResizedCrop:
```
liopeer: This is not how we name or group tests of functions/methods.

tests/_transforms/test_view_transform.py
```diff
@@ -257,26 +253,69 @@ def test_view_transform_all_args_combinations(
         assert img.dtype == torch.float32
 
 
-class TestGetRandomResizedCrop:
-    def test_ratio_defaults(self) -> None:
-        args = _get_random_resized_crop_args()
-        assert args.scale is not None
-        assert args.scale.ratio_as_tuple() == (3 / 4, 4 / 3)
+class TestViewTransformWithoutRandomResizedCrop:
```
liopeer: Also not how we name tests.

tests/_transforms/test_random_resized_crop.py
liopeer: Fixed in c8cd762c. `get_random_resized_crop` is a function, so it now has module-level tests named after it (`test_get_random_resized_crop`, `test_get_random_resized_crop__ratio`) instead of the wrapper class.

I also dropped the separate `test_ratio_defaults`, which was asserting `RandomResizedCropArgs().ratio_as_tuple()` rather than testing the function at all. The default ratio is now covered where it matters, by checking it reaches the constructed transform.

tests/_transforms/test_view_transform.py
liopeer: Fixed in c8cd762c. Both cases moved onto the existing `TestViewTransform` as `test_no_random_resized_crop__keeps_whole_image` and `test_record_geometry__no_random_resized_crop`, matching the `test_record_geometry_*` neighbours there.

PR #972, opened 2026-09-16
feat(torchvision): add multi-scale feature support to ResNet and ConvNeXt backbones

### LINKED ISSUE #927: Add multi-scale feature support to Torchvision
- [ ] extend the torchvision models to allow emitting multi-scale features

lorinczszabolcs:
## What has changed and why?

Part of #925. Closes #927. Dense prediction tasks need intermediate feature maps from several backbone stages. This PR implements the multi-scale feature interface for the `torchvision` ResNet and ConvNeXt wrappers.

- `ResNetModelWrapper` exposes the four residual stages and `ConvNeXtModelWrapper` exposes the four feature stages.
- Feature dimensions and strides come from a single dummy forward, since torchvision exposes no `feature_info`.
- CNN stages are returned as-is, like the other CNN wrappers.
- `TorchvisionPackage` is now a `MultiScaleFeaturePackage`.
- Backbones that cannot expose stages raise a clear error.

### Design choices

Two points from the issue thread that I resolved with a default, easy to change if you would prefer otherwise:

1. **ShuffleNetV2 is left unsupported.** Its `conv5` shares `stage4`'s /32 stride, so it does not add a clean extra scale. Requesting multi-scale features raises a clear error instead of an ambiguous pyramid, matching your preference for failing over ambiguous fallbacks. I can switch to `stage2/3/4` instead if you would rather support it.
2. **The last ConvNeXt stage is returned without the final LayerNorm.** torchvision keeps ConvNeXt's final `LayerNorm` in the classifier, not in `features`, and the earlier stages have no such norm, so the last stage is returned as-is for consistency. The DINOv3 ConvNeXt wrapper does apply its final norm to the last stage, so if you would rather match that I can apply it here too.

## How has it been tested?

New unit tests in `tests/_models/torchvision/` (`test_resnet.py`, `test_convnext.py`, `test_shufflenet.py`) cover the feature dimensions and strides, output shapes, output order, invalid indices, consistency with `forward_features` for the last stage, and the clear error for ShuffleNetV2. Run them with `uv run --frozen pytest tests/_models/torchvision/`.

- [x] Yes


src/lightly_train/_models/package.py
```diff
@@ -78,3 +78,26 @@ def get_model_wrapper(
     ) -> MultiScaleFeatureViT | MultiScaleFeatureCNN:
         """Wrap the underlying model with a ModelWrapper that supports multi-scale features."""
         ...
+
+    @classmethod
```
liopeer: I think this should not be needed (that's already what the `MultiScaleFeaturePackage` exists to ensure).

liopeer: Same for `list_multiscale_model_names`. It's either all or nothing. Otherwise it becomes too complicated.

src/lightly_train/_task_models/linear_semantic_segmentation/config.py
```diff
@@ -249,5 +249,42 @@ class DINOv3ConvNextSmallEupe(LinearSemanticSegmentationConfig):
     class DINOv3ConvNextBaseEupe(LinearSemanticSegmentationConfig):
         backbone_name: str = "dinov3/convnext-base-eupe"
 
+    # --- Torchvision (CNN backbones) ---
```
liopeer, line 282: Do we need this? Can't we work with the fallback? Would prefer not having this grow into absolute insanity, i.e. we should not list all models here.

src/lightly_train/_models/torchvision/torchvision.py
```diff
@@ -7,12 +7,128 @@
...
+class TorchvisionModelWrapper(Module, MultiScaleFeatureCNN):
     _torchvision_models: list[type[Module]]
     # Regex pattern for matching model names.
     _torchvision_model_name_pattern: str
+
+    # Cache for the multi-scale feature dimensions and strides. Filled on first use.
+    _multiscale_cache: tuple[list[int], list[int]] | None = None
+
+    def multiscale_feature_dims(self) -> list[int]:
+        dims, _ = self._get_multiscale_dims_and_strides()
```
liopeer: Should not make it a private method if called here.

src/lightly_train/_models/torchvision/torchvision.py
```diff
@@ -7,12 +7,128 @@
...
+        """Returns the cached feature dimensions and strides, reading them on first use."""
+        if self._multiscale_cache is None:
+            self._multiscale_cache = _multiscale_dims_and_strides(
+                model=self.get_model(),
+                extract_stages=self._extract_multiscale_stages,
+            )
+        return self._multiscale_cache
+
+
+def _multiscale_dims_and_strides(
```
liopeer: Are these being cached somewhere afterwards?

liopeer (approved):
LGTM! 🚀

PR #976, opened 2026-09-21
docs(licensing): state that AGPL-3.0 covers trained model weights

liopeer:
## Why

Our docs describe what the **Commercial License** grants — "Covers model training and deployment", "keeping your code and models private" — but nowhere state the converse: that under AGPL-3.0 we consider trained weights to be covered too. A reader has to infer it, and the natural inference from "keeping your models private" is confidentiality, not copyleft.

Ultralytics, the closest precedent, ships the same stock AGPL-3.0 `LICENSE` but says it explicitly on https://www.ultralytics.com/license: *"All Ultralytics YOLO trained models fall under the AGPL-3.0 License by default. The AGPL-3.0 License covers the training code and the models produced by that training code."* This PR adopts the same explicit assertion.

## What

One canonical statement plus short pointers, rather than restating the argument in five places:

- `docs/source/licensing.md` — new **Model Weights** section (the single source of truth), covering both the obligation and the Commercial/Community escape hatch.
- `src/lightly_train/_license.py` — the weights statement added to `LICENSE_INFO`. This is the highest-leverage surface: it is already stamped into every checkpoint and ONNX export, so it reaches users who never open the docs. No changes needed at the three consumers, it propagates automatically.
- `README.md` + `docs/source/index.md` — one sentence on the AGPL-3.0 bullet. These two blocks are byte-identical duplicates, so both get the same edit.
- `docs/source/faq.md` — a short dropdown following the existing pattern.

**`LICENSE` is unchanged** and remains stock, unmodified AGPL-3.0, so our OSI status and packaging are unaffected.

## Please review the wording carefully

The assertion that copyleft reaches weights is a contested reading — the mainstream analysis treats weights as program *output* rather than a derivative work. This PR deliberately puts the claim in the docs, where it functions as a stated position, and not in the license grant, where it would have to survive legal scrutiny. **Worth having legal sign off on the exact wording before merging.**

Out of scope by decision: the upstream weight-license constraint (checkpoints fine-tuned from DINOv3 / LingBot Vision are governed by Meta's DINOv3 License Agreement regardless of which Lightly license the user holds). That stays documented in `NOTICE` and `licences/`.

## Verification

- `make format` clean.
- `make docs` builds with no new warnings; verified the new section, the FAQ dropdown, and the index bullet all render in the generated HTML.
- `pytest tests/_task_models/ltdetr_object_detection/test_model_io_metadata.py tests/test__checkpoint.py tests/_callbacks/test_callback_helpers.py` — 12 passed, 1 skipped. Tests reference `LICENSE_INFO` symbolically rather than by literal text, so the notice change is safe.


docs/source/licensing.md
```diff
@@ -25,6 +25,33 @@ experimenting with LightlyTrain. Full access to all features without cost.
...
+AGPL-3.0 license covers the training code **and the model weights produced by that
+training code**, so weights trained under AGPL-3.0 are subject to AGPL-3.0 by default.
+
+The obligation applies once you distribute the weights or make them available to others
+over a network. It then means publishing the weights under AGPL-3.0 together with the
+complete corresponding source code of any application that uses them, including your
+modifications, training scripts, and configuration files.
+
+Training and using weights purely internally, without distributing them and without
+exposing them to outside users over a network, does not trigger this obligation.
```
liopeer: This is not actually what we want to say. Any use of the weights in a commercial context need a commercial license. And any non-OSS use under AGPL3.0 (non-commercial) needs a community license.

src/lightly_train/_license.py
```diff
@@ -13,6 +13,8 @@
     "LightlyTrain License Notice\n"
     "\n"
     "Model training and inference in commercial settings require a valid Commercial License.\n"
+    "Model weights trained with LightlyTrain under AGPL-3.0 are subject to AGPL-3.0: the license\n"
+    "covers the training code and the model weights produced by that training code.\n"
```
liopeer: We don't really need a second sentence. We should just extend the first one and say something like "usage of code and weights for training and inference ..."

README.md
```diff
@@ -644,7 +643,9 @@ Lightly**Train** offers flexible licensing options to suit your specific needs:
 
 - **AGPL-3.0 License**: Perfect for open-source projects, academic research, and
   community contributions. Share your innovations with the world while benefiting from
-  community improvements.
+  community improvements. Model weights trained with LightlyTrain under AGPL-3.0 are
```
liopeer: We can leave this one away.

docs/source/faq.md
```diff
@@ -72,6 +72,17 @@ See the [Licensing](licensing.md) page for pricing details and plan inclusions.
 Please [contact us](https://www.lightly.ai/contact) to discuss the best licensing option for your project!
 ```
 
+```{dropdown} <h6>Do I have to open-source my trained model weights?<a class="headerlink" id="do-i-have-to-open-source-my-trained-model-weights" href="#do-i-have-to-open-source-my-trained-model-weights" title="Link to this heading">¶</a></h6>
+If you use Lightly**Train** under the AGPL-3.0 license, yes. The AGPL-3.0 license covers the training
+code and the model weights produced by that training code, so weights trained under AGPL-3.0 are
+subject to AGPL-3.0 by default.
+
+This applies once you distribute the weights or serve them to outside users over a network;
+purely internal training and use does not trigger it. To distribute or deploy your weights
```
liopeer: We should not be explicit about this.

PR #977, opened 2026-09-21
refactor(data): add keypoint detection data args for COCO and YOLO

liopeer:
First PR towards keypoint detection support (see `NOTES_keypoint_detection_ecpose.md`). It covers only the **data args** — the user-facing config classes and the keypoint set resolution. Reading labels off disk is the follow-up PR, stacked on this one.

## What's in it

| File | What |
| --- | --- |
| `_data/keypoint_helpers.py` | `KeypointSetArgs` + the pure resolvers that combine a dataset's keypoint set with a user-declared one |
| `_data/keypoint_detection_dataset.py` | `COCO`/`YOLOKeypointDetectionDataArgs`, mirroring `instance_segmentation_dataset.py` |

## Design decisions worth a look

**No per-dataset presets.** No COCO-17 or CrowdPose-14 preset, no baked sigma table, no baked flip pairs. The keypoint set comes from the dataset (COCO `categories[]`, YOLO `kpt_shape`/`flip_idx`/`kpt_names`) or from the user's `keypoints` argument. Where a format carries nothing — COCO has no sigmas and no flip pairs, YOLO has no sigmas and no skeleton — the field stays `None` and whichever consumer needs it decides. Nothing is invented here.

**Values given twice must agree.** Neither the dataset nor the user takes precedence; a conflict is an error naming both values. Silent precedence in either direction would be impossible to debug.

**One shared keypoint set.** If several included COCO categories declare `keypoints` they must declare the same ones, and the same for YOLO's per-class `kpt_names`. A category that is *excluded* via `ignore_classes` may differ freely.

**Keypoint set resolved at construction.** A bad keypoint config fails when the args are built, not much later when labels are read. COCO `categories[].skeleton` is one-indexed and is converted to zero-indexed on read; `flip_idx` must be a permutation and its own inverse, since flipping twice has to be the identity.

**Filtering fields ship here, enforced in the follow-up.** `min_keypoints`, `include_crowd` and the `skip_if_*` flags are part of the data args schema and feed the mmap hashes, but the code that acts on them lives in the dataset args. Shipping the schema whole avoids changing it in the next PR.

**`min_keypoints` defaults to 0.** The reader reproduces the file faithfully and drops nothing. Filtering zero-keypoint instances is standard in pose training, but that's a training-policy call for the model PR. `include_crowd=False` is the one default that does drop data, so it's an explicit field.

## Deliberately not wired up

No `KeypointDetectionTrainTaskConfig`, no `keypoint_detection` in `TrainTaskConfig.task`, no `train_keypoint_detection()`, no `__init__.py` export, no docs. That `task` field is a closed `Literal`; adding a member would advertise a task that `TASK_TRAIN_MODEL_CLASSES` cannot serve, so it would validate and then crash at model construction.

The `format` discriminator ships on both classes from day one, and `TestKeypointDetectionDataConfig` exercises the discriminated union plus the YAML path in the test file — that's where a field missing from the data args would be dropped silently by `load_data_yaml_if_path`. Those few lines move into `train_task.py` unchanged once a model lands. `YOLOOrientedObjectDetectionDataArgs` is the cautionary example of a data args class that shipped without `format` and so can't join a union.

## Commit type

`refactor` rather than `feat`: per `CONTRIBUTING.md`, `feat` means a user-facing feature and bumps the minor version, and nothing here is reachable by users. `feat` belongs to the PR that ships `train_keypoint_detection`. Happy to change it if you'd rather this cut a release.

## Verification

```
make format && make static-checks          # mypy clean over 650 files
uv run --frozen pytest tests/_data -q      # 369 passed
```


src/lightly_train/_data/file_helpers.py
```diff
@@ -471,6 +471,98 @@ def open_yolo_instance_segmentation_label(
...
+            else:
+                # Exporters commonly write the flag as a float, e.g. "2.000000".
+                vis = int(values[offset + 2])
+                if vis not in keypoint_helpers.VISIBILITIES:
+                    raise ValueError(
+                        f"Expected keypoint visibility to be one of "
+                        f"{list(keypoint_helpers.VISIBILITIES)}, got {vis} for keypoint "
+                        f"{i} in '{label_path}' on line {line_number}."
+                    )
+                if vis == keypoint_helpers.VISIBILITY_UNLABELED:
```
liopeer: What does this mean?

src/lightly_train/_data/keypoint_helpers.py
```diff
@@ -0,0 +1,391 @@
...
+from pydantic import Field, model_validator
+from typing_extensions import Self
+
+from lightly_train._configs.config import PydanticConfig
+
+# Visibility convention, shared by COCO and YOLO pose.
+VISIBILITY_UNLABELED = 0
+VISIBILITY_OCCLUDED = 1
+VISIBILITY_VISIBLE = 2
+VISIBILITIES = (VISIBILITY_UNLABELED, VISIBILITY_OCCLUDED, VISIBILITY_VISIBLE)
```
liopeer: This should be an enum.

src/lightly_train/_data/keypoint_detection_dataset.py
```diff
@@ -0,0 +1,236 @@
...
+
+    format: Literal["yolo"] = "yolo"
+    path: PathLike
+    train: PathLike
+    val: PathLike
+    test: PathLike | None = None
+    """Accepted for compatibility with YOLO data configs. Task training consumes only
+    train and val."""
+    names: dict[int, str]
+    kpt_shape: list[int]
```
liopeer: That's always exactly 2 values, right? So we should probably annotate with tuple, but let lists correctly resolve as well. Pydantic supports that.

src/lightly_train/_data/keypoint_detection_dataset.py
```diff
@@ -0,0 +1,236 @@
...
+    per instance: ``class_id x_center y_center width height`` followed by
+    ``kpt_shape[0]`` keypoints of ``kpt_shape[1]`` values each, normalized to [0, 1].
+    """
+
+    format: Literal["yolo"] = "yolo"
+    path: PathLike
+    train: PathLike
+    val: PathLike
+    test: PathLike | None = None
+    """Accepted for compatibility with YOLO data configs. Task training consumes only
```
liopeer: First line of docstring always needs to fit onto a single line. If more needed -> separate paragraph after the first line (google conventions). There are more locations in this PR where this is wrong.

src/lightly_train/_data/keypoint_detection_dataset.py
```diff
@@ -0,0 +1,236 @@
...
+    train and val."""
+    names: dict[int, str]
+    kpt_shape: list[int]
+    """``[num_keypoints, num_dims]``. ``num_dims`` is 2 for (x, y) or 3 for
+    (x, y, visibility)."""
+    flip_idx: list[int] | None = None
```
liopeer, line 52: What happens if not given?

src/lightly_train/_data/keypoint_detection_dataset.py
```diff
@@ -0,0 +1,236 @@
...
+    kpt_names: dict[int, list[str]] | None = None
+    keypoints: KeypointSetArgs | None = None
```
liopeer: If yolo has no place for it, then why do we have it here???

liopeer: I think it's called `kpt_oks_sigmas` in real yolo. But only that.

src/lightly_train/_data/keypoint_detection_dataset.py
```diff
@@ -0,0 +1,236 @@
...
+    """Keypoint set fields the YOLO config has no place for, e.g. the OKS ``sigmas``.
+    Values given here and in the fields above must agree."""
+    ignore_classes: set[int] | None = Field(default=None, strict=False)
+    skip_if_label_file_missing: bool = False
+    min_keypoints: int = 0
+    """Drop instances with fewer labeled keypoints."""
```
liopeer: AFAIK none of these are supported by original yolo. We're making our lives difficult if we include these from the beginning. Can still be added later, but not now.

src/lightly_train/_data/keypoint_detection_dataset.py
```diff
@@ -0,0 +1,236 @@
...
+        return self
+
+    def _resolve_keypoint_set(self) -> KeypointSetArgs:
+        return keypoint_helpers.resolve_yolo_keypoint_set(
+            kpt_shape=self.kpt_shape,
+            flip_idx=self.flip_idx,
+            kpt_names=self.kpt_names,
+            keypoints=self.keypoints,
+            included_class_ids=self.included_classes.keys(),
+        )
```
liopeer: Why split this into two methods, that's so bad to read. But I think we don't want this KeypointSet thing anyway.

src/lightly_train/_data/keypoint_detection_dataset.py
```diff
@@ -0,0 +1,676 @@
...
+
+    def resolve_data_paths(self, base_dir: Path) -> None:
+        self.path = data_helpers.resolve_path(self.path, base_dir=base_dir)
+
+    def train_data_mmap_hash(self) -> str:
+        return str(
+            (
+                (Path(self.path) / self.train).resolve(),
+                self.names,
+                self.keypoint_set.model_dump_json(),
```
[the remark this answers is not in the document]
liopeer: Agree, can be added.

src/lightly_train/_data/keypoint_detection_dataset.py
liopeer: I think `ignore_classes` we can leave though (but only if OD/instance segmentation YOLO format here in the repo support it like this as well).

tests/_data/test_task_data_args.py
```diff
@@ -462,3 +469,99 @@ def test_resolves_paths_relative_to_data_config_file(self, tmp_path: Path) -> No
             data_args.val.annotations
             == (tmp_path / "configs" / "annotations/val.json").resolve()
         )
+
```
liopeer: This is the wrong module. Test modules have the same names as the implementation modules (except for the `test_` prefix).

src/lightly_train/_data/keypoint_detection_dataset.py
```diff
@@ -0,0 +1,676 @@
...
+            visibility: list[list[int]] = []
+
+            if image_id in annotations_by_image_id:
+                for annotation in annotations_by_image_id[image_id]:
+                    if annotation.get("iscrowd", 0) and not self.include_crowd:
+                        continue
+
+                    instance_keypoints, instance_visibility = self._parse_keypoints(
+                        annotation=annotation,
+                        num_keypoints=num_keypoints,
```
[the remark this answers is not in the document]
liopeer: Fixed on the stacked reader branch in 439dddf0: excluded category IDs are checked before keypoint parsing.

src/lightly_train/_data/keypoint_detection_dataset.py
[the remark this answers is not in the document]
liopeer: Fixed in f1a710e9: both hashes include the full kpt_shape, with a 2D-versus-3D regression test.

src/lightly_train/_data/keypoint_detection_dataset.py
liopeer: Fixed in f1a710e9: kpt_shape is tuple[int, int] with Field(strict=False), so YAML lists still parse.

src/lightly_train/_data/keypoint_detection_dataset.py
liopeer: Fixed in af506cc0: field docstrings now use a one-line summary and separate body.

src/lightly_train/_data/keypoint_detection_dataset.py
liopeer: When flip_idx is absent, horizontal flips must be disabled. This is now stated inline in af506cc0.

src/lightly_train/_data/keypoint_detection_dataset.py
liopeer: Replaced the generic keypoints wrapper with native kpt_oks_sigmas in f1a710e9.

src/lightly_train/_data/keypoint_detection_dataset.py
liopeer: Removed min_keypoints. Kept ignore_classes and skip_if_label_file_missing because both match the existing YOLO object-detection and instance-segmentation args.

src/lightly_train/_data/keypoint_detection_dataset.py
liopeer: Removed KeypointSetArgs and the split resolver/property path in f1a710e9.

tests/_data/test_task_data_args.py
liopeer: Moved the keypoint data-args tests into test_keypoint_detection_dataset.py in f1a710e9.

tests/_data/test_keypoint_helpers.py
```diff
@@ -0,0 +1,63 @@
...
+    def test_valid(self) -> None:
+        assert keypoint_helpers.validate_kpt_shape((17, 3)) == (17, 3)
+
+    @pytest.mark.parametrize("kpt_shape", [(0, 3), (3, 4)])
+    def test_invalid(self, kpt_shape: tuple[int, int]) -> None:
+        with pytest.raises(ValueError):
+            keypoint_helpers.validate_kpt_shape(kpt_shape)
+
+
+class TestValidateYOLOKeypointMetadata:
```
liopeer: Not how we name tests etc.

tests/_data/test_keypoint_helpers.py
```diff
@@ -0,0 +1,65 @@
...
+    keypoint_helpers.validate_flip_idx([0, 2, 1], 3)
+
+
+@pytest.mark.parametrize("flip_idx", [[0, 0, 2], [1, 2, 0]])
+def test_validate_flip_idx__invalid(flip_idx: List[int]) -> None:
+    with pytest.raises(ValueError, match="flip_idx"):
+        keypoint_helpers.validate_flip_idx(flip_idx, 3)
+
+
+def test_validate_kpt_names__invalid() -> None:
```
gabrielfruet, line 37: We only have the invalid variant here. 

I think tests can also be used as documentation to understand the code. So I think it's worth it to add

src/lightly_train/_data/keypoint_detection_dataset.py
```diff
@@ -0,0 +1,220 @@
...
+from lightly_train._data import (
+    data_helpers,
+    file_helpers,
+    keypoint_helpers,
+)
+from lightly_train._data.task_data_args import TaskDataArgs
+from lightly_train.types import PathLike
+
+
+class YOLOKeypointDetectionDataArgs(TaskDataArgs):
```
gabrielfruet, line 29: Why not separate files? we have `coco_object_detection_dataset` and `yolo_object_detection_dataset`

src/lightly_train/_data/keypoint_helpers.py
```diff
@@ -0,0 +1,100 @@
...
+        )
+    if num_dims not in (2, 3):
+        raise ValueError(
+            f"Expected the number of dimensions in 'kpt_shape' to be 2 for (x, y) or "
+            f"3 for (x, y, visibility), got {num_dims}."
+        )
+    return kpt_shape
+
+
+def validate_flip_idx(flip_idx: list[int] | None, num_keypoints: int) -> None:
```
gabrielfruet, line 30: Although internal, I think it's worth it to add a better documentation on the functions here. At least was pretty hard to me to understand flip idx and sigmas.

gabrielfruet (approved):
Small stuff, I will not block because of this but I highly recommend addressing

PR #962, opened 2026-09-01
fix(dinov3): fix and add missing LT-DETR -notpretrained backbone aliases

minutechreview:
## What has changed and why?

Fixes #933. DINOv3 LT-DETR object detection backbones could not reliably turn off pretraining. DINOv2 already does this correctly via dedicated `*-notpretrained` model names; DINOv3 didn't:

- `dinov3/vitt16-notpretrained-ltdetr` existed but was **broken**: it was registered on the exact same config class as `dinov3/vitt16-ltdetr`, so the "notpretrained" name silently still loaded pretrained weights.
- The other 8 DINOv3 LT-DETR sizes (ConvNeXt-tiny/small/base/large, ViT-tiny+/small/base/large) had **no** `-notpretrained-ltdetr` alias at all.

This follows the exact pattern already used for DINOv2 (`DINOv2ViTSmallNotPretrained`, `DINOv2ViTBaseNotPretrained`, `DINOv2ViTLargeNotPretrained`, `DINOv2ViTGiantNotPretrained` in `config.py`), and the same subclass-only-overrides-`backbone_name` idiom already used elsewhere in the file (e.g. `EdgeCrafterECViTTinyNotPretrained`).

### Changes

1. **`src/lightly_train/_models/dinov3/dinov3_package.py`** (`DINOv3Package.MODEL_NAME_TO_INFO`): added `functools.partial(builder, pretrained=False)` "no weights" entries for `convnext-tiny`, `convnext-small`, `convnext-base`, `convnext-large`, `vits16`, `vitb16`, `vitl16` (mirroring the existing `vitt16-notpretrained` / `vitt16plus-notpretrained` entries).
2. **`src/lightly_train/_task_models/ltdetr_object_detection/config.py`**: added one dedicated `*NotPretrained` config class per DINOv3 LT-DETR size, each registered under `dinov3/<size>-notpretrained-ltdetr` and pointing `backbone_name` at the new no-weights entry. Fixed the existing broken `dinov3/vitt16-notpretrained-ltdetr` alias by removing it from `DINOv3ViTTiny`'s registration and giving it its own `DINOv3ViTTinyNotPretrained` class.
3. **Bonus (explicitly requested in the issue)**: added `dinov2/vits14-noreg-notpretrained-ltdetr`, wiring up the previously-unused `vits14-noreg-notpretrained` backbone entry, following the same pattern.
[...]

PR #981, opened 2026-09-21
fix(dinov3): wire LT-DETR -eupe aliases to EUPE backbones

### LINKED ISSUE #980: [BUG] Weights-modifier aliases resolve to the wrong backbone: all six `-eupe-ltdetr` and both `-notpretrained-linear` (same bug class as #933)
### Describe the Bug

#933 found that `dinov3/vitt16-notpretrained-ltdetr` was registered on the *same* config class as `dinov3/vitt16-ltdetr`, so the `-notpretrained` name silently still loaded pretrained weights. #962 fixes that for DINOv3 LT-DETR.

The same wiring mistake exists in two more places that #933 did not cover. The underlying defect is general:

> An alias is registered on a config class whose `backbone_name` does not carry the alias's weights modifier, so the alias silently resolves to a **different set of weights than its name promises**.

Both remaining cases are on `main` today and are independent of #962.

### 1. All six `-eupe-ltdetr` aliases load non-EUPE weights

`src/lightly_train/_task_models/ltdetr_object_detection/config.py` registers every `-eupe-ltdetr` alias as an extra name on the *plain* config class:

| Alias | Resolves to class | Its `backbone_name` | Weights actually loaded |
|---|---|---|---|
| `dinov3/vitt16-eupe-ltdetr` | `DINOv3ViTTiny` | `dinov3/vitt16` | Lightly distillationv2 |
| `dinov3/vits16-eupe-ltdetr` | `DINOv3ViTSmall` | `dinov3/vits16` | LVD-1689M |
| `dinov3/vitb16-eupe-ltdetr` | `DINOv3ViTBase` | `dinov3/vitb16` | LVD-1689M |
| `dinov3/convnext-tiny-eupe-ltdetr` | `DINOv3ConvNeXtTiny` | `dinov3/convnext-tiny` | LVD-1689M |
| `dinov3/convnext-small-eupe-ltdetr` | `DINOv3ConvNeXtSmall` | `dinov3/convnext-small` | LVD-1689M |
| `dinov3/convnext-base-eupe-ltdetr` | `DINOv3ConvNeXtBase` | `dinov3/convnext-base` | LVD-1689M |

Verified at runtime:

```python
from lightly_train._task_models.ltdetr_object_detection.config import LTDETR_MODEL_REGISTRY

LTDETR_MODEL_REGISTRY.get(alias="dinov3/vitt16-eupe-ltdetr")().backbone_name
# 'dinov3/vitt16'          <- expected 'dinov3/vitt16-eupe'
LTDETR_MODEL_REGISTRY.get(alias="dinov3/convnext-base-eupe-ltdetr")().backbone_name
# 'dinov3/convnext-base'   <- expected 'dinov3/convnext-base-eupe'
```

All six `-eupe` backbone entries already exist in `DINOv3Package.MODEL_NAME_TO_INFO` (pointing at `huggingface.co/facebook/EUPE-*`), so nothing is missing at the backbone layer — only the config classes are miswired.

**This one is licensing-relevant in both directions.** `docs/source/object_detection.md` states that models with EUPE weights are under the [FAIR Noncommercial Research License](https://github.com/facebookresearch/EUPE?tab=License-1-ov-file) and lists all six aliases in the "DINOv3 ViT/ConvNext backbones with EUPE weights" dropdowns. So today:

- a user who picks `-eupe-ltdetr` to *get* EUPE weights does not get them, and
- a user who avoids `-eupe-ltdetr` for license reasons is avoiding something that never loaded EUPE weights anyway.

For contrast, the EoMT registry gets this right — `dinov3/vitt16-eupe-eomt` has its own `ViTTinyEUPE16` class with `backbone_name = "dinov3/vitt16-eupe"`. That is exactly the fix pattern.

### 2. Both `-notpretrained-linear` aliases load pretrained weights

`src/lightly_train/_task_models/linear_semantic_segmentation/config.py` has the *literal* #933 bug — the alias is a second name on the pretrained class:

```python
@LINEAR_SEG_MODEL_REGISTRY.register(
    "dinov3/vitt16-linear",
    "dinov3/vitt16-notpretrained-linear",   # <- same class as the pretrained alias
)
class DINOv3ViTT16(LinearSemanticSegmentationConfig):
    backbone_name: str = "dinov3/vitt16"
```

Same for `dinov3/vitt16plus-notpretrained-linear` on `DINOv3ViTT16Plus`. Verified at runtime:

```python
LINEAR_SEG_MODEL_REGISTRY.get(alias="dinov3/vitt16-notpretrained-linear")().backbone_name
# 'dinov3/vitt16'   <- expected 'dinov3/vitt16-notpretrained'
```

A user asking for a randomly-initialized linear-probe backbone silently downloads and loads distillationv2 weights — which makes linear-probe baseline numbers wrong, not just surprising. The `vitt16-notpretrained` / `vitt16plus-notpretrained` backbone entries already exist. The DINOv2 block directly above (`DINOv2ViTSmallNotPretrained` etc.) is wired correctly and is the template.

### Root Cause / Where to Fix

Same two-line pattern used by #962 and by the DINOv2/EoMT entries — give each modifier alias its own subclass that overrides only `backbone_name`:

1. `ltdetr_object_detection/config.py` — remove the six `-eupe-ltdetr` aliases from the plain classes; add `DINOv3ViTTinyEUPE`, `DINOv3ViTSmallEUPE`, `DINOv3ViTBaseEUPE`, `DINOv3ConvNeXtTinyEUPE`, `DINOv3ConvNeXtSmallEUPE`, `DINOv3ConvNeXtBaseEUPE`, each with `backbone_name = "dinov3/<size>-eupe"`.
2. `linear_semantic_segmentation/config.py` — remove the two `-notpretrained-linear` aliases from `DINOv3ViTT16` / `DINOv3ViTT16Plus`; add `DINOv3ViTT16NotPretrained` / `DINOv3ViTT16PlusNotPretrained` with `backbone_name = "dinov3/vitt16-notpretrained"` / `"dinov3/vitt16plus-notpretrained"`.

### Expected Behavior

For every registered alias, the resolved config's `backbone_name` carries the same weights modifier as the alias. `<x>-eupe-<task>` loads EUPE weights; `<x>-notpretrained-<task>` loads no weights.

### Suggested Regression Test

#962 adds `test_notpretrained_ltdetr_alias__resolves_to_distinct_backbone`, which is per-alias. Since this bug has now appeared three times in three registries, a single registry-wide invariant test would be cheaper than chasing it a fourth time: for every alias in every task-model registry, assert that the resolved `backbone_name` matches the alias with its task suffix stripped (with an explicit allowlist for the intentional short names such as `ltdetrv2-s`/`-m`/`-l`/`-x` and the `edgecrafter/*` aliases). I scanned all 10 `_task_models/*/config.py` registries this way on `main` (`7c5927d9`) — the eight aliases above are the only real violations; every other diff is one of those intentional short names.

### Documentation

Once the aliases resolve correctly, the EUPE dropdowns in `docs/source/object_detection.md` become accurate as written and need no change. No doc change is needed for the linear-semseg aliases either — they are currently undocumented.

### Additional Context

Found while reviewing #962, which fixes the DINOv3 LT-DETR half of this. Filing separately rather than expanding that PR's scope, since #962 matches #933 exactly and these touch a different registry (and, for EUPE, a licensing-facing claim that deserves its own changelog entry).

Parsiffall1:
## What has changed and why?

Fixes #980 (EUPE half).

All six `dinov3/<size>-eupe-ltdetr` aliases were registered as extra names on the *plain* LT-DETR config classes, so `LTDETR_MODEL_REGISTRY.get(alias=...).backbone_name` resolved to the non-EUPE backbone. Users (and the docs dropdown that lists these aliases under EUPE) silently got Lightly distillation / LVD-1689M weights instead of the FAIR Noncommercial Research License EUPE weights.

This follows the same subclass-only-overrides-`backbone_name` pattern already used for:
- EoMT EUPE (`ViTTinyEUPE16` etc. with `backbone_name = "dinov3/vitt16-eupe"`)
- LT-DETR `-notpretrained` (#962 / #933)

### Changes

1. **`src/lightly_train/_task_models/ltdetr_object_detection/config.py`**: remove the six `-eupe-ltdetr` aliases from the plain classes; add `DINOv3ConvNeXtTinyEUPE` / `SmallEUPE` / `BaseEUPE` and `DINOv3ViTTinyEUPE` / `SmallEUPE` / `BaseEUPE`, each pointing `backbone_name` at the existing `dinov3/<size>-eupe` backbone entry.
2. **Regression test** `test_eupe_ltdetr_alias__resolves_to_distinct_backbone`: for each alias, assert a distinct config class and the expected `-eupe` `backbone_name`.
3. **`CHANGELOG.md`**: Unreleased > Fixed entry.

### Out of scope

The `-notpretrained-linear` half of #980 is **already fixed on main** (dedicated `DINOv3ViTT16NotPretrained` / `DINOv3ViTT16PlusNotPretrained` classes). Verified locally; not touched here.

## How has it been tested?

```bash
uv run --frozen pytest \
  tests/_task_models/ltdetr_object_detection/test_task_model.py::test_eupe_ltdetr_alias__resolves_to_distinct_backbone \
  tests/_task_models/ltdetr_object_detection/test_task_model.py::test_notpretrained_ltdetr_alias__resolves_to_distinct_backbone \
  -v
# 16 passed
```

## Checklist

- [x] Conventional Commits PR title
- [x] Related tests green locally
- [x] CHANGELOG updated


liopeer (approved):
LGTM! Thx for the contribution 🚀

PR #983, opened 2026-09-21
fix: reinitialize detection heads when class counts differ

### LINKED ISSUE #982: [BUG] Object detection class head is not re-initialized when the number of classes changes
### 🧠 Describe the Bug

Fine-tuning object detection from a COCO-pretrained checkpoint onto a different number of classes does **not** re-initialize the classification head. It silently keeps the checkpoint's leading rows.

`_reuse_or_reinit` in `src/lightly_train/_task_models/object_detection_components/hooks.py:168` truncates when the checkpoint has more classes:

```python
if num_classes_checkpoint > num_classes_module:
    # Checkpoint has more classes: truncate to module's expected size.
    adjusted_weights = checkpoint_weight[:num_classes_module, :].to(device)
```

So fine-tuning COCO (80 classes) onto 3 classes starts from the COCO rows for `person`, `bicycle` and `car`. Training still converges, but from an arbitrary bias that depends on COCO's class ordering.

This contradicts:

- The `train_object_detection` docstring: *"Otherwise, the classification head will be re-initialized."*
- The module docstring, which claims the hooks work by "truncating or zero-/re-initializing".
- `src/lightly_train/_task_models/eomt/hooks.py:85`, where the segmentation tasks already re-initialize correctly.

`denoising_class_embed_reuse_or_reinit_hook` (same file, line 31) has the same problem.

### 🔁 Steps to Reproduce

```python
import torch

from lightly_train._task_models import task_model_helpers
from lightly_train._task_models.ltdetr_object_detection.task_model import (
    LTDETRObjectDetection,
)

ckpt = torch.load(
    task_model_helpers.download_checkpoint(checkpoint="ltdetrv2-s-coco"),
    weights_only=False,
    map_location="cpu",
)
model = LTDETRObjectDetection(
    **{**ckpt["model_init_args"], "classes": {0: "cat"}, "load_weights": False}
)

fresh = model.decoder.dec_score_head[0].weight.detach().clone()
model.load_train_state_dict(state_dict=ckpt["train_model"])
after = model.decoder.dec_score_head[0].weight
coco = ckpt["train_model"]["ema_model.model.decoder.dec_score_head.0.weight"]

print("head was re-initialized:", torch.equal(fresh, after))
print("head equals COCO row 0 ('person'):", torch.equal(after[0], coco[0]))
```

Output:

```
head was re-initialized: False
head equals COCO row 0 ('person'): True
```

### 📌 Expected Behavior

When the class counts differ, keep the module's own initialization rather than reusing checkpoint rows, the way `eomt/hooks.py` already does:

```python
state_dict[weight_key] = head_module.weight.detach().clone()
state_dict[bias_key] = head_module.bias.detach().clone()
```

When the counts match, reuse the checkpoint as before.

### 🛠️ Scope

Small and self-contained:

- Fix `_reuse_or_reinit` and `denoising_class_embed_reuse_or_reinit_hook` in `object_detection_components/hooks.py`. For the denoising embedding, keep the trailing `padding_idx` row.
- Update the stale docstrings in that file and in `dfine_decoder.py` / `rtdetrv2_decoder.py`, which describe the truncate/pad behavior.
- Add `tests/_task_models/object_detection_components/test_hooks.py`. These hooks currently have no test coverage. Both decoders (`DFINETransformer`, `RTDETRTransformerv2`) register them, so cover matching, larger and smaller class counts.
- Note the behavior change in `CHANGELOG.md` under `### Changed`.

Happy to help anyone who picks this up.

saitakarcesme:
## What has changed and why?

Closes #982. Loading an object detection checkpoint with a different class count previously truncated or partially reused its classification weights, assigning arbitrary checkpoint classes to the new dataset. Keep the target model's initialized encoder/decoder score heads and denoising class embeddings instead. Preserve the checkpoint's trailing padding embedding, which is not a dataset-specific class. Matching class counts continue to load checkpoint weights unchanged.

The shared hooks serve both D-FINE and RT-DETRv2. Update their comments and the changelog to describe the new behavior.

## How has it been tested?

- New CPU regression tests exercise both real decoder classes through nested `load_state_dict(strict=True)`, including smaller/equal/larger class counts and denoising enabled/disabled. Additional cases cover bias-free heads, dtype preservation, detached copies, and missing checkpoint weights.
- Before the fix: 12 failed / 7 passed. After the fix: all 19 passed.
- Entire `tests/_task_models/object_detection_components` directory: 66 passed on Python 3.12, macOS arm64.
- Repository formatting, Ruff lint, Markdown formatting, documentation code examples (168 passed / 12 skipped), and all pre-commit hooks passed. Ruff's directory scan stalled reading an ancestor workspace ignore file, so Ruff was run with `--no-respect-gitignore`; no user ignore files were modified.
- Full `make type-check` reports 54 diagnostics in 37 files from missing optional dependencies (ONNX, MLflow, bitsandbytes, etc.) and existing third-party typing. Running the same command against an unmodified HEAD archive produces the exact same 54 diagnostics; no new errors were introduced. Full type-check is **not** claimed as passing.
- GPU/export and full training suites were not run. Python 3.13 initially hit a native-dependency segmentation fault during imports; all reported test results above use Python 3.12.

- [x] Yes

- [x] Yes — shared hook docstrings and decoder comments.

Codex assisted with implementation, review and testing. No independent human review or CLA signature is claimed.


liopeer:
@saitakarcesme Could you sign the CLA?

liopeer (approved):
LGTM! Thx for the contribution 🚀

saitakarcesme:
The CLA is now signed and the license/cla check is green. All CI checks, including the Windows unit tests, have also passed. Thanks for reviewing!

